PR #1, opened 2023-06-14
chore: lets use `pnpm`

maraisr:
This should speed up our installs by like a super lot. And given that all workspace deps are linked by default, if you rebuild in one package — it'll reflect automatically in others. 

Does this through a symlink.

Install:

```shell
pnpm install
```

Compile:

```shell
pnpm -r compile
#     ^ recursively find all `compile` scripts and run them
```

Later when we get to versioning, we will use something like [changesets](https://github.com/changesets/changesets) to automate this for us.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,0 @@
-**/node_modules/**
-.vscode
-yarn-error.log
-*.tsbuildinfo
-*.tgz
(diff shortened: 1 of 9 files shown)


rbalicki2:
Awesome, love it, thanks for doing this!!!

PR #5, opened 2024-01-25
Fixing demos

edmondop:
Without this fix, running the demo blows up with the following stacktrace:

```
- wait compiling /_error (client and server)...
- error ./pages/_app.tsx
Error: Cannot find module '../../libs/isograph-babel/BabelPluginIsograph'
Require stack:
```


rbalicki2:
Excellent! Thank you for catching and fixing this :)

We should maybe add a step to our CI pipeline to ensure that `npm run build` runs in both demos.

PR #8, opened 2024-01-28
Aligning syntax between iso and isoFetch using keywords fields/entrypoint

edmondop:


rbalicki2 (approved):
HELL YEAH!!!! Thank you so much for sticking with this and getting through the rebases. I love that this simplifies the mental model (as in, there is only one magical function) and that this is easier to expand upon (as in, adding more keywords is easier than adding more isoX functions.)

PR #9, opened 2024-01-31
Demonstration of possible iso overloads

rbalicki2:


demos/github-demo/src/isograph-components/RepositoryLink.tsx
```diff
@@ -1,33 +1,29 @@
-import { iso } from '@isograph/react';
+import { iso } from "@isograph/react";
 
-import type { ResolverParameterType as RepositoryLinkParams } from '@iso/Repository/RepositoryLink/reader.isograph';
+import { Link } from "@mui/material";
 
-import { Link } from '@mui/material';
-
-export const RepositoryLink = iso<RepositoryLinkParams>`
+export const RepositoryLink = iso(`
```
rbalicki2, line 5: Overloads only work when we pass a string, not use iso with a template literal, so we'll have to convert everything to this syntax

demos/github-demo/src/isograph-components/UserLink.tsx
```diff
@@ -4,11 +4,10 @@ import type { ResolverParameterType as UserLinkParams } from '@iso/Actor/UserLin
...
 
-export const UserLink = iso<UserLinkParams>`
-  field Actor.UserLink @component {
+export const UserLink = iso(` field Actor.UserLink @component {
     login,
   }
-`(UserLinkComponent);
+`)(UserLinkComponent);
 
 function UserLinkComponent(props: UserLinkParams) {
```
rbalicki2, line 12: it looks like if the function is defined separately, it's arg type is not inferred, so we need to specifically say `UserLinkParams` here

libs/isograph-react/src/iso.tsx
```diff
@@ -1,8 +1,52 @@
...
+};
+
+type WhitespaceCharacter = ' ' | '\n';
+type Whitespace<In> = In extends `${WhitespaceCharacter}${infer In}` ? Whitespace<In> : In;
+
+type MatchesWhitespaceAndString<TString extends string, T> =
+  Whitespace<T> extends `${TString}${string}` ? T : never;
+
+export function iso<T>(
+  param: T & MatchesWhitespaceAndString<'field Repository.RepositoryLink', T>,
```
rbalicki2, line 40: I didn't try allowing whitespace in this string, but we could also just enforce in the compiler that no whitespace is allowed between the type name and the dot and the field name, and only one whitespace between the field and typename.

PR #12, opened 2024-02-03
Generating overloads for Iso at compile time

edmondop:
Generate ISO overloads at compile time by adding a new artifact, named `iso.ts` in the `__isograph` subfolder under src, that contains all possible overloads of the iso function for the project, removing the need to use generics.

Update the Babel plugin and the demo accordingly


crates/isograph_cli/src/isograph_literals.rs
```diff
@@ -75,7 +75,7 @@ pub(crate) static ISOGRAPH_FOLDER: &'static str = "__isograph";
 lazy_static! {
     // This is regex is inadequate, as iso<typeof foo`...`> is invalid, and
     // it's certainly possible to want that.
-    static ref EXTRACT_ISO_LITERAL: Regex = Regex::new(r"(export const ([^ ]+) =\s+)?iso(<[^`]+>)?`([^`]+)`(\()?").unwrap();
+    static ref EXTRACT_ISO_LITERAL: Regex = Regex::new(r"(export const ([^ ]+) =\s+)?iso\(?(<[^`]+>)?`([^`]+)`\)?(\()?").unwrap();
```
rbalicki2, line 78: Let's go ahead and make these parentheses required and update the pet and github demos accordingly.

crates/isograph_cli/src/generate_artifacts.rs
```diff
@@ -64,16 +64,123 @@ pub(crate) fn generate_and_write_artifacts(
...
+        content.push_str(&entrypoint_overload);
+    }
+    content.push_str(
+        "
+export function iso(_queryText: string): IdentityWithParam<any> {
+  // The name `identity` here is a bit of a double entendre.
+  // First, it is the identity function, constrained to operate
+  // on a very specific type. Thus, the value of b Declare`...`(
+  // someFunction) is someFunction. But furthermore, if one
+  // write b Declare`...` and passes no function, the resolver itself
```
rbalicki2, line 155: Lol this comment is pompous and outdated, and provides no value. Let's remove it lol

crates/isograph_cli/src/generate_artifacts.rs
```diff
@@ -64,16 +64,123 @@ pub(crate) fn generate_and_write_artifacts(
...
+TString extends string,
+T
+> = Whitespace<T> extends `${TString}${string}` ? T : never;",
+    );
+    let entrypoints_overload = schema
+        .entrypoints
+        .iter()
+        .map(|resolver_field_id| build_iso_overload_for_entrypoint(*resolver_field_id, schema));
+    let fields_overload = schema
+        .resolvers
```
rbalicki2, line 137: Oops — it looks like we're generating too many overloads. As a follow up, we should filter on `.filter(|resolver| matches!(resolver.action_kind, ResolverActionKind::NamedImport(_)))`. This is janky, but it is currently the way to distinguish between resolvers that are auto-generated (i.e. for auto generated `__refetch` fields) and client-defined ones (which are the only ones that need overloads)

demos/github-demo/src/isograph-components/UserRoute.tsx
```diff
@@ -1,17 +1,17 @@
-import React from 'react';
-import { iso, useLazyReference, useRead } from '@isograph/react';
-import { Container } from '@mui/material';
+import React from "react";
```
rbalicki2, line 1: if you do `Preferences: Open Workspace Settings (JSON)` and add `  "prettier.configPath": ".prettier.config.js"`, this formatting issue should be fixed.

I think VSCode doesn't like the fact that we're not using prettierrc. It's something I'll look into. But in the mean time, that seems to fix it. Or run `pnpm run format`

libs/isograph-babel-plugin/compileTag.js
```diff
@@ -1,42 +1,40 @@
...
 const typeAndFieldRegex = new RegExp(
-  '\\s*(entrypoint|field)\\s*([^\\.\\s]+)\\.([^\\s\\(]+)',
-  'm',
+  "\\s*(entrypoint|field)\\s*([^\\.\\s]+)\\.([^\\s\\(]+)",
+  "m"
 );
 
 function getTypeAndField(path) {
-  const quasis = path.node.quasi.quasis;
+  const quasis = path.node.arguments[0].quasis;
```
rbalicki2, line 30: This will probably throw if `iso()` is called by itself, i.e. if there is no argument

rbalicki2 (changes requested):
Just a few minor nits, overall looks amazing!! Let's get this cleaned up and shipped!!!!

crates/isograph_cli/src/generate_artifacts.rs
edmondop: Fixed with [`951b23a` (#12)](https://github.com/isographlabs/isograph/pull/12/commits/951b23a68cd520ff8d85981ebf3d7e2c280990cb)

crates/isograph_cli/src/generate_artifacts.rs
edmondop: Fixed with [`951b23a` (#12)](https://github.com/isographlabs/isograph/pull/12/commits/951b23a68cd520ff8d85981ebf3d7e2c280990cb)

libs/isograph-babel-plugin/compileTag.js
edmondop: This line here should protect from those problems

demos/github-demo/src/isograph-components/UserRoute.tsx
edmondop: Fixed with [`8abe8a5` (#12)](https://github.com/isographlabs/isograph/pull/12/commits/8abe8a5bc56b06b5154d46ea4253b1952627cc8c)

libs/isograph-babel-plugin/compileTag.js
rbalicki2: What if arguments exists, but is of length 0?

crates/isograph_cli/src/generate_artifacts.rs
```diff
@@ -64,16 +64,116 @@ pub(crate) fn generate_and_write_artifacts(
...
+
+type WhitespaceCharacter = ' ' | '\\n';
+type Whitespace<In> = In extends `${WhitespaceCharacter}${infer In}`
+? Whitespace<In>
+: In;
+
+type MatchesWhitespaceAndString<
+TString extends string,
+T
+> = Whitespace<T> extends `${TString}${string}` ? T : never;",
```
rbalicki2, line 130: let's add some line breaks here, otherwise it shows up as

```
> = Whitespace<T> extends `${TString}${string}` ? T : never;export function iso<T>(
```

libs/isograph-babel-plugin/compileTag.js
edmondop: I was so sure that the evaluation rules of javascript would treat an array of length 0 as falsy, but I just checked and I was horribly wrong :)

README.md
```diff
@@ -119,11 +119,11 @@ type Mutation
 In the above example, the `set_user_name` field will be made available on every `User` object, under the key `set_user_name` (this will be customizable.) So, one could write a resolver:
 
 ```js
-export const UpdateUserNameButton = iso`
+export const UpdateUserNameButton = iso(`
   User.UpdateUserNameButton {
     set_user_name,
   }
-`(({ data: { set_user_name } }) => {
+`)(({ data: { __set_user_name } }) => {
```
rbalicki2, line 126: this should be `set_user_name` after rebasing, magic mutation fields no longer have `__`.

libs/isograph-babel-plugin/compileTag.js
rbalicki2: `iso()` is always a mistake, yeah. We should throw an error here.

edmondop (commented):
I think this should fix it all [`6bc88d4` (#12)](https://github.com/isographlabs/isograph/pull/12/commits/6bc88d4a404c1649c16aab50b9b8658a5f41f778) hopefully

rbalicki2 (approved):
Aww yeah!!! Amazing!

Issue #10, opened 2024-02-02
Build Isograph for Windows


rbalicki2:
* in ci.yml, Windows is disabled:
```
          # - target: x86_64-pc-windows-msvc
          #   os: windows-latest
          #   build-name: isograph_cli.exe
          #   artifact-name: isograph_cli-bin-win-x64
```
* If this is enabled, we get the following error:
```
error: failed to get `intern` as a dependency of package `common_lang_types v0.0.4 (D:\a\isograph\isograph\crates\common_lang_types)`

Caused by:
  failed to load source for dependency `intern`

Caused by:
  Unable to update https://github.com/facebook/relay.git#f53a93272a1d756c60de795090fe70ca9532bb67

Caused by:
  path too long: 'C:/Users/runneradmin/.cargo/git/checkouts/relay-de0260293980c89f/bbddbf3/compiler/crates/relay-compiler/tests/compile_relay_artifacts/fixtures/abstract-type-refinement-dont-flatten-type-discriminator-nested-fragment-spread-within-inline-fragment-different-fields.expected'; class=Filesystem (30)
```

This might be easy to investigate and fix! The Relay compiler also uses intern, and the build step is very similar: https://github.com/facebook/relay/blob/main/.github/workflows/ci.yml

Thank you!

PR #21, opened 2024-02-15
fixed "path too long" error on windows ci workflow target

adrtivv:
Resolves issue #10

Checked the [relay's ci workflow config](https://github.com/facebook/relay/blob/main/.github/workflows/ci.yml), didn't find a similar solution, they're probably using some other way to implement the same fix.

Cargo internally uses [libgit2](https://github.com/libgit2/libgit2) for git related workflows, and the library has [open issues](https://github.com/libgit2/libgit2/issues?q=longpaths+) regarding this that haven't been fixed yet. 

More info can be found here:- https://github.com/rust-lang/cargo/issues/13020

**TL;DR**:- Windows sucks, but made it work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,10 +32,11 @@ jobs:
             os: macos-latest
             build-name: isograph_cli
             artifact-name: isograph_cli-macos-arm64
-          # - target: x86_64-pc-windows-msvc
-          #   os: windows-latest
-          #   build-name: isograph_cli.exe
-          #   artifact-name: isograph_cli-bin-win-x64
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
+            longpaths: true
+            build-name: isograph_cli.exe
+            artifact-name: isograph_cli-bin-win-x64
     runs-on: ${{ matrix.target.os }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


rbalicki2:
Love it!! Thank you! I'll review and merge tomorrow. Thank you so much for doing this. This will be part of the 0.1.1 release!

.github/workflows/ci.yml
```diff
@@ -32,10 +32,10 @@ jobs:
...
             build-name: isograph_cli
             artifact-name: isograph_cli-macos-arm64
-          # - target: x86_64-pc-windows-msvc
-          #   os: windows-latest
-          #   build-name: isograph_cli.exe
-          #   artifact-name: isograph_cli-bin-win-x64
+          - target: x86_64-pc-windows-msvc
+            os: windows-latest
+            build-name: isograph_cli.exe
+            artifact-name: isograph_cli-bin-win-x64
```
rbalicki2, line 38: Could you please add a field here (e.g. longpaths: true), as well as the same field for the other OS's, and not have a special case for windows below?

rbalicki2 (changes requested):
Love it!! Let's just use a separate field here (i.e. `longpaths: true`) instead of special casing it, and then let's get this landed.

Thank you so much for enabling Windows!!!

.github/workflows/ci.yml
adrtivv: added it similar to how it's done for the cargo **cross** compilation, it's not needed for UNIX based systems(linux, macos etc)

rbalicki2:
🎉  Love it, thank you!!

PR #38, opened 2024-03-27
Make `additionalProps` conditionally required

Andarist:
Note that I didn't check what you are trying to accomplish with this `Record<string, never>` etc. I also didn't have time to even recheck if this compiles right now - on the spot I couldn't find a clear instruction on how to build all of the packages here. In a sense - this is just a quick driveby PR 😉

### DIFF
--- a/libs/isograph-react/src/EntrypointReader.tsx
+++ b/libs/isograph-react/src/EntrypointReader.tsx
@@ -1,23 +1,27 @@
-import * as React from 'react';
+import * as React from "react";
 import {
   ExtractReadFromStore,
   IsographEntrypoint,
   type FragmentReference,
   useResult,
-} from './index';
+} from "./index";
 
 export function EntrypointReader<
   TProps extends Record<any, any>,
   TEntrypoint extends IsographEntrypoint<any, React.FC<TProps>>,
->(props: {
-  queryReference: FragmentReference<
-    ExtractReadFromStore<TEntrypoint>,
-    React.FC<TProps>
-  >;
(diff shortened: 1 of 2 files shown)


libs/isograph-react/src/EntrypointReader.tsx
```diff
@@ -1,23 +1,27 @@
-import * as React from 'react';
+import * as React from "react";
```
Andarist, line 1: this was an unintended change but it seems that it's how your Prettier is configured here

libs/isograph-react/src/EntrypointReader.tsx
```diff
@@ -1,23 +1,27 @@
...
+    queryReference: FragmentReference<
+      ExtractReadFromStore<TEntrypoint>,
+      React.FC<TProps>
+    >;
+    additionalProps?: TProps;
+  } & (TProps extends Record<string, never>
+    ? {}
+    : {
+        additionalProps: unknown;
+      }),
```
Andarist, line 23: the idea here is that `{ prop?: Whatever } & { prop: unknown }` is basically equivalent of `{ prop: Whatever }`. So by creating an intersection like this we can easily~ make some "preceding" optional properties required

rbalicki2:
Thanks for the contribution! This inspired the fix seen here: [4b202045d5e114b6fb11938b0fd561e254415f9d](https://github.com/isographlabs/isograph/commit/90264a004d75f52bc985e2c67889b83db438c9db)

Issue #24, opened 2024-02-17
Numeric literals are generated as strings in normalization ASTs


rbalicki2:
```
      pullRequests(last: 1) {
        totalCount
      }
```

results in

```
      {
        kind: "Linked",
        fieldName: "pullRequests",
        alias: null,
        arguments: [
          [
            "last",
            { kind: "Literal", value: "1" },
          ],
        ],
        selections: [
          {
            kind: "Scalar",
            fieldName: "totalCount",
            alias: null,
            arguments: null,
          },
        ],
      },
```

This maybe isn't broken, because the literal value is stringified anyway, but it probably should be generated as a number in the AST.

ismajl-ramadani:
Can I work on this?

rbalicki2:
Yes, absolutely! Let me know whether you would like to touch base, happy to walk you through what roughly needs to happen and get you acquainted with the codebase. The base place to reach out is on the discord https://discord.gg/kDCcN3EDR6 (the #isograph channel on the GraphQL community discord)

rbalicki2:
Thank you!!

Yes, I think that should do it. For some manual testing, please make sure:

* the generated files look correct. Run `pnpm watch-rs` in one terminal, and `pnpm build-demos` in another once it has compiled
* this looks like it will just affect the generated files for the GitHub demo, so click around the github demo, and make sure it works. (I did a quick pass to make sure it works, it seems to.) The GitHub demo can be annoying to set up (I think), so alternatively
* change PetDetailRoute to have

```
  field Query.PetDetailRoute($id: ID!) @component {
    foo: pet(id: $id) {
      id
    }
    pet(id: 0) {
      name
      PetCheckinsCard
      PetBestFriendCard
      PetPhraseCard
      PetTaglineCard
    }
  }
```

and make sure that you can click on the first pet (Makayla) and see data.

If this all looks good to you, we can get this landed!

ismajl-ramadani:
Thanks for the additional help. I've tested it with the pet-demo as you suggested and it seems to work. I created a PR and added the small fix and also generated artifacts for the github-demo. Other demos looks like they don't have any integer literal, so they should be fine. 

Please let me know if I have missed something.

PR #41, opened 2024-04-06
Numeric literals as strings generated in ast fix

ismajl-ramadani:


rbalicki2:
LGTM! Closes #24 thank you so much for making this fix

Issue #28, opened 2024-02-21
We should split reader.ts into several artifacts


rbalicki2:
## Problem

* Currently, reader artifacts contain: the generated TypeScript `__param` type, the generated `__outputType` type, and the reader AST.
* If a resolver is unreachable from an entrypoint, we do not need to generate the reader AST. But we do need to generate the types. So, we generate unneeded reader ASTs.
* If the resolver is never selected by another resolver **and** is not an entrypoint, we do not need to generate the output type. So, we generate some unneeded output types.
* However, we need to define a `__param` type for all user-defined resolvers, because it is used by `iso.ts`. We currently generate unneeded `__param` types for `__refetch` artifacts (though, seemingly not for magic mutation fields.)

**Note**: the output and param types that we generate that are not needed are not as big of a deal as the unneeded reader AST, because:
* output types are easy to create
* `__refetch` fields are rarer
So, if not generating the output and param types adds complexity, we can punt on it.

## Solution

* So, instead, we should split reader.ts into reader.ts (containing the reader AST and reader artifact), `param_type.ts` and `output_type.ts`.

## Details

* `Type/field/param_type.ts` should import the `output_type` of all selected resolvers and export the param type
* `Type/field/output_type.ts` should contain the `ResolverReadOutType` (e.g. `export type Query__HomeRoute__outputType = (React.FC<ExtractSecondParam<typeof resolver>>);`) and nothing else.
* `Type/field/reader.ts` should define the reader AST. It should import the user-defined exported resolver function (i.e. `export const pet_stats_card = iso(...)`). It should import the param and output type. It should define and export the artifact.

## Follow ups:

* We should only generate types, and not reader ASTs, for unreachable client fields
* Maybe this has some implication for __refetch fields. That seems to work correctly now, but it would be worth double checking whether we can simplify things.
* I'm hoping that this results in cleaner code, which will make it easier to do things like asynchronously import JavaScript (see #33)

ismajl-ramadani:
Can I take this?

PR #52, opened 2024-04-10
Split render into several artifacts #28

ismajl-ramadani:
WIP


rbalicki2:
Thank you for starting this work!! I'll go ahead and make sure this lands

PR #87, opened 2024-06-09
Splitting Compiler from CLI to support LSP

edmondop:


rbalicki2 (approved):
Love it! I'll land! Thank you!!!

PR #89, opened 2024-07-22
chore: Make rust clippy happy

Xuanwo:
This PR will make rust clippy happy. Most changes will make code easier to read or can avoid extra alloc.

To avoid further, we can introduce clippy checks in CI.


crates/isograph_lang_parser/src/peekable_lexer.rs
```diff
@@ -42,6 +42,9 @@ impl<'source> PeekableLexer<'source> {
     /// Get the next token (and advance)
     pub fn parse_token(&mut self) -> WithSpan<IsographLangTokenKind> {
         // Skip over (and record) any invalid tokens until either a valid token or an EOF is encountered
+        //
+        // Remove this allow after logic changed.
+        #[allow(clippy::never_loop)]
```
rbalicki2, line 47: Do you understand why clippy is complaining here?

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -503,6 +501,7 @@ derive_display!(ReaderAst);
 pub(crate) struct NormalizationAstText(pub String);
 derive_display!(NormalizationAstText);
 
+#[allow(dead_code)]
```
rbalicki2, line 504: You can go ahead and remove this struct entirely! Thanks for finding this!

relay-crates/common/src/diagnostic_check.rs
```diff
@@ -156,7 +156,7 @@ mod escalate_tests {
 
         escalate(DiagnosticSeverity::WARNING, &mut diagnostics);
```
rbalicki2, line 157: The relay-crates folder contains forks of crates that actually exist in https://github.com/facebook/relay; if there's a good way to disable clippy for these folders, that would be better than modifying them (as I don't expect to modify the files in this folder at all.)

But if that's not easy, we can merge these changes no problem

You also may want to make these changes upstream to Relay! I'm sure they'd appreciate it, or (maybe) have some clippy config that causes this code to pass.

crates/isograph_lang_parser/src/peekable_lexer.rs
Xuanwo: Since this block returns directly which leads to `never_loop`. For example:

```rust
pub fn parse_token(&mut self) -> WithSpan<IsographLangTokenKind> {
    loop {
        let kind = self
            .lexer
            .next()
            .unwrap_or(IsographLangTokenKind::EndOfFile);
        match kind {
            IsographLangTokenKind::Error => {
                panic!() // panic happened here.
            }
            _ => {
                // return happened here.
                return abc;
            }
        }
    }
}
```

I've left `allow` here, assuming you'll add more logic that requires the loop.

crates/graphql_artifact_generation/src/query_text.rs
```diff
@@ -65,6 +65,27 @@ fn write_variables_to_string<'a>(
     }
 }
 
+/// TODO: Please double check if `schema` really needed.
```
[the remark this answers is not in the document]
Xuanwo: Fixed.

crates/graphql_artifact_generation/src/generate_artifacts.rs
Xuanwo: Removed.

relay-crates/common/src/diagnostic_check.rs
Xuanwo: Got it, let's disable them.

crates/isograph_lang_parser/src/peekable_lexer.rs
rbalicki2: Ah I see, makes sense. Thanks for highlighting this, yes, I believe we want the loop there. The Relay code does stuff in that branch, but it's not a high priority for me to recover from errors at the moment: https://github.com/facebook/relay/blob/a1ce4999fd68e7bcc66d542be5d91daa1a602356/compiler/crates/graphql-syntax/src/parser.rs#L2901-L2923

crates/graphql_artifact_generation/src/query_text.rs
```diff
@@ -114,16 +135,16 @@ fn write_selections_for_query_text<'a>(
                     inline_fragment.selection_map.values(),
                     indentation_level + 1,
                 );
-                query_text.push_str(&format!("{}", "  ".repeat(indentation_level as usize)));
-                query_text.push_str(&format!("}},\\\n"))
+                query_text.push_str(&"  ".repeat(indentation_level as usize).to_string());
+                query_text.push_str("}},\\\n")
```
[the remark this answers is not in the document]
rbalicki2: Oh I mean, this changes behavior (and thus fails CI), since `format!("}}")` is requivalent to `"}".to_string()`

rbalicki2 (changes requested):
Just one change requested (`}}` to `}`), then we're g2g!

crates/graphql_artifact_generation/src/query_text.rs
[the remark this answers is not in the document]
Xuanwo: Oh, sorry for my misunderstanding. Now it fixed.

rbalicki2:
Shipped!!!

Issue #56, opened 2024-04-14
Figure out how to use Babel/SWC plugins with Vitest


rbalicki2:
* The tests in isograph-react (which will land soon) will use iso literals. But no babel plugin is configured, so we have to import the entrypoint and use that directly. Not ideal!

PR #93, opened 2024-09-13
Use babel plugin with vitest

PatrykWalach:
Used babel plugin in `isograph-react` 
Moved `isograph.config.json` up for babel plugin to find it
Added `vite-plugin-commonjs`, because `require()` doesn't work in vite 

Closes Issue #56


rbalicki2:
Zajebiste! Dzięki za to. Commenting to enable the workflow run. (I think that's how it works.) If I don't finish reviewing it today, I should be able to get to it by at latest Wednesday. But overall, looks solid, and as long as nothing is obviously broken, we can land it

rbalicki2:
Oh, there's actually a button that I click to enable the workflow run. TIL

rbalicki2:
Please ensure that `pnpm build-isograph-react-demo` works, and also run `pnpm i` in the root. I think you also probably want to move the `schema.graphql` file.

At that point, as long as `pnpm -r test` still passes, we should be g2g!

PatrykWalach:
I've checked that `pnpm build-isograph-react-demo` and `pnpm -r test` passed locally.


I had to rerun `pnpm i` with pnpm@8 because I've used a newer version by mistake. If you would like I could open a PR to include `"packageManager": "pnpm@8.15.4"` in root `package.json`. This way everyone contributing would use correct version automatically. The `ci.yml` wouldn't require specifying the versions, so there would be one place to update when upgrading `pnpm`. It would also crash in places where `yarn` is still being used, I assume these commands should be replaced anyways.

rbalicki2:
Awesome! I'm going to merge this. Thank you so much for getting this done.

I'm going to revert the change to add the `.exe` in package.json. I haven't tested Isograph on a windows machine, but it looks like we should have some way to run either `isograph_cli` or `isograph_cli.exe`, depending on what exists.

Re: `pnpm@8`. Presumably, that doesn't prevent someone who installs `@isograph/compiler` in their project from running `yarn run iso`, right? It's just for commands that are run when developing? If so, then by all means, that sounds great! Thank you!!

PR #94, opened 2024-09-15
Specify pnpm version in `package.json`

PatrykWalach:
> Re: `pnpm@8`. Presumably, that doesn't prevent someone who installs `@isograph/compiler` in their project from running `yarn run iso`, right? It's just for commands that are run when developing? If so, then by all means, that sounds great! Thank you!!

Yes, it doesn't prevent anyone from using different package manager in their project


rbalicki2 (approved):
Love it, merging

PR #95, opened 2024-09-18
fix windows generated paths

PatrykWalach:


rbalicki2:
Nice. Feel free to do a runtime check in the babel extension to see if you're in windows, and in Rust, let's do something like

https://stackoverflow.com/a/47805781/2338813

rbalicki2:
LFGGGG

PR #104, opened 2024-09-20
Fix refetch on variables changes

### LINKED ISSUE #96: `useLazyReference` does not refetch when props change (post commit)
* `useLazyReference` correctly refetches when props change pre-commit. This is because `useLazyDisposableState` gets passed a new (presumably empty) cache. When it renders with an empty cache pre-commit, it will make a network request.
    * the pre-commit behavior is good
* post-commit, we need to refetch if we are passed new variables
* this can be done, I believe, in `useLazyDisposableState`.
    * However, there's a slight chance that it should appear in a function that is essentially a clone of `useLazyDisposableState` + the added behavior.

anyway, I'll flesh this out once I think about specifics.

PatrykWalach:
My line of thinking was that in `useCachedPrecommitValue` there are three cases for reading from cache, one of these is for pre-commit, so I've just copied the other two to read on cache change dad5d343eea9281a614c662d3607b72f661ff84e.
This seemed to work pretty well, but then I've remember what you told me about cleaning up, and noticed there's a chance `itemCleanupPairRef` wouldn't be disposed so I've replaced it with `state` from `useUpdatableDisposableState()`, this let me remove all the clean up code 83344ab37ae7fb5e4139192814a9b7826bb4e970.

My concern here is that now `useLazyDisposableState` is pretty similar to `useDisposableState`.

Do I need any more tests for this?
And should I expose the `refetch` in `useLazyReference`?

Closes #96


rbalicki2:
Great work!!! Thank you for tackling this. I realize when we talked about this I had the wrong understanding of what needed to be done and wasn't very clear on my end. Sorry!

So, I think we should expect that this will take a few back and forths.

As a result, feel free to make the changes to the actual file before spending too much time on the test file. Or feel free to get reviews while the test file is broken.

## We should not refetch during a commit

We should not update this during a commit. We should instead kick off the fetch during render.

- this is more performant (in general)
- if another component suspends during that render, we will not commit this hook until after that suspense completes, so fetching in an effect is potentially far less efficient
- there will be one commit where the render variables are `variable_2` but the network request is `network_request(variables_1)`
  - this is even worse than it sounds; if as a result of the state change to `network_request(variables_2)`, the tree suspends, _and_ that is wrapped in a `useTransition`, we could potentially show the inconsistent state for a while.

So, given this, we should refetch during render.

## How `useLazyDisposableState` current works

Let's describe how `useLazyDisposableState` works, because I think we can make small changes to this in order to enable the actual behavior we want:

- `useLazyDisposableState` can be in one two states: pre-commit or post-commit. This state is stored in `useHasCommittedRef` in `useCachedPreCommitValue`. Call this the **inner state**.
- In the pre-commit inner state, `useLazyDisposableState` returns `preCommitItem?.state`,
- during the initial commit, we assign that item to a ref (`itemCleanupPairRef`) and the inner state transitions to post-commit.
- In the post commit state, `useLazyDisposableState` returns `itemCleanupPairRef.current?.[0]`
- when `useLazyDisposableState` unmounts, it expects `itemCleanupPairRef` to not be empty and disposes of the item.

## Step 1: changes to `useCachedPrecommitValue`

The goal of step 1 is this: in `useCachedPrecommitValue`, whenever we encounter a new parent cache, to notice and reset the inner state to pre-commit. On the next commit, behave as previously described.

However, because we're going to have to change two callsites, it might be easier to first create a new hook `useCachedResponsivePrecommitValue`, which is based on `useCachedPrecommitValue`, and then update the callsites appropriately.

So, in `useCachedResponsivePrecommitValue`, there's no more need to call `useHasCommittedRef`, instead define a ref `lastCommittedParentCache` with type `null | ParentCache`, initialized to `null`.

We have three possible states:

- pre-initial commit (`lastCommittedParentCache.current === null`)
- post-initial commit, parent cache unchanged (`lastCommittedParentCache.current === parentCacheFromProps`)
- post-initial commit, parent cache has changed (`lastCommittedParentCache.current !== parentCacheFromProps`)

Note: Pre-initial commit, the hook behaves correctly! As in, if we re-render with new variables, we will correctly refetch. Pre-commit behavior should not change.

> `parentCacheFromProps` should still be called `parentCache`; calling it `parentCacheFromProps` is just to make this explanation clearer

We may choose to remove the non-responsive version last. (But TBH it seems fair to keep it.)

### Render behavior is similar:

- if `lastCommittedParentCache.current === parentCacheFromProps`, then return null.
- if `lastCommittedParentCache.current === null` or `lastCommittedParentCache.parentCache !== parentCacheFromProps`, then behavior is unchanged:

```
  // Safety: item is only safe to use (i.e. guaranteed not to have been disposed)
  // during this tick.
  const [cacheItem, item, disposeOfTemporaryRetain] =
    parentCache.getOrPopulateAndTemporaryRetain();

  return { state: item };
```

### Effect behavior

- the dependency array should contain `parentCacheFromProps`
- add `lastCommittedParentCache.current = parentCacheFromProps`
- otherwise, unchanged
- update doc comments and function name

## Step 2: Changes to `useLazyDisposableState`

Now, the function that is the second parameter to `useCachedResponsivePrecommitValue` can be called multiple times, so we have to account for that.

### Render behavior

- unchanged, if we receive a "preCommitItem" from `useCachedResponsivePrecommitValue` (even though the name "preCommitItem" is inaccurate), we should prefer that. And if we haven't received that, we will have called the `onCommit` function at least once.

### `onCommit` (second parameter to `useCachedResponsivePrecommitValue`) behavior

Summary of thinking: `onCommit` will be called at most once per commit (duh). Once a component commits, the renders associated with previous commits will never be shown to the user (source: I manually tested this many years ago, and it may change.)

So, when we call `onCommit`, we can clean up everything from previous commits. So, in the `onCommit`:

- dispose of `itemCleanupPairRef.current`, if it exists
- `itemCleanupPairRef.current = pair`

### Effect behavior

The same, except the cleanup function must close over `itemCleanupPairRef`, not `itemCleanupPairRef.current`, because otherwise we would try to re-dispose of the pair created during the initial commit, and not the last-created pair.

### Testing

Besides unit testing, we should be able to manually test. I'll add that as a follow up!

## Changes to `useDisposableState`

Let's punt on this for now, let's see how this goes and reassess.

PatrykWalach:
If tests pass, this should be good to go

rbalicki2:
Hell yeah!!! Love it.

The video shows some manual testing that shows that it is fixed. At first, we're on main, and we don't see the UI update, even though the URL changes. On your branch, the UI does changes.

This is because the PetDetailRoute is never unmounted, so on main we run into the issue that `useLazyReference` continues to show the old content.

But anyway, IMO it's not great that requires us to worry about this! This is not an issue within Isograph apps — the pet with id 0 and the pet with id 1 have different components (in the `===` sense), so this would not happen. This only occurs on the outside, in regular React-land.

https://github.com/user-attachments/assets/b310d3db-a095-4fb8-a91c-f872e65e0894

Issue #68, opened 2024-04-16
CLI command to create a new isograph config


rbalicki2:
Also we should update the quick start accordingly

lanesawyer:
I've got experience with `clap` and am interested in this issue.

I noticed you're using `structop` but that project was essentially consumed by `clap` a while back and the project is in maintenance mode. Any opposition to moving to `clap` as part of this CLI improvement, or are you deliberately avoiding the extra overhead (and features) of `clap`?

rbalicki2:
* 100%, happy to migrate to clap! There is no concrete reason we're using `structopt`.
* Love it! Let's hop on a call at some point in time and discuss details. Can you join the #isograph channel on the discord, where we can figure out a good time (https://discord.gg/kDCcN3EDR6)?

rbalicki2:
The gist of this (poorly written) issue is:

* I would like to add a subcommand, along the lines of https://nextjs.org/docs/pages/api-reference/create-next-app that asks you several things. We can provide default values (or lack of values) for the other Isograph config items:
     * where is the graphql schema,
     * where is the folder where Isograph components are located
     * should we modify the tsconfig.json (to add `"@iso": ["./src/components/__isograph/iso.ts"],`)
     * should we create or modify `.babelrc.json`?
* this CLI command should create the `isograph.config.json` and optionally modify the other files.

lanesawyer:
Starting to look into this now that the `clap` migration is done!

So it would be something like this:

`isograph_cli create-app [project name]`

With at least these options and their behaviors:
- `-h or --help`: Describes the available commands
  - Comes for "free" with the CLI library
- `-s or --schema`: Location of their GraphQL schema
  - Required
  - If file not found, log error message encouraging them to create the schema file
  - If file not found, does _not_ prevent the process from completing successfully
- `-c or --components`: Location of the Isograph components
  - Not required
  - Makes the folder if it doesn't already exist
  - Default to `./src/components` if not provided
- `--import-alias`: Specifies the import alias they want to use
  - Not required
  - Defaults to `@iso/`
  - Logs error message if `tsconfig.json` from `--tsconfig` is not found
  - Mutually exclusive with `--no-import-alias`
- `--no-import-alias`: Don't modify their `tsconfig.json` file with an import alias
  - Not required
  - Mutually exclusive with `--import-alias`
- `--tsconfig`: Location of `tsconfig.json`
  - Not required
  - Default to current working directory's `tsconfig.json`
[...]

PR #106, opened 2024-09-21
chore(cli): structop to clap migration

lanesawyer:
# `structopt` to `clap` Migration

**Warning: I can't run Isograph locally because I'm on Linux and there are some issues getting that working, so this could use testing to make sure the commands actually still works!**

But the help command outputs very similar stuff, and it compiles, which is a very good sign when working with Rust 😁 

## What
- Migrates CLI from the `stuctopt` crate (which is deprecated) to `clap` in preparation for work in #68
- Renames the `Command` enum to `Commands` to align with `clap` conventions

## How
- Followed migration guides:
  - For `structopt` to `clap` v3: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrate-structopt
  - For `clap` v3 to v4: https://github.com/clap-rs/clap/blob/master/CHANGELOG.md#migrating

### Before
image

### After
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -38,15 +38,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ansi_term"
-version = "0.12.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
-dependencies = [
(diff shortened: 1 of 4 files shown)


rbalicki2:
LGTM!!! Thank you for taking this on!

Most of the other enums are named singularly (e.g. `enum DeserializationError` not `enum DeserializationErrors`) but I can just revert that name change in main

Issue #100, opened 2024-09-20
Bug: useSkipLimitPagination doesn't subscribe to changes


rbalicki2:
* changes to fields that are only selected as part of the pagination query probably are not subscribed to
* it needs to call `useSubscribeToMultiple`

PR #108, opened 2024-09-22
fix useSkipLimitPagination doesn't subscribe to changes

PatrykWalach:
When I was working on #107 and noticed `useSkipLimitPagination` I tried adding a subscription. Would love to know if this code would be a good start for issue #100.

It seems to work fine in pet demo.

There's still stuff to refactor, this check needs to moved.
```tsx
if (networkRequestStatus === null) {
    throw new Error(
      'FragmentReference is unexpectedly disposed. \
      This is indicative of a bug in Isograph.',
    );
  }
  ```

In short, what I've done:
- raised array slicing up so I can call `useSubscribeToMultiple` with completed fragment references before the if statement.
- changed `readCompletedFragmentReferences` so it returns subscriptions and sets data into `readOutDataAndRecordsRef`.

### DIFF
--- a/libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts
+++ b/libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts
@@ -1,9 +1,12 @@
-import { LoadableField } from '../core/reader';
+import { LoadableField, type ReaderAst } from '../core/reader';
 import { useIsographEnvironment } from '../react/IsographEnvironmentProvider';
 import { ItemCleanupPair } from '@isograph/disposable-types';
 import { FragmentReference } from '../core/FragmentReference';
 import { maybeUnwrapNetworkRequest } from '../react/useResult';
-import { readButDoNotEvaluate } from '../core/read';
+import {
+  readButDoNotEvaluate,
+  type WithEncounteredRecords,
+} from '../core/read';
 import {
   UNASSIGNED_STATE,
   useUpdatableDisposableState,
@@ -12,7 +15,9 @@ import {
   createReferenceCountedPointer,
   ReferenceCountedPointer,
 } from '@isograph/reference-counted-pointer';
+import { useRef, useState } from 'react';
 import { getPromiseState, readPromise } from '../core/PromiseWrapper';
+import { useSubscribeToMultiple } from '../react/useReadAndSubscribe';
 
 type SkipOrLimit = 'skip' | 'limit';
 type OmitSkipLimit<TArgs> = keyof Omit<TArgs, SkipOrLimit> extends never
@@ -36,8 +41,12 @@ type ArrayFragmentReference<TItem> = FragmentReference<
   ReadonlyArray<TItem>
 >;
 
+type LoadedFragmentReference<TItem> = ItemCleanupPair<
+  ReferenceCountedPointer<ArrayFragmentReference<TItem>>
+>;
+
 type LoadedFragmentReferences<TItem> = ReadonlyArray<
-  ItemCleanupPair<ReferenceCountedPointer<ArrayFragmentReference<TItem>>>
+  LoadedFragmentReference<TItem>
 >;
 
 function flatten<T>(arr: ReadonlyArray<ReadonlyArray<T>>): ReadonlyArray<T> {
@@ -50,17 +59,12 @@ function flatten<T>(arr: ReadonlyArray<ReadonlyArray<T>>): ReadonlyArray<T> {
   return outArray;
 }
 
-/**
- * NOTE: this hook does not subscribe to changes. This is a known
- * issue. If you are running into this issue, reach out on GitHub/
- * Twitter, and we'll fix the issue.
- */
 export function useSkipLimitPagination<
   TArgs extends {
     skip: number | void | null;
     limit: number | void | null;
   },
-  TItem,
+  TItem extends object,
 >(
   loadableField: LoadableField<TArgs, Array<TItem>>,
 ): UseSkipLimitReturnValue<TArgs, TItem> {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by rbalicki2** [comment]
Awesome! I should be able to take a look today

File: libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts [hunk 1]
```diff
@@ -75,44 +79,74 @@ export function useSkipLimitPagination<
...
+          records: readOutDataAndRecordsRef.current[i].records,
+          callback(data) {
+            readOutDataAndRecordsRef.current[i] = {
+              records: data,
+              results: readerWithRefetchQueries.readerArtifact.resolver(
+                data.item,
+                undefined,
+              ) as ReadonlyArray<any>,
+            };
+            rerender({});
```
rbalicki2, line 143: This isn't ideal. I need to think about how to solve this, but we should probably use state (instead of a ref + state) because this will lead to tearing (i.e. rendering inconsistent data) in the following case:

- a user is visible in a `List` component, which renders a `UserDetail` component. In `List`, we use `useSkipLimitPagination`. `UserDetail` is an `@component`, i.e. it goes through `componentCache`, so it calls `useReadAndSubscribe`.
- `useSkipLimitPagination` uses a ref + state
- `useReadAndSubscribe` calls `setState(data)` in the callback, i.e. does not have a ref
- We're initially rendering consistent data
- We receive an update, `v1`. That updates the ref in `List` and schedules a re-render, and calls `setState(data)` in `UserDetail`
- The user then types some characters in a textbox in `List` (this tearing can occur if the user types in other components, too). React will prioritize responding to the user input more highly than processing the state changes.
- So, `List` (and `UserDetail`) are re-rendered. `List` reads from the ref, showing the updated data. `UserDetail` reads from state, so shows the stale data!

rbalicki2: Does this happen with `useLazyDisposableState`? I don't think so.

Fact: If a render results in a commit, no previous render will ever commit (and thus be written to the DOM). So, we only have to worry about subsequent renders. So, we can see:

- first time rendered with `cache1`, use `preCommitValue`
- commit, ref is set to `cache1`
- subsequent render with cache2, this might suspend and be part of a transition. We make a network request. Or for whatever reason, React responds to this render very slowly.
- subsequent render with cache1. This is okay, the ref still contains `cache1`.
- the render with `cache2` commits, ref is updated to contain `cache2`
- at this point, we will never see any "stale" renders with `cache1`, so we won't accidentally make any network requests.

So I think it's safe! But I could be making mistakes.

libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts
```diff
@@ -152,47 +174,93 @@ export function useSkipLimitPagination<
         },
       ];
 
+      Promise.all([
```
PatrykWalach, line 177: `getFetchMore` is called only once for every result, so it makes sense to "await"  the initial result here, then push it to `readOutDataAndRecords` array. This fixes problem introduced with changing `areObjectEqualWithDeepEquality`.

Then we can subscribe all the `readOutDataAndRecords`, we don't have to slice `loadedReferences`.

Does this makes sense?

rbalicki2:
😭 looks like your changes to correctly only re-render modified components surfaced a bug in pagination, i.e. useSkipLimitPagination doesn't re-render. LOL 😭  I guess I'll prioritize this

libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts
rbalicki2: As much as possible, it's best to avoid calling `setState` any time except synchronously in response to a user event. i.e. we should synchronously call `setState` when `fetchMore` is called.

Calling `setState` in a callback makes it really easy for bugs (race conditions) to sneak in.

So, as much as possible, we should limit that to framework code: stuff like `PromiseWrapper`, Isograph store subscriptions, etc. should handle all of the async setting of state. And these should either be simple enough to not need testing, or (eventually) be well tested.

That will also make it easier to avoid race conditions when we e.g. add error handling, automatic refetching, etc.

What you have here (`cloned = [...original]; cloned[i] = new`) looks correct.

------

So, we should call `setState` in the same tick that `fetchMore` is called and "unwrap" during render.

Naively unwrapping all of the fragment references gives us a suspenseful pagination hook (see [this](https://github.com/isographlabs/isograph/commit/7fbb7f13c59ad7377581bd3c9088d08cbdf583c0#diff-94897d429e727e10ebdebf89fb27b8bbe95084c6bdd41a91c501c28e6a29a6d9R76-R98))

The extra work required to return a disciminated instead of suspending is annoying! I'm not wed to it. I think once we finish the `Readable` PR, we can dump all of that complexity to the user! The user will just call `readAllInArray` or `readAllButLastOneInArray` or whatever.

PatrykWalach: How about this to fix the rerender https://github.com/isographlabs/isograph/compare/main...PatrykWalach:fix-skip-limit-does-not-rerender? This simplifies the hook quite a bit. Could be landed as separate PR.

PatrykWalach: Using ".then" was a quick and dirty way to do it, the better way was `startTransition` so I've tried it here https://github.com/isographlabs/isograph/compare/main...PatrykWalach:cc70dc0a530b7fbefb9ed95250f58f325974c229, and it worked well, but the `pendingFragment` was missing, so I used `useDeferredValue` instead.

rbalicki2:
Amazing work!! I'm super excited that we're tackling this. Some thoughts:

* this throws due to the fact that `readOutDataAndRecords.length < completedFragmentReference.length` here being , so `readCompletedFragmentReferences` throws on `readOutDataAndRecords[i].item`.
* As I see it, the fundamental issue is that we have two pieces of state, and it's easy for them to go out sync (or for bugs to sneak in.) I'm especially concerned about bugs sneaking in as other people make changes.
* So let's start by doing as much as we can with a single piece of state.

Let's split this into two steps. The first we can land now, and the second will require some thinking.

## First, during render always call `readButDoNotEvaluate`

* keep a single piece of state (i.e. `useUpdatableDisposableState<      LoadedFragmentReferences<TReadFromStore, TItem>    >();`)
* in render, call 
```
  const readOutFragmentReferences = completedFragmentReferences.map(
    (fragmentReference) => ({
      fragmentReference,
      records: readButDoNotEvaluate(
        environment,
        fragmentReference,
        networkRequestOptions,
      ),
    }),
  );
```
* pass that to `subscribeCompletedFragmentReferences` and `readCompletedFragmentReferences`
* in the subscribe callback, call `rerender({})`, i.e. throw away the read out values that we get from subscribe
* at this point, let's ship what we have

## Second, figure out how to re-use the values from the callback

* this can be a separate step! I think it's doable, but it seems hard.
* In this step, we figure out how to also keep the `WithEncounteredRecords<TReadFromStore>` in that same piece of state. Maybe something like:
```
// state 1
    useUpdatableDisposableState<
      { fragmentReference: ArrayFragmentReference<TReadFromStore, TItem>, records: WithEncounteredRecords<TReadFromStore> }[]
    >();
```
* However, we may not have the `WithEncounteredRecords` for the pending item (as we would suspend when calling `readButDoNotEvaluate`)! My second thought is to model the state as:
```
// state 2
{
  completed: { fragmentReference: ArrayFragmentReference<TReadFromStore, TItem>, records: WithEncounteredRecords<TReadFromStore> }[],
  pending: ArrayFragmentReference<...> | null,
}
```
* But I'm not convinced that will work, as that does require a setState in a callback... so this second step requires some more thinking.

## How can we get something like state1 to work?

* My hope is that state1 can be made to work, somehow, and I think the way to do it will be to call `readButDoNotEvaluate` during render on the last item, when it goes from pending -> not. This would have to behave in a `useLazyDisposableState`-kind of way.
* Anyway, hence we should start by shipping step 1 😂

rbalicki2:
This looks good to me! I just added a commit to master that modifies the a list, and it turns out that this doesn't correctly re-render, which is surprising. I'll investigate, but AFAICT this PR looks great and can land

rbalicki2:
It works!! 🙌 , my testing setup was just wrong.

PR #110, opened 2024-09-22
added config and docs for node/pnpm setup using fnm/corepack 

adrtivv:
I didn't add the docs for automating some redundant commands when using fnm. Basically when configuring fnm for our shells we can pass some flags to automate commands like `fnm install`, `fnm use`, `corepack enable` etc. Fnm detects the fields `engines.node` and `packageManager` in `package.json` and automatically executes all that for us so that we can start using `node` and `pnpm` commands as soon as we change directory into the project.

Fnm doesn't enable all those [flags](https://github.com/Schniz/fnm/blob/master/docs/configuration.md) by default for us so we have to do it manually. Here's what we need to add to our `~/.bashrc` on unix based systems to make that work(notice the `$USER` shell environment variable, you can replace it with your user name):
```bash
FNM_PATH="/home/$USER/.local/share/fnm"
if [ -d "$FNM_PATH" ]; then
  export PATH="$FNM_PATH:$PATH"
  eval "$(fnm env --corepack-enabled --resolve-engines --use-on-cd --version-file-strategy=recursive)"
fi
```
The setup should be similar(ish) on other shells and windows/macos. You can add all this in the isograph docs if you want to. Other than that fnm commands are similar to nvm.


rbalicki2:
LGTM. Two questions:
- let's use `yarn` in the vscode-extension folder. There's a yarn.lock there. (Or switch to pnpm)
- When I run `fnm use` I see: `error: Can't find version in dotfiles. Please provide a version manually to the command.`. Is this expected? We don't have a `.node-version` file. If it's expected, can we document that?

adrtivv:
> When I run fnm use I see: error: Can't find version in dotfiles. Please provide a version manually to the command

Yeah it seems we need to append `--resolve-engines` flag to both `fnm install` and `fnm use`.

> It looks like that's because I changed my bash_profile to include eval "$(fnm env --use-on-cd --shell bash)" (as suggested by fnm documentation)

fnm documentation says to do that in `~/.bashrc` not in `~/.bash_profile`. you use macos do you not? macos uses zsh by default so you'll need to do it in your `~/.zshrc`

> This project's package.json defines "packageManager": "yarn@pnpm@9.11.0". However the current global version of Yarn is 1.22.21.

that's weird? there's no `yarn@pnpm@9.11.0` in `package.json`, there's only `pnpm@9.11.0`

> In particular yarn@pnpm@9.11.0 seems sus. Should it just be pnpm@9.11.0?

yeah it should be

rbalicki2:
Okay awesome, thanks for the clarifications! This should be good to land. Thank you @saminoorsyed  for running through these as well

Re- yarn or npm in vscode extension. The instructions say to yarn in there, so after i land I'll just update the package JSON to use yarn

adrtivv:
> Re- yarn or npm in vscode extension. The instructions say to yarn in there, so after i land I'll just update the package JSON to use yarn

I would suggest using pnpm throughout the codebase including the vs code extension instead of having to switch to npm or yarn otherwise it could result in problems later on.

rbalicki2:
Okay! That's a good call. I think it's only using yarn because that's what
happened to be done in the Relay vscode extension, and I know for a fact we
didn't use it for any particular reason

PR #117, opened 2024-09-23
recycle old nodes

### LINKED ISSUE #92: Loadable fields (et al) cause unnecessary re-renders
* loadable fields cause unnecessary re-renders, because they are different from render to render.
* I can provide more details, but this is a great first issue!
* The sketch of the fix is to use the reader AST to compare the before and after parameters, and (when we encounter a loadable field) always consider it equal.
* This is fine, because the loadable field doesn't close over anything other than the id, so that should be fine! (Though this can probably be confirmed)

PatrykWalach:
If there's a need this could easily be done in immutable way.

Closes #92

### DIFF
--- a/libs/isograph-react/src/core/areEqualWithDeepComparison.ts
+++ b/libs/isograph-react/src/core/areEqualWithDeepComparison.ts
@@ -1,38 +1,34 @@
 import type { ReaderAst, ReaderLinkedField, ReaderScalarField } from './reader';
-export function areEqualWithDeepComparison(
+export function mergeUsingReaderAst(
   field: ReaderScalarField | ReaderLinkedField,
   oldItem: unknown,
   newItem: unknown,
-): boolean {
+): unknown {
   if (newItem === null) {
-    return oldItem === null;
+    return oldItem === null ? oldItem : newItem;
   }
 
   if (newItem === undefined) {
-    return oldItem === undefined;
+    return oldItem === undefined ? oldItem : newItem;
   }
 
   if (Array.isArray(newItem)) {
     if (!Array.isArray(oldItem)) {
-      return false;
+      return newItem;
     }
 
-    return areEqualArraysWithDeepComparison(field, oldItem, newItem);
+    return mergeArraysUsingReaderAst(field, oldItem, newItem);
   }
 
   switch (field.kind) {
     case 'Scalar':
-      return newItem === oldItem;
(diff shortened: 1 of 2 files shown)


libs/isograph-react/src/core/areEqualWithDeepComparison.ts
```diff
@@ -42,29 +38,34 @@ export function areEqualWithDeepComparison(
   }
 }
 
-export function areEqualArraysWithDeepComparison(
+export function mergeArraysUsingReaderAst(
```
PatrykWalach, line 41: It's probably not worth it to recycle all elements in an array, since it would be O(n^2).

But there's a potential to recycle `Nodes` using `id` in O(n).

rbalicki2: I'm fine with just recycling if the new items are appended (i.e. they're only recycled if their order is the same).

But if you want to implement an O(1) solution, that would be awesome, and will probably help make the situation where users prepend more performant. Let's do it as a separate PR.

I'm not sure there's an obvious circumstance in which someone puts a specific item from an array into a dependency array, though. You would probably do something like `users.find(...)` or `users[0]` or something. I suppose that's not impossible, but doesn't seem like a common pattern.

libs/isograph-react/src/core/cache.ts
```diff
@@ -262,13 +262,12 @@ function callSubscriptions(
...
             );
 
-            if (
-              !areEqualObjectsWithDeepComparison(
-                subscription.readerAst,
-                subscription.encounteredDataAndRecords.item,
-                newEncounteredDataAndRecords.item,
-              )
-            ) {
+            const mergedItem = mergeObjectsUsingReaderAst(
```
rbalicki2, line 265: I think we need to replace `newEncounteredDataAndRecords.item` with `mergedItem`, e.g. on lines 276 and 280. What do you think?

rbalicki2 (changes requested):
Awesome!!! Looks great, just lmk if we need to make that change and we can land this. Amazing work

libs/isograph-react/src/core/cache.ts
rbalicki2: Wait, we don't. We just mutate the new object instead of returning an entirely new object. Okay, awesome, let's land

rbalicki2 (approved):

PR #119, opened 2024-09-24
Validate that no excessive variable are passed

### LINKED ISSUE #97: Validate that no excessive variable are passed
```
export const foo = iso(`
  field User.Foo {}
`)((props) => {});

export const bar = iso(`
  field User.Bar {
    Foo(invalidArg: true)
  }
`)(fn);
```

should fail to compile, because `invalidArg` is not an argument to `User.Foo`.

The task is to add function that is called after https://github.com/isographlabs/isograph/blob/75a6eaf1ab694fd51d571394c24c132050191e59/crates/isograph_schema/src/validate_schema.rs#L971 is called in `validate_schema`, with signature

```
fn validate_no_extraneous_arguments(
    argument_definitions: &[WithLocation<ValidatedVariableDefinition>],
    arguments: &[WithLocation<SelectionFieldArgument>],
  // other args?
) -> ValidateSelectionsResult<()>
```

which results `Err` if there are arguments that are not allowed by the definitions.

This task does not include type checking! That's a bigger task and I want to make sure we're doing it correctly.

PatrykWalach:
Closes #97


rbalicki2 (commented):
Let's just special-case the `id` field for now.

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -968,13 +977,63 @@ pub fn categorize_field_loadability<'a>(
...
+    location: Location,
+) -> ValidateSelectionsResult<()> {
+    let extra_arguments: Vec<_> = arguments
+        .iter()
+        .filter_map(|arg| {
+            // TODO remove this
+            // With @exposeField on Query, id field is needed because the generated
+            // query is like node(id: $id) { ... everything else }, but that
+            // id field is added in somewhere else
+            if arg.item.name.item.lookup() == "id" {
```
rbalicki2, line 992: nit: let's define a variable up top `let foo = "id".intern().into()` and compare `arg.item.name.item == foo` so that we do work on every iteration. We could also move it into a `lazy_static!` block, which would allow us to reuse the same value across calls to `validate_no_extraneous_arguments`

rbalicki2: e.g. https://github.com/isographlabs/isograph/blob/d8aff7180684e69152eb93578b59673be4e54159/crates/graphql_artifact_generation/src/generate_artifacts.rs#L44

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -968,13 +980,68 @@ pub fn categorize_field_loadability<'a>(
...
+) -> ValidateSelectionsResult<()> {
+    let extra_arguments: Vec<_> = arguments
+        .iter()
+        .filter_map(|arg| {
+            // TODO remove this
+            // With @exposeField on Query, id field is needed because the generated
+            // query is like node(id: $id) { ... everything else }, but that
+            // id field is added in somewhere else
+
+            if arg.item.name.item.lookup() == ID.lookup() {
```
rbalicki2, line 1000: no need to call `.lookup` here, you can compare the items directly. (If you can't, then change the type of the `FieldArgumentName` to whatever the left hand side is)

PR #120, opened 2024-09-24
#116: add variables to resolver data

### LINKED ISSUE #116: client field first parameter: `({ data, variables })` instead of `data`
* Right now, we don't give you the ability to read the fragment variables. What a bummer! That seems useful.
* Another advantage of this is that passing an object here allows us to, in the future, pass additional props as part of that first parameter. `refetch` might be useful, for example.
* This is a great first issue, it exposes you to both the compiler and the runtime

## Gist

- Modify `EagerReaderArtifact` and `ComponentReaderArtifact`'s types (but not `RefetchReaderArtifact`)
- Wherever we call `.resolver(data, ...)`, we should instead pass `{ data, variables }`. (Except when we call `field.refetchReaderArtifact.resolver(...)` in `read.ts`)
- Modify https://github.com/isographlabs/isograph/blob/d57417f09b33df676192f7176f4f465c99567c41/crates/graphql_artifact_generation/src/iso_overload_file.rs#L82-L95 to reflect the new parameter
- Rebuild the demos (`pnpm run build-demos`)
- Spot check to ensure stuff works

ch1ffa:
Closes #116

This PR exposes `variables` field to `EagerReaderArtifact` and `ComponentReaderArtifact` resolvers. 

It has been temporarily added as `Variables` type. In the next PR it should be replaced by generated type.

### DIFF
--- a/libs/isograph-react/src/core/cache.ts
+++ b/libs/isograph-react/src/core/cache.ts
@@ -276,7 +276,6 @@ function callSubscriptions(
                   new: newEncounteredDataAndRecords.item,
                 });
               }
-              // TODO deep compare values
               subscription.callback(newEncounteredDataAndRecords);
             } else {
               // @ts-expect-error


libs/isograph-react/src/core/reader.ts
```diff
@@ -17,32 +17,45 @@ export type TopLevelReaderArtifact<
...
   readonly componentName: ComponentOrFieldName;
   readonly readerAst: ReaderAst<TReadFromStore>;
   readonly resolver: (
-    data: TReadFromStore,
+    data: ResolverData<TReadFromStore, TVariables>,
     runtimeProps: TComponentProps,
   ) => React.ReactNode;
 };
 
+export type ResolverData<
```
ch1ffa, line 51: not sure about this type name

rbalicki2: `ResolverFirstParameter`?

crates/graphql_artifact_generation/src/eager_reader_artifact.rs
```diff
@@ -124,6 +124,8 @@ pub(crate) fn generate_eager_reader_param_type_artifact(
         &mut loadable_field_encountered,
     );
 
+    // TODO: import generated variable types
```
rbalicki2, line 127: Nit: Change `0` to `1` above so that the indentation is correct

crates/graphql_artifact_generation/src/eager_reader_artifact.rs
```diff
@@ -136,7 +138,11 @@ pub(crate) fn generate_eager_reader_param_type_artifact(
     let param_type_content = format!(
         "{param_type_import_statement}\n\
         {loadable_field_import}\
-        export type {reader_param_type} = {client_field_parameter_type};\n",
+        {variables_import_statement}\n\
+        export type {reader_param_type} = {{\n\
```
rbalicki2, line 142: Let's fix the indentation here by adding `{indent}` where `let indent = "  "`

libs/isograph-react/src/core/reader.ts
```diff
@@ -17,32 +17,45 @@ export type TopLevelReaderArtifact<
...
+  TReadFromStore extends Object,
+  TVariables = Variables,
+> = {
+  readonly data: TReadFromStore;
+  readonly variables: TVariables | null;
```
rbalicki2, line 56: I just pushed a commit that makes `variables` non-null in `FragmentReference`, so no need to make it null here

rbalicki2 (changes requested):
Just a few minor changes and then this looks g2g!! I'm super excited about this. I wasn't expecting that `Variables` would be part of the `param_type`, but that's certainly the correct move. Awesome!

rbalicki2:
Sorry for the last minute change. Let's not name it `variables`. Variables are global in GraphQL, which will confuse users.

Let's instead call it either `arguments` or `parameters`. `arguments` I think would be suboptimal, because one cannot name a variable arguments because JavaScript is really silly.

So, by process of elimination, let's name this field `parameters`. Thoughts?

ch1ffa:
Sorry, I'll make a new one, accidentally created PR from the main branch and it's a pain to rebase onto new commits

PR #128, opened 2024-09-25
#116: add parameters field to resolver

### LINKED ISSUE #116: client field first parameter: `({ data, variables })` instead of `data`

ch1ffa:
Closes #116 

This PR exposes `parameters` field to `EagerReaderArtifact` and `ComponentReaderArtifact` resolvers.

It has been temporarily added as `Variables` type. In the next PR it should be replaced by a generated type.

### DIFF
--- a/libs/isograph-react/src/core/reader.ts
+++ b/libs/isograph-react/src/core/reader.ts
@@ -1,5 +1,5 @@
 import { Factory } from '@isograph/disposable-types';
-import { FragmentReference } from './FragmentReference';
+import { FragmentReference, Variables } from './FragmentReference';
 import {
   ComponentOrFieldName,
   DataId,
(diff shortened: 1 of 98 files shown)


libs/isograph-react/src/core/reader.ts
```diff
@@ -17,32 +17,45 @@ export type TopLevelReaderArtifact<
...
     runtimeProps: TComponentProps,
   ) => React.ReactNode;
 };
 
+export type ResolverFirstParameter<
+  TReadFromStore extends Object,
+  TVariables = Variables,
+> = {
+  readonly data: TReadFromStore;
+  readonly parameters: TVariables;
```
ch1ffa, line 56: renamed field name to `parameters` only, didn't touch type names

rbalicki2 (approved):
Love it! LGTM! Shipittt (after CI passes)

PR #132, opened 2024-09-26
fix `useSkipLimitPagination` to refetch on status change

PatrykWalach:
```tsx
const firstParameter = {
	data,
	parameters: fragmentReference.variables,
}

return readerWithRefetchQueries.readerArtifact.resolver(
	firstParameter,
	undefined
) as ReadonlyArray<any>
 ```
The recent change to add parameters was the issue causing errors.

### CODE REVIEW

File: libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts [hunk 1]
```diff
@@ -105,8 +107,13 @@ export function useSkipLimitPagination<
         fragmentReference.readerWithRefetchQueries,
       );
 
+      const firstParameter = {
```
rbalicki2, line 110: 😭 can't believe we missed this

rbalicki2 (approved):
Looks great to me!!! Thanks for catching and fixing this.

PR #133, opened 2024-09-26
Validate no unused variables

PatrykWalach:
Started this.
Need to pass a better location.
Add argument recursion, (`[1, $foo, 2, $bar]`).

Closes #133

### DIFF
--- a/crates/isograph_lang_types/src/client_field_declaration.rs
+++ b/crates/isograph_lang_types/src/client_field_declaration.rs
@@ -336,6 +336,18 @@ impl NonConstantValue {
     pub fn reachable_variables(&self) -> Vec<VariableName> {
         match self {
             NonConstantValue::Variable(name) => vec![*name],
+            NonConstantValue::Object(object) => {
+                return object
+                    .iter()
+                    .flat_map(|pair| pair.value.item.reachable_variables())
+                    .collect();
+            }
+            NonConstantValue::List(list) => {
+                return list
+                    .iter()
+                    .flat_map(|element| element.item.reachable_variables())
+                    .collect();
+            }
             _ => vec![],
         }
     }
--- a/crates/isograph_schema/src/validate_schema.rs
+++ b/crates/isograph_schema/src/validate_schema.rs
@@ -445,12 +445,12 @@ fn validate_all_variables_are_used(
     used_variables: BTreeSet<VariableName>,
 ) -> ValidateSelectionsResult<()> {
     let unused_variables: Vec<_> = variable_definitions
-        .iter()
+        .into_iter()
         .filter_map(|variable| {
             let is_used = used_variables.contains(&variable.item.name.item);
 
             if !is_used {
-                return Some(variable.clone());
+                return Some(variable);
             }
             None
         })
--- a/crates/isograph_schema/src/validate_schema.rs
+++ b/crates/isograph_schema/src/validate_schema.rs
@@ -721,11 +721,11 @@ fn validate_client_field_definition_selections_exist_and_types_match(
                 &mut used_variables,
             )
         })
-        .collect::<Result<_, _>>();
+        .collect::<Result<_, _>>()?;
 
     validate_all_variables_are_used(variable_definitions, used_variables)?;
 
-    return validated_selection_set;
+    return Ok(validated_selection_set);
 }
 
 fn validate_client_field_definition_selection_exists_and_type_matches(
--- a/crates/isograph_schema/src/validate_schema.rs
+++ b/crates/isograph_schema/src/validate_schema.rs
@@ -1,4 +1,4 @@
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 use common_lang_types::{
     FieldArgumentName, IsographObjectTypeName, Location, SelectableFieldName, UnvalidatedTypeName,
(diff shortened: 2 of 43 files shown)


crates/isograph_schema/src/validate_schema.rs
```diff
@@ -667,11 +702,14 @@ fn validate_client_field_definition_selections_exist_and_types_match(
     parent_object: &SchemaObject,
     server_fields: &[ValidatedSchemaServerField],
     client_field_args: &ClientFieldArgsMap,
+    variable_definitions: Vec<WithSpan<UnvalidatedVariableDefinition>>,
 ) -> ValidateSelectionsResult<Vec<WithSpan<ValidatedSelection>>> {
     // Currently, we only check that each field exists and has an appropriate type, not that
     // there are no selection conflicts due to aliases or parameters.
 
-    selection_set
+    let mut used_variables: Vec<VariableName> = vec![];
```
rbalicki2, line 710: Let's keep a `BTreeSet` instead of a `Vec` here.

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -440,6 +440,33 @@ fn validate_and_transform_client_fields(
     }))
 }
 
+fn validate_all_variables_are_used(
+    variable_definitions: Vec<WithSpan<UnvalidatedVariableDefinition>>,
+    used_variables: Vec<VariableName>,
```
rbalicki2, line 445: love that we're consuming it here, this forces us to collect all used variables before we validate

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -711,11 +758,16 @@ fn validate_client_field_definition_selection_exists_and_type_matches(
...
                         client_field_args,
+                        &mut used_variables2,
                     )
                 },
             )
         })
-    })
+    });
+
+    used_variables.append(&mut used_variables2);
```
rbalicki2, line 768: This annoys me, but I've run into this issue before as well. This is what I would expect us to do.

rbalicki2 (changes requested):
Besides the `BTreeSet` changes, let's also extract some of this into methods where appropriate, e.g. `impl NonConstantValue { fn variables(&self) -> Vec<VariableName> }` and the like.

That will make the recursion easy too

rbalicki2:
https://github.com/isographlabs/isograph/blob/main/crates/isograph_lang_types/src/client_field_declaration.rs#L336-L337

crates/isograph_lang_types/src/client_field_declaration.rs
```diff
@@ -336,6 +336,18 @@ impl NonConstantValue {
     pub fn reachable_variables(&self) -> Vec<VariableName> {
         match self {
             NonConstantValue::Variable(name) => vec![*name],
+            NonConstantValue::Object(object) => {
+                return object
+                    .iter()
+                    .flat_map(|pair| pair.value.item.reachable_variables())
+                    .collect();
+            }
+            NonConstantValue::List(list) => {
```
rbalicki2, line 345: nice, this might fix some bugs haha

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -440,6 +440,31 @@ fn validate_and_transform_client_fields(
...
+    variable_definitions: Vec<WithSpan<UnvalidatedVariableDefinition>>,
+    used_variables: BTreeSet<VariableName>,
+) -> ValidateSelectionsResult<()> {
+    let unused_variables: Vec<_> = variable_definitions
+        .iter()
+        .filter_map(|variable| {
+            let is_used = used_variables.contains(&variable.item.name.item);
+
+            if !is_used {
+                return Some(variable.clone());
```
rbalicki2, line 453: use `into_iter()` so we can avoid a clone

demos/github-demo/src/isograph-components/PullRequestRoute.tsx
```diff
@@ -10,9 +10,9 @@ import {
 import { Container } from '@mui/material';
 
 export const PullRequest = iso(`
-  field Query.PullRequest($repositoryOwner: String!, $repositoryName: String!, $pullRequestNumber: Int!, $last: Int!) @component {
+  field Query.PullRequest($repositoryOwner: String!, $repositoryName: String!, $pullRequestNumber: Int!) @component {
     Header
-    PullRequestDetail
+    PullRequestDetail(repositoryOwner: $repositoryOwner, repositoryName: $repositoryName, pullRequestNumber: $pullRequestNumber)
```
rbalicki2, line 15: 😅  I thought we already validated for undefined variables! whoops #134

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -680,9 +716,14 @@ fn validate_client_field_definition_selections_exist_and_types_match(
                 schema_data,
                 server_fields,
                 client_field_args,
+                &mut used_variables,
             )
         })
-        .collect::<Result<_, _>>()
+        .collect::<Result<_, _>>()?;
+
+    validate_all_variables_are_used(variable_definitions, used_variables)?;
```
rbalicki2, line 724: #135 added a task to return multiple errors from this function.
It doesn't really matter for watch mode, but it is nice to see all the errors in VSCode when we work on the language server

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -440,6 +440,31 @@ fn validate_and_transform_client_fields(
...
+                return Some(variable.clone());
+            }
+            None
+        })
+        .collect();
+
+    if !unused_variables.is_empty() {
+        return Err(WithLocation::new(
+            ValidateSelectionsError::UnusedVariables { unused_variables },
+            Location::generated(),
```
rbalicki2, line 462: I'm surprised that `ClientField`'s do not have location information for the name.

If you want to do that, you can change the `ClientField`'s `name` field to be `WithLocation<SelectableFieldName>`.

Happy to also do `Location::generated` here in the mean time. (I think I want to rethink a few things about how we store locations.) If we do that, we might want to add the client field's parent type and name and render it as `Type.name` in the error message

demos/github-demo/src/isograph-components/PullRequestRoute.tsx
PatrykWalach: For now I'll leave it to someone else as a first good issue

crates/isograph_schema/src/validate_schema.rs
PatrykWalach: This is something I would like to do later on, maybe after I ran out of issues, we talked about in the last video call

crates/isograph_schema/src/validate_schema.rs
PatrykWalach: I think it's good enough for now, this is easy to tweak later, and probably will with the work on vs code extension

PatrykWalach:
Should be good now

rbalicki2 (approved):
LGTM shipittt

PR #140, opened 2024-09-28
fix extra final "argument" being ignored instead of erroring

PatrykWalach:
Added support for empty lists as well


rbalicki2 (approved):
Amazing, great work, shipittt

PR #141, opened 2024-09-28
Using workspaces for dependency management

edmondop:
This pr simplifies dependency management and ensure we are coherent on the version we use across the different crates


rbalicki2 (approved):
LGTM SHIPIT

PR #142, opened 2024-09-30
chore: Run most Rust tests in CI

lanesawyer:
# chore: Run most Rust tests in CI

## What
- Adds another job to the CI Actions that runs `cargo test` for the entire repository

## How
- Copied the `rustfmt` job and tweaked it
- Added it to the dependent jobs for `all-checks-passed`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,6 +203,20 @@ jobs:
         run: cargo clippy
       - name: 'Check working directory status'
         run: './scripts/check-git-status.sh'
+  
+  cargo_test:
+    name: Run cargo test (excluding relay tests)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          components: rustfmt
+      - name: Run cargo test
+        run: cargo test
 
   all-checks-passed:
     name: All checks passed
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -204,6 +204,20 @@ jobs:
       - name: 'Check working directory status'
         run: './scripts/check-git-status.sh'
 
+  cargo_test:
+    name: Run cargo test (excluding relay tests)
```
lanesawyer, line 208: How's the name? Happy to tweak it!

We could also run it as part of the `rustfmt` just to consolidate the tooltchain download stuff. Would shave some seconds off the run time.

rbalicki2: Name SGTM, also keeping it a separate step sounds good to me. I think it's fine, because these steps are run in parallel and aren't likely to slow down the `all-checks-passed` step, so this is great.

.github/workflows/ci.yml
```diff
@@ -204,6 +204,20 @@ jobs:
       - name: 'Check working directory status'
         run: './scripts/check-git-status.sh'
 
+  cargo_test:
```
rbalicki2, line 207: nit: can we use `cargo-test` for consistency? Or rename everything with underscores? (Happy with either)

rbalicki2 (changes requested):
minor nit re: kebab vs snake case

.github/workflows/ci.yml
lanesawyer: Good catch, my brain was in Rust mode!

rbalicki2:
shipitttt

Issue #65, opened 2024-04-16
Babel transform (and SWC transform eventually) should also replace iso field literals


rbalicki2:
* Right now, iso field literals contain a very long string. This is useless and ignored at runtime.
* We should modify the babel transform to replace `` iso(`field blah blah`) `` with `(x) => x` (or ... if possible... with just the value that is passed to `iso(...)`, i.e. the actual resolver.
* cc @edmondop the SWC plugin should do this too
* The goal here is to not include the field string in the code that is sent to the browser.

PR #146, opened 2024-09-30
add jsdoc to babel plugin

PatrykWalach:
I would like to do #65 next, and having types would help a lot

I've used jsdoc to not add a build step for now, it can be added later, I hope that's ok

### DIFF
--- a/libs/isograph-babel-plugin/compileTag.js
+++ b/libs/isograph-babel-plugin/compileTag.js
@@ -3,6 +3,12 @@
 const pathModule = require('path');
 const os = require('os');
 
+/**
+ * @typedef {import("@babel/core")} babel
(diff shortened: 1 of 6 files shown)


libs/isograph-babel-plugin/compileTag.js
```diff
@@ -85,10 +109,10 @@ function compileImportStatement(t, path, type, field, artifactType, config) {
 
   path.replaceWith(
     t.memberExpression(
-      t.CallExpression(t.Identifier('require'), [
-        t.StringLiteral(fileToArtifact),
+      t.callExpression(t.identifier('require'), [
```
rbalicki2, line 112: do these both exist? interesting

Issue #134, opened 2024-09-27
Validate no undefined variables


rbalicki2:
If a client field uses a variable that hasn't been defined, emit an error

rbalicki2:
Done

PR #147, opened 2024-10-02
Generate parameter types

ch1ffa:
This PR introduces type generation for `parameters` field that makes autocomplete work for resolver parameters, mutation and `useLazyReference` arguments.

There are a couple of open questions though:

1. We need to filter out `__typename` variables from mutation inputs and I'm not sure where is the good place for it
2. Should we generate parameters for a client field type when it's used as a parameter? It's just 2 lines of code, but what's a use case for it?
3. In the github-demo, inside `HomeRoot` there is a parameter that is used as a `lazyLoadReference` parameter, but is absent in the `Query.HomePage` field. It resides deeper in the child component `RepositoryList`. Should we expose all children parameters in the parent component then?
4. `encountered_fields` in `SelectableServerFieldId::Object` are not sorted and field order is not stable for now. Can we replace `HashMap` with `BTreeMap` for `encountered_fields`?

### DIFF
--- a/crates/graphql_artifact_generation/src/format_parameter_type.rs
+++ b/crates/graphql_artifact_generation/src/format_parameter_type.rs
@@ -1,0 +1,172 @@
+use common_lang_types::SelectableFieldName;
+use graphql_lang_types::{GraphQLNonNullTypeAnnotation, GraphQLTypeAnnotation};
+
+use isograph_lang_types::{ClientFieldId, SelectableServerFieldId, ServerFieldId};
+use isograph_schema::{FieldDefinitionLocation, TypeAnnotation, UnionVariant, ValidatedSchema};
+
+pub(crate) fn format_parameter_type(
+    schema: &ValidatedSchema,
+    type_: GraphQLTypeAnnotation<SelectableServerFieldId>,
+    indentation_level: u8,
+) -> String {
+    match type_ {
+        GraphQLTypeAnnotation::Named(named_inner_type) => {
+            format!(
+                "{} | null | void",
+                format_server_field_type(schema, named_inner_type.item, indentation_level)
+            )
+        }
+        GraphQLTypeAnnotation::List(list) => {
+            format!(
+                "ReadonlyArray<{}> | null",
+                format_server_field_type(schema, *list.inner(), indentation_level)
+            )
+        }
+        GraphQLTypeAnnotation::NonNull(non_null) => match *non_null {
+            GraphQLNonNullTypeAnnotation::Named(named_inner_type) => {
+                format_server_field_type(schema, named_inner_type.item, indentation_level)
+            }
+            GraphQLNonNullTypeAnnotation::List(list) => {
+                format!(
+                    "ReadonlyArray<{}>",
+                    format_server_field_type(schema, *list.inner(), indentation_level)
+                )
+            }
+        },
+    }
+}
+
+fn format_server_field_type(
+    schema: &ValidatedSchema,
+    field: SelectableServerFieldId,
+    indentation_level: u8,
+) -> String {
+    match field {
+        SelectableServerFieldId::Object(object_id) => {
+            let mut s = "{\n".to_string();
+            for (name, field_definition) in schema
+                .server_field_data
+                .object(object_id)
+                .encountered_fields
+                .iter()
+                .filter(|x| matches!(
+                    x.1,
+                    FieldDefinitionLocation::Server(server_field_id) if schema.server_field(*server_field_id).is_discriminator == false),
+                )
+            {
+                let field_type =
+                    format_field_definition(schema, name, field_definition, indentation_level + 1);
+                s.push_str(&field_type)
+            }
+            s.push_str(&format!("{}}}", "  ".repeat(indentation_level as usize)));
+            s
+        }
+        SelectableServerFieldId::Scalar(scalar_id) => schema
+            .server_field_data
+            .scalar(scalar_id)
+            .javascript_name
+            .to_string(),
+    }
+}
+
+fn format_field_definition(
+    schema: &ValidatedSchema,
+    name: &SelectableFieldName,
+    type_: &FieldDefinitionLocation<ServerFieldId, ClientFieldId>,
+    indentation_level: u8,
+) -> String {
+    match type_ {
+        FieldDefinitionLocation::Server(server_field_id) => {
+            let type_annotation = &schema.server_field(*server_field_id).associated_data;
+            format!(
+                "{}readonly {}: {},\n",
+                "  ".repeat(indentation_level as usize),
+                name,
+                format_type_annotation(schema, type_annotation, indentation_level + 1),
+            )
+        }
+        FieldDefinitionLocation::Client(_) => {
+            panic!(
+                "Unexpected object. Client fields are not supported as parameters, yet. \
+                This is indicative of an unimplemented feature in Isograph."
+            )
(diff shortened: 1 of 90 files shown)


rbalicki2:
Re: GitHub Demo

* Ooh interesting, that's actually indicative of a bug. The reason this currently validates is that the `first` parameter is optional, so passing `null` is fine
* We should review whether we want optional parameters, though. TBH it seems like a footgun. It seems more explicit to have a default value of `null`.
* And we currently don't validate unused parameters, but we should: #134 
* So we have some weird behavior! We send the `first` parameter to the backend, and it is used by the `watchers` and `pullRequests` field, but we alias that field as if it has a null parameter:
Screenshot 2024-10-02 at 11 51 46 AM
* So uh, bugs all around, but coincidentally, not broken in practice in this specific case.

----

The fix:

* validate no unused fields
* that will force us to thread the `first` field down to the appropriate subfield, or drop it
* that will cause us to not have a mismatch between what the reader thinks we're doing (i.e. it thinks we're passing null), and what the server sees (us passing first: 15)

Separately, that `first` field is doing exactly nothing, so for now, I'll remove that field in master

rbalicki2:
RE: encountered fields BTreeMap, yes, 100%, we should (probably) almost always be using `BTreeMap/Set` everywhere instead of `HashMap/Set`. And we should always use the BTree version if we're writing the output to a file or to an error message, or otherwise relying on order

So go ahead and make that change, please!

ch1ffa:
> RE: encountered fields BTreeMap, yes, 100%, we should (probably) almost always be using `BTreeMap/Set` everywhere instead of `HashMap/Set`. And we should always use the BTree version if we're writing the output to a file or to an error message, or otherwise relying on order
> 
> 
> 
> So go ahead and make that change, please! (Bonus points if it is a separate PR that lands independently, but no big deal either way)

Sure, I'll make a separate one. But it should be merge first or checks is going to be failed

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -614,6 +614,114 @@ fn format_type_for_js_inner(
     }
 }
 
+pub(crate) fn generate_parameters<'a>(
+    schema: &ValidatedSchema,
+    argument_definitions: impl Iterator<Item = &'a VariableDefinition<SelectableServerFieldId>>,
+    indentation_level: u8,
```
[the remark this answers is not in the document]
ch1ffa: If we want to allow client fields as params we can call `generate_variables` from the `FieldDefinitionLocation::Client(_)` branch again and preserve indentation.

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -614,6 +614,114 @@ fn format_type_for_js_inner(
...
+    field: SelectableServerFieldId,
+    indentation_level: u8,
+) -> String {
+    match field {
+        SelectableServerFieldId::Object(object_id) => {
+            let mut parameters_type = "{\n".to_string();
+            for (name, field_definition) in schema
+                .server_field_data
+                .object(object_id)
+                .encountered_fields
```
rbalicki2, line 680: I think a good solution would be to make `SchemaServerField` an enum with variants `Discriminator`, `StrongId`, and `EverythingElse`. But, in the short term, I think we should make a separate PR that:

* adds a field to `SchemaServerField`
```
// TODO remove this. This is indicative of poor modeling.
is_discriminator: bool
```
* set that to true only in [`process_type_definition:691`](https://github.com/isographlabs/isograph/blob/0254e4b5afdec037715e443012b9e32b408645fc/crates/isograph_schema/src/process_type_definition.rs#L691-L698)
* add a method to to `SchemaObject` that is called `non_discriminator_fields`. Also, add a `// TODO remove this, this is indicative of poor modeling` comment 😂 
* call that here

rbalicki2:
Re 2: I'm not sure what this would mean, quite yet! So let's continue to panic if we encounter that. This should not be possible to do, I think, so panicking is the appropriate behavior

crates/graphql_artifact_generation/src/generate_artifacts.rs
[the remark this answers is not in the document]
rbalicki2: ah, I see, yeah... I don't think it makes sense (at least yet) to have client fields are parameters. And if necessary, we can add the indentation back when we figure out what that means

crates/graphql_artifact_generation/src/generate_artifacts.rs
rbalicki2: #150 for an actual solution, which needs some thinking

crates/graphql_artifact_generation/src/format_parameter_type.rs
```diff
@@ -0,0 +1,172 @@
...
+            let type_annotation = &schema.server_field(*server_field_id).associated_data;
+            format!(
+                "{}readonly {}: {},\n",
+                "  ".repeat(indentation_level as usize),
+                name,
+                format_type_annotation(schema, type_annotation, indentation_level + 1),
+            )
+        }
+        FieldDefinitionLocation::Client(_) => {
+            panic!(
```
rbalicki2, line 89: Ooooh interesting, I was thinking about this, and this might break 😂  There's nothing stopping you from defining a client field on an InputObject. LOL. Anyway, this is fine as it is, as people should not be doing that?

The reason you're allowed to is that Isograph doesn't have a distinction between input and output objects (IMO, that's silly and won't make sense in non-GraphQL worlds). So anyway, this panic is good

libs/isograph-react/src/core/FragmentReference.ts
```diff
@@ -7,6 +7,18 @@ export type VariableValue = string | number | boolean | null | object;
 
 export type Variables = { readonly [index: string]: VariableValue };
 
+export type ExtractData<T> = T extends {
+  data: infer D extends object;
```
rbalicki2, line 11: ooooh I didn't know you could do `infer D extends object,` that's neat

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -621,6 +622,25 @@ fn format_type_for_js_inner(
...
+    argument_definitions: impl Iterator<Item = &'a VariableDefinition<SelectableServerFieldId>>,
+) -> String {
+    let mut s = "{\n".to_string();
+    let indent = "  ";
+    for arg in argument_definitions {
+        let is_optional = !matches!(arg.type_, GraphQLTypeAnnotation::NonNull(_));
+        s.push_str(&format!(
+            "{indent}readonly {}{}: {},\n",
+            arg.name.item,
+            if is_optional { "?" } else { "" },
```
rbalicki2, line 636: I wonder if we should make optional fields also `?`, not just optional arguments for consistency. Will DM you, this looks amazing and can land

rbalicki2 (approved):
Awesome!!! Love it. Landing. Thank you for this absolutely amazing work

crates/graphql_artifact_generation/src/generate_artifacts.rs
rbalicki2: #160

crates/graphql_artifact_generation/src/format_parameter_type.rs
rbalicki2: #161

PR #148, opened 2024-10-02
Babel replace iso field literals

### LINKED ISSUE #65: Babel transform (and SWC transform eventually) should also replace iso field literals

PatrykWalach:
Do we want to throw, instead of `x => x`, is lack of parenthesis an issue?

Closes #65

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,6 +204,20 @@ jobs:
       - name: 'Check working directory status'
         run: './scripts/check-git-status.sh'
 
+  cargo-test:
+    name: Run cargo test (excluding relay tests)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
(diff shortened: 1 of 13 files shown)


rbalicki2:
Oh... it looks like it was caching PetSummaryCard.

We should find a good way to completely clear the babel cache.

rbalicki2:
In what circumstances would we return `x => x`? That seems like it will never occur, since the compiler will reject iso literals without a function attached to them.

(Though, it is conceivable to eventually allow that, as a way of defining a "chunk of untransformed data".)

rbalicki2:
Before:
Screenshot 2024-10-02 at 1 10 06 PM

After:
Screenshot 2024-10-02 at 1 10 53 PM

download (1)

Issue #150, opened 2024-10-02
SchemaServerField/SchemaObject modeling


rbalicki2:
My overall thoughts:
* `SchemaObject` should have three fields:
     * `discriminator: Option<DiscriminatorField>`
     * `strong_id: Option<StrongIdField>`
     * `all_other_fields: BTreeSet<SchemaServerField>`
* All three of these types should be convertible to some common type
* `SchemaObject` should have methods like `.fields()`, `.non_discriminator_fields()` etc

Somewhere, we have something similar: when you sort fields (for printing), `__typename` and `id` come first. I can't find it right now, though

rbalicki2:
Closed as not subsumed by other issues, i.e. I am not sure this will make that much sense in the incremental world.

PR #156, opened 2024-10-05
add is_discriminator

ch1ffa:
This PR adds `is_discriminator` field as a temporary solution for #147 to filter `__typename` field. Actual solution is #150.


rbalicki2:
LGTM!

PR #158, opened 2024-10-05
Deep compare arguments

### LINKED ISSUE #126: Deep compare arguments
* After #120, we only re-render components if their read out fragment data changes. But, we should also re-render components if their variables changes, since we're now exposing that to the user as well. So, we should deep compare variables.
* We can probably use `readerAst.arguments` to compare parameters

PatrykWalach:
`readerAst` is an array of ast nodes, there is no `readerAst.arguments`, it should be ok to compare these without ast, right?

Closes #126

### DIFF
--- a/libs/isograph-react/src/core/areEqualWithDeepComparison.ts
+++ b/libs/isograph-react/src/core/areEqualWithDeepComparison.ts
@@ -110,3 +110,82 @@ export function mergeObjectsUsingReaderAst(
 
   return canRecycle ? oldItemObject : newItemObject;
 }
+
+export function areEqualWithDeepComparison(
+  oldItem: unknown,
+  newItem: unknown,
+): boolean {
+  if (newItem === null) {
+    return oldItem === null;
+  }
+
+  if (newItem === undefined) {
+    return oldItem === undefined;
+  }
+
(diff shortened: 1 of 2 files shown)


PatrykWalach:
No need to take a look at this now, I'll have to take a look what merge from `main` has changed

PatrykWalach:
https://github.com/isographlabs/isograph/blob/3f50636d6d1a968ef0a88bf3f52f2d3ea81b769a/libs/isograph-react/src/core/componentCache.ts#L14-L16
Right now we just create a new component whenever, `fragmentReference.variables` change, so we don't have to do anything to rerender.
Don't we lose all the state doing this? We should probably change cache to have two layers:
- `unique id` for every `useLazyReference`; not every `fragmentReference`, because it changes whenever variables change
- component name

this way we can subscribe to change of `variables` and `data` with this `unique id`.
Is that what we want to do here?
If so, how do we generated a stable id, do we just use `useId`?

rbalicki2:
> Right now we just create a new component whenever, fragmentReference.variables change, so we don't have to do anything to rerender.
> Don't we lose all the state doing this? 

This is a really good point. I'll think about this. Thanks for pointing that out.

PR #159, opened 2024-10-06
#101 one way relay style pagination

### LINKED ISSUE #101: Feature: one way relay-style pagination
We need a hook that supports Relay-style pagination. This is will be a really fun task. It will probably not be the final version of the Relay-style pagination. That depends on 

The return type should be

```
type UseCursorPaginationReturn<TItem> = {
  readonly kind: 'Pending',
  readonly loadedSoFar: ReadOnlyArray<TItem>,
  readonly pendingFragment: FragmentReference<any, ReadonlyArray<TItem>>;
} | {
  readonly kind: 'Complete',
  readonly fetchMore: (count: number) => void;
  readonly results: ReadonlyArray<TItem>;
  readonly hasMore: boolean;
}
```

More details TBA

PatrykWalach:
Started on this, add to the github demo, somehow have to map the `FragmentReference`
```tsx
  //@ts-expect-error map Connection<TItem> to ReadonlyArray<TItem>
  pendingFragment: mostRecentFragmentReference,
```

Closes #101


rbalicki2:
Awesome!! I will be somewhat busy for the next few days, but I'll take a look as soon as I can.

demos/github-demo/src/isograph-components/__isograph/User/RepositoryConnection/entrypoint.ts
```diff
@@ -0,0 +1,205 @@
+import type {IsographEntrypoint, NormalizationAst, RefetchQueryNormalizationArtifactWrapper} from '@isograph/react';
+import {User__RepositoryConnection__param} from './param_type';
+import {User__RepositoryConnection__output_type} from './output_type';
+import readerResolver from './resolver_reader';
+const nestedRefetchQueries: RefetchQueryNormalizationArtifactWrapper[] = [];
+
+const queryText = 'query RepositoryConnection ($first: Int, $after: String, $id: ID!) {\
```
PatrykWalach, line 7: There's a bug that generates `const queryText = 'mutation RepositoryConnection ...`, but only in this place.
I've manually replaced `mutation` with `query` for testing

rbalicki2: [that's what I'm referring to here](https://github.com/isographlabs/isograph/blob/3f50636d6d1a968ef0a88bf3f52f2d3ea81b769a/docs-website/docs/backlog.md?plain=1#L47)

I think this is due to `generate_entrypoint_artifacts_with_client_field_traversal_result` getting the first fetchable type: `schema.fetchable_types.iter().next()`, which can be a mutation.

Try adding

```
    pub fn query_root_operation_name(&self) -> RootOperationName {
        *self
            .fetchable_types
            .iter()
            .find(|(_, root_operation_name)| root_operation_name.0 == "query")
            .expect("Expected query to be found")
            .1
    }
```

and see if that fixes your issue.

Anyway, this is a weird issue and we should think about what a good solution looks like, becuase ☝️ is not a good solution 😂

libs/isograph-react/src/loadable-hooks/usePagination.ts
```diff
@@ -0,0 +1,315 @@
+import { ItemCleanupPair } from '@isograph/disposable-types';
```
rbalicki2, line 1: hm... I'm thinking this might be more accurate to call this useConnectionSpecPagination? Is that too verbose?

rbalicki2 (commented):
Holy cow, I'm so excited that it works!!! Anyway, just wanted to express how excited I am that **this freaking works**. This is huge. This is going to unlock Isograph use for so many schemas (most immediately Bolt Foundry's)!!!

Anyway, it might take me a few days to finish reviewing, but I'll get to it!

demos/github-demo/src/isograph-components/__isograph/User/RepositoryConnection/entrypoint.ts
PatrykWalach: I've fixed this

libs/isograph-react/src/loadable-hooks/usePagination.ts
```diff
@@ -0,0 +1,315 @@
...
+      });
+
+      const results = readCompletedFragmentReferences(
+        completedFragmentReferences,
+      );
+      return {
+        results: results.edges,
+        kind: 'Pending',
+        //@ts-expect-error map Connection<TItem> to ReadonlyArray<TItem>
+        pendingFragment: mostRecentFragmentReference,
```
PatrykWalach, line 296: How do you feel about this, is it possible to map the fragment to return just the edges, or should it return whole `Connection`?

libs/isograph-react/src/loadable-hooks/usePagination.ts
```diff
@@ -0,0 +1,315 @@
...
+> = ReadonlyArray<LoadedFragmentReference<TReadFromStore, TItem>>;
+
+type LoadedFragmentReference<
+  TReadFromStore extends { parameters: object; data: object },
+  TItem,
+> = ItemCleanupPair<
+  ReferenceCountedPointer<FragmentReference<TReadFromStore, TItem>>
+>;
+
+function flatten<TItem>(
```
rbalicki2, line 57: since this is run on every render, can we make this use a for loop instead of reduce? That will be more efficient (i.e. fewer intermediate objects)

In addition, can we simplify what's going on with the `pageInfo`? That seems to come from the last item, always, so can we just do `arr[arr.length - 1]?.pageInfo ?? { hasNextPage: true, endCursor: null }` where necessary?

libs/isograph-react/src/loadable-hooks/usePagination.ts
rbalicki2: I would probably map it, so that it's consistent with the shape of `results`. I'm not 1000% sure.

rbalicki2 (changes requested):
looks amazing!!! I'll do some additional manual testing, but it looks ready to land. Let's just replace the reduce with a loop and remap the fragment, and we're g2g

PatrykWalach:
What are use cases for `pendingFragment`?
If I use it to suspend, then after rerender `usePagination` will return `kind: "Complete"` so it's not possible to read value of the fragment.

Wouldn't fragment with return type `void` be a better fit.

rbalicki2:
Sorry that this took so long for me to look at deeply.

A few nits, and it should be landable.

* Happy with leaving the pending fragment as-is.
* In UserRepositoryList, we no longer need to reverse the results we get. They should instead come in the correct order, since we're paginating forward. i.e. we want this to show things newest-to-oldest.
* Second, my thinking is that if we call `pagination.fetchMore(undefined, 1)` 2x and `pagination.fetchMore(undefined, 2)` once, the results should be the same. But they're actually reversed in the latter case.
* I think the reason this is happening is that we are attempting to reverse the natural order of the GitHub connection. If, instead, we do:

```
  field User.RepositoryConnection($first: Int, $after: String) {
    repositories(first: $first, after: $after) {
      pageInfo {
        hasNextPage
        endCursor
```

we're good (except for the stupid query/mutation mixup, which I'll work on today)

rbalicki2:
awesome, merged and landed! 3a0c06c030f4493715de92bd040c11a97c2d8e67

PR #162, opened 2024-10-07
return multiple errors

### LINKED ISSUE #135: Return multiple errors from validate_client_field_definition_selections_exist_and_types_match
* we can report lots of errors, but we stop when we encounter the first error.

PatrykWalach:
`validate_no_extraneous_arguments` and `assert_no_missing_arguments` should be collected as well

Close #135

### DIFF
--- a/crates/isograph_schema/src/validate_schema.rs
+++ b/crates/isograph_schema/src/validate_schema.rs
@@ -286,6 +286,18 @@ fn get_all_errors_or_all_ok<T, E>(
     }
 }
 
+fn get_all_errors_or_tuple_ok<T1, T2, E>(
+    a: Result<T1, Vec<E>>,
+    b: Result<T2, Vec<E>>,
+) -> Result<(T1, T2), Vec<E>> {
+    match (a, b) {
+        (Ok(v1), Ok(v2)) => Ok((v1, v2)),
+        (Err(e1), Err(e2)) => Err(e1.into_iter().chain(e2.into_iter()).collect()),
+        (_, Err(e)) => Err(e),
+        (Err(e), _) => Err(e),
+    }
+}
+
 fn get_all_errors_or_all_ok_iter<T, E>(
     items: impl Iterator<Item = Result<T, impl Iterator<Item = E>>>,
 ) -> Result<Vec<T>, Vec<E>> {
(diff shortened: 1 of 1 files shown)


crates/isograph_schema/src/validate_schema.rs
```diff
@@ -286,6 +286,18 @@ fn get_all_errors_or_all_ok<T, E>(
     }
 }
 
+fn get_all_errors_or_tuple_ok<T1, T2, E>(
```
rbalicki2, line 289:
```
fn get_all_errors_or_tuple_ok<T1, T2, E, E1: IntoIterator<Item = E>, E2: IntoIterator<Item = E>>(
    a: Result<T1, E1>,
    b: Result<T2, E2>,
) -> Result<(T1, T2), Vec<E>> {
    match (a, b) {
        (Ok(v1), Ok(v2)) => Ok((v1, v2)),
        (Err(e1), Err(e2)) => Err(e1.into_iter().chain(e2.into_iter()).collect()),
        (_, Err(e)) => Err(e.into_iter().collect()),
        (Err(e), _) => Err(e.into_iter().collect()),
    }
}
```
we can avoid an allocation by allowing the error type to impl `IntoIterator`

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -505,15 +518,40 @@ fn validate_client_field_selection_set(
...
+            .map_err(|errs| {
+                errs.into_iter()
+                    .map(|err| {
+                        validate_selections_error_to_validate_schema_error(
+                            err,
+                            parent_object,
+                            unvalidated_client_field.name,
+                        )
+                    })
+                    .collect::<Vec<_>>()
```
rbalicki2, line 530: we can avoid collecting here as a result

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -564,8 +586,16 @@ fn validate_use_refetch_field_strategy(
...
+    .map_err(|errs| {
+        errs.into_iter()
+            .map(|err| {
+                validate_selections_error_to_validate_schema_error(
+                    err,
+                    parent_object,
+                    client_field_name,
+                )
+            })
+            .collect::<Vec<_>>()
```
rbalicki2, line 598: I don't really see how to avoid collecting here, though, without specifying some ugly type

rbalicki2 (changes requested):
just a minor nit, more of a "it's fun to make our code as efficient as possible" rather than an actual complaint about the code

PatrykWalach:
This should be good to go now, unless I should do something about `validate_no_extraneous_arguments` and `assert_no_missing_arguments` here

PR #165, opened 2024-10-08
change loadable field type generation

ch1ffa:
This PR fixes `@ts-error` we temporarily added in #147.

Instead of calculating `missing_arguments` we now use generated `__parameters` and omit provided arguments. Provided arguments are generated inline for now. Probably can be put in a separate file later.

### DIFF
--- a/crates/graphql_artifact_generation/src/import_statements.rs
+++ b/crates/graphql_artifact_generation/src/import_statements.rs
@@ -51,3 +51,18 @@ pub(crate) fn param_type_imports_to_import_statement(
     }
     output
 }
+
+pub(crate) fn param_type_imports_to_import_param_statement(
+    param_type_imports: &ParamTypeImports,
+) -> String {
+    let mut output = String::new();
+    for type_and_field in param_type_imports.iter() {
+        output.push_str(&format!(
+            "import {{ type {}__param }} from '../../{}/{}/param_type';\n",
+            type_and_field.underscore_separated(),
+            type_and_field.type_name,
+            type_and_field.field_name,
+        ));
+    }
+    output
+}
--- a/crates/isograph_schema/src/validate_schema.rs
+++ b/crates/isograph_schema/src/validate_schema.rs
@@ -1165,6 +1165,25 @@ pub fn get_missing_arguments<'a>(
         .collect()
 }
 
+pub fn get_provided_arguments<'a>(
+    argument_definitions: impl Iterator<Item = &'a ValidatedVariableDefinition> + 'a,
+    arguments: &[WithLocation<SelectionFieldArgument>],
+) -> Vec<ValidatedVariableDefinition> {
+    argument_definitions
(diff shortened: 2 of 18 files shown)


crates/graphql_artifact_generation/src/import_statements.rs
```diff
@@ -51,3 +51,18 @@ pub(crate) fn param_type_imports_to_import_statement(
     }
     output
 }
+
+pub(crate) fn param_type_imports_to_import_param_statement(
```
ch1ffa, line 55: Maybe we should consolidate generated type names and filenames so we can use just one function here and pass `RESOLVER_PARAM_TYPE` for example

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -1165,6 +1165,25 @@ pub fn get_missing_arguments<'a>(
         .collect()
 }
 
+pub fn get_provided_arguments<'a>(
+    argument_definitions: impl Iterator<Item = &'a ValidatedVariableDefinition> + 'a,
+    arguments: &[WithLocation<SelectionFieldArgument>],
+) -> Vec<ValidatedVariableDefinition> {
+    argument_definitions
+        .filter_map(|definition| {
+            let user_has_supplied_argument = arguments
```
ch1ffa, line 1174: We didn't treat optional arguments and arguments with default values as missing. But should we? Seems like it's ok to allow users to pass optional arguments that are not provided as literals to `@loadable`

demos/pet-demo/src/components/PetCheckinsCard.tsx
```diff
@@ -50,7 +50,7 @@ export const CheckinDisplay = iso(`
 });
 
 export const PetCheckinsCardList = iso(`
-  field Pet.PetCheckinsCardList($skip: Int, $limit: Int) {
+  field Pet.PetCheckinsCardList($skip: Int!, $limit: Int!) {
```
ch1ffa, line 53: useSkipLimitPagination hook expects it to be required

rbalicki2:
Love it!! I will try to review this tomorrow morning, but no promises, maybe it will have to wait until Saturday

demos/pet-demo/src/components/PetCheckinsCard.tsx
rbalicki2: love that this is catching bugs!

rbalicki2 (approved):
Looks amazing!!!

I noticed some inconsistent tabbing in a multi-line string in Rust, but I'll fix that in a quick follow up diff. I'm also going to extract the `{ parameters: object, data: object }` type that shows up everywhere into its own type alias.

I'm so excited about this. This was such a noticeable missing feature. In at least some talks I gave, I couldn't help but allude to it, saying things like "it's all typesafe! Except arguments, but you get the picture" or something.

Anyway, tremendous work.

PR #167, opened 2024-10-09
customize store id

PatrykWalach:
I've added `getDataId` function to `environment`, this is similar to how `relay` does it.
I've also generated `__typenames` on `normalizationAst`, so they can be used to customize id

But this is not working yet, because `readData` is missing the `__typename`?

#166


PatrykWalach:
https://github.com/isographlabs/isograph/blob/3f50636d6d1a968ef0a88bf3f52f2d3ea81b769a/libs/isograph-react/src/core/read.ts#L493
https://github.com/isographlabs/isograph/blob/3f50636d6d1a968ef0a88bf3f52f2d3ea81b769a/libs/isograph-react/src/core/read.ts#L424
I think these will need to be replaced with `environment.getDataId`, but I' missing `__typename`

rbalicki2:
I think it's a great idea for us to not require global IDs. However, I think we should instead make the cache a two-layer cache of typename -> id -> record.

This would require always knowing the typename:

- encoding the typename in the reader AST, if it is known
- auto-selecting when creating the merged AST, and adding it to the normalization AST

How does that sound?

I am also open to making this customizable at compile time (i.e. making it a config option), so that if you do have globally unique IDs, Isograph can be more efficient. (But that realistically is low urgency, but might make for a nice task for someone who wants to pick it up)

libs/isograph-react/src/tests/__isograph/Query/nodeField/entrypoint.ts
```diff
@@ -20,6 +20,7 @@ const normalizationAst: NormalizationAst = [
         { kind: "Variable", name: "id" },
       ],
     ],
+    concreteType: "Economist",
```
PatrykWalach, line 23: This has been manually, changed from "Node" to "Economist"

PatrykWalach:
> encoding the typename in the normalization AST, if we're selecting a field with a concrete type*

this is kind of what I've done, but there's at least one bug there

> auto-selecting when creating the merged AST, and adding it to the normalization AST

this is a part I don't understand, could you clarify this a bit more, what is the merged AST, and what do you mean by auto-selecting typename?

PatrykWalach:
Now this is at the point, where demos are working, but there's still a bug to fix with `concreteType: "Node"`

libs/isograph-react/src/tests/__isograph/Query/nodeField/entrypoint.ts
PatrykWalach: https://github.com/isographlabs/isograph/blob/e5ce3596bd385777f5d26c3cd36e5e59810a94d2/libs/isograph-react/src/tests/nodeQuery.ts#L6-L12

So there's no issue in rust code, this just can't be narrowed down without proper interface support

PatrykWalach:
Now should be ready to merge, if main gets formatted

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -51,21 +52,40 @@ export function garbageCollectEnvironment(environment: IsographEnvironment) {
...
+    if (dataById == null) continue;
+
+    // delete all objects
+    if (!retainedIds[typeName]) {
+      delete environment.store[typeName];
+      continue;
+    }
+
+    for (const dataId in dataById) {
+      if (!retainedIds[typeName].has(dataId)) {
```
rbalicki2, line 66: let's store `retainedIds[typename]` in a local variable, so that we don't need to do the lookup on every loop iteration

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -51,21 +52,40 @@ export function garbageCollectEnvironment(environment: IsographEnvironment) {
...
 function recordReachableIds(
   store: IsographStore,
   retainedQuery: RetainedQuery,
-  mutableRetainedIds: Set<DataId>,
+  mutableRetainedIds: RetainedIds,
 ) {
   recordReachableIdsFromRecord(
     store,
-    store[ROOT_ID],
+    store['Query'][ROOT_ID],
```
rbalicki2, line 88: presumably, this would break if a user had a `schema { query: SomeotherType }` block in their schema, which is a silly thing to do, but allowed and we should handle it

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -101,9 +121,11 @@ function recordReachableIdsFromRecord(
         }
 
         for (const nextRecordId of ids) {
-          const nextRecord = store[nextRecordId];
+          const nextRecord = store[selection.concreteType]?.[nextRecordId];
           if (nextRecord != null) {
-            mutableRetainedIds.add(nextRecordId);
+            (mutableRetainedIds[selection.concreteType] ??= new Set()).add(
```
rbalicki2, line 126: lol let's split this into two separate lines for clarity. But TIL about `??=`, which I love and hate

libs/isograph-react/src/core/read.ts
```diff
@@ -182,6 +201,7 @@ function readData<TReadFromStore>(
               networkRequest,
               networkRequestOptions,
               mutableEncounteredRecords,
+              field.concreteType,
```
rbalicki2, line 204: once we auto select `__typename`, we can do `field.concreteType ?? record.__typename`

libs/isograph-react/src/tests/garbageCollection.test.ts
```diff
@@ -10,32 +11,36 @@ import {
...
 import { nodeFieldRetainedQuery } from './nodeQuery';
 
-const getDefaultStore = () => ({
-  [ROOT_ID]: {
-    me: { __link: '0' },
-    you: { __link: '1' },
-    node____id___0: {
-      __link: '0',
+const getDefaultStore = (): IsographStore => ({
+  Query: {
```
rbalicki2, line 15: I want to generate `@isograph/react` :-P meaning we don't have to hard code `Query`, and can avoid emitting features that don't exist (e.g. if a given reader AST node is never generated, have a simpler `read.ts`). Eventually!

rbalicki2 (changes requested):
Amazing work. Let's discuss this. I'm on board with this, but I want to make sure we're aligned on all the details

libs/isograph-react/src/core/garbageCollection.ts
PatrykWalach: I've solved this by generating `concreteType: "Query" | "Mutation"` for each entrypoint and refetch query.
Now we have separate cache for `Query` and `Mutation`, this is better then a single `__ROOT` right?

rbalicki2: That's really good idea, let me think about it some more and get back to you

PatrykWalach:
There's also an edge case, when someone, nests `Query`, `github` does it.
```graphql
type Query {
  relay: Query!
}
```

We need to normalize this in `__ROOT`

rbalicki2:
Oh yeah, great catch

PatrykWalach:
I think generating `concreteType` for `readerAst` was a mistake, we don't need it if we add `__typename` to every `__link`, when normalizing
It needs to be added for `union` and `interfaces` anyways

rbalicki2:
My instinct is that we should include the concrete type in the reader AST where the type of the linked field is concrete, and select typename where the type of the linked field is abstract

PatrykWalach:
This should be ready now, the `__typenames` are generated when required and `concreteTypes` are null when they should.

I'd like to refactor some variable names, right now, we have `__typename`, `typeName` and `conreteType`.
Which one do you prefer the most. `concreteType` makes a lot sense to me, but if starting variables with `_` is not an issue, then `__typename` is pretty good as well.


I'm also not in love with `NonNullLink` as a name.
A lot of things, like subscription should now contain a `NonNullLink`, so maybe it should be promoted to just `Link` and old `Link` should be changed to `StoreLink` since it's not present outside of the store.
Let me know your thoughts about this too.

```tsx
export type Link = {
  readonly __link: DataId;
  readonly __typename?: TypeName;
};

export type NonNullLink = {
  readonly __link: DataId;
  readonly __typename: TypeName;
};

type AnyChangesToRecordSubscription = {
  readonly kind: 'AnyChangesToRecord';
  readonly callback: () => void;
  readonly recordId: DataId;
  readonly typeName: TypeName;
};
```

rbalicki2:
Amazing! Sorry that I was unable to review it, my computer was bricked and I just got access to my personal laptop last night. But anyway, I should be able to look today

demos/pet-demo/src/components/__isograph/Pet/PetCheckinsCard/__refetch__0.ts
```diff
@@ -62,6 +64,8 @@ const normalizationAst: NormalizationAst = [
 const artifact: RefetchQueryNormalizationArtifact = {
   kind: "RefetchQuery",
   queryText,
+  concreteType: "Mutation",
+  queryType: "Query",
```
rbalicki2, line 68: is this correct?

libs/isograph-react/src/tests/__isograph/Query/subquery/entrypoint.ts
```diff
@@ -0,0 +1,65 @@
...
+      {
+        kind: "Linked",
+        fieldName: "node",
+        arguments: [
+          [
+            "id",
+            { kind: "Variable", name: "id" },
+          ],
+        ],
+        concreteType: null,
```
rbalicki2, line 32: we should have a follow up to remove the `__typename` selection from below. It only really needs to show up in the query text. In the query text, `__typename` will never be aliased, so we can do `selection.concreteType ?? data.__typename`.

#178. Feel free to do it as part of this PR, it should be a relatively small change. But also, happy to have it be a follow up or left for someone else.

libs/isograph-react/src/tests/__isograph/Query/subquery/resolver_reader.ts
```diff
@@ -0,0 +1,47 @@
...
+        kind: "Linked",
+        fieldName: "node",
+        alias: null,
+        arguments: [
+          [
+            "id",
+            { kind: "Variable", name: "id" },
+          ],
+        ],
+        concreteType: null,
```
rbalicki2, line 24: I don't think we need the `__typename` in the reader AST. Instead, we should make the type of the linked field be `NonNullLink`, e.g. `{ Query: { node____id___3: { __link: '3', __typename: 'User' } }` or whatnot

rbalicki2 (changes requested):
let's subdivide this PR:

- first, let's ship changes that add the `__typename` selection to every abstract field in every query text (and possibly also normalization ASTs, if that's annoying to skip)
    - (I'm not convinced we need it for ReaderASTs! Prove me wrong, though)
- then, let's ship the changes that restructure the runtime

It will be a lot easier to ship those changes independently, simply in terms of being confident that things are correct

libs/isograph-react/src/core/cache.ts
```diff
@@ -149,15 +153,20 @@ export function normalizeData(
       variables,
     );
   }
+
+  environment.store[root.concreteType] ??= {};
+  environment.store[root.concreteType]![root.id] ??= {};
```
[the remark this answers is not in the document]
rbalicki2: I had a change of heart: Since we have so many multi-level maps, I guess the `??=` operator is quite useful. So let's just embrace it.

libs/isograph-react/src/core/cache.ts
```diff
@@ -335,15 +352,29 @@ function callSubscriptions(
...
+  set1: EncounteredIds,
+  set2: EncounteredIds,
+): boolean {
+  for (const typeName in set1) {
+    const set2Ids = set2[typeName];
+    if (!set2Ids) {
+      continue;
+    }
+
+    for (const id of set1[typeName]) {
```
rbalicki2, line 365: this will likely perform better if we assign `set1[typeName]` to an intermediate variable and avoid the lookup on every iteration.

Maybe this can be simplified with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/isDisjointFrom. (I'll add a task to figure out polyfills)

rbalicki2: #179

libs/isograph-react/src/core/cache.ts
```diff
@@ -335,15 +352,29 @@ function callSubscriptions(
...
+      if (set2Ids.has(id)) {
+        return true;
+      }
     }
   }
   return false;
 }
 
+export interface EncounteredIds {
```
rbalicki2, line 374: any reason this is an interface instead of a type, and any reason this isn't `Map<TypeName, Set<DataId>>`?

libs/isograph-react/src/core/cache.ts
```diff
@@ -410,7 +444,10 @@ function normalizeDataIntoRecord(
     }
   }
   if (recordHasBeenUpdated) {
-    mutableEncounteredIds.add(targetParentRecordId);
+    mutableEncounteredIds[targetParentRecordId.concreteType] ??= new Set();
```
rbalicki2, line 447: likewise, let's assign this to an intermediate value (`const foo = mutableEncounteredIds[targetParentRecordId.concreteType] ??= new Set(); foo.add(...)`

libs/isograph-react/src/core/cache.ts
```diff
@@ -481,9 +519,21 @@ function normalizeLinkedField(
...
         mutableEncounteredIds,
+        root,
       );
 
-      dataIds.push({ __link: newStoreRecordId });
+      const __typename =
+        astNode.concreteType ?? networkResponseObject[TYPENAME_FIELD_NAME];
+      if (!__typename) {
+        throw new Error(
+          'Unexpected missing __typename in network response when normalizing a linked field.' +
```
rbalicki2, line 529: missing a space bewteen the strings

libs/isograph-react/src/core/cache.ts
```diff
@@ -497,13 +547,26 @@ function normalizeLinkedField(
...
       mutableEncounteredIds,
+      root,
     );
 
+    let __typename =
+      astNode.concreteType ?? networkResponseData[TYPENAME_FIELD_NAME];
+
+    if (!__typename) {
+      throw new Error(
+        'Unexpected missing __typename in network response when normalizing a linked field.' +
```
rbalicki2, line 558: same here, missing a space

libs/isograph-react/src/core/cache.ts
```diff
@@ -481,9 +519,21 @@ function normalizeLinkedField(
...
+        astNode.concreteType ?? networkResponseObject[TYPENAME_FIELD_NAME];
+      if (!__typename) {
+        throw new Error(
+          'Unexpected missing __typename in network response when normalizing a linked field.' +
+            'This is indicative of bug in Isograph.',
+        );
+      }
+      dataIds.push({
+        __link: newStoreRecordId,
+        ...(!astNode.concreteType && { __typename: __typename }),
```
rbalicki2, line 535: these link objects can always have a `__typename` field, here and below

libs/isograph-react/src/core/cache.ts
```diff
@@ -563,32 +631,47 @@ function normalizeNetworkResponseObject(
...
     variables,
     index,
+    queryType,
   );
+  const __typename =
+    astNode.concreteType ?? networkResponseData[TYPENAME_FIELD_NAME];
+
+  if (!__typename) {
+    throw new Error(
+      'Unexpected missing __typename in network response object.' +
```
rbalicki2, line 654: missing space

demos/pet-demo/src/components/__isograph/Pet/PetCheckinsCard/__refetch__0.ts
```diff
@@ -22,11 +22,13 @@ const normalizationAst: NormalizationAst = [
...
+    concreteType: "MakeCheckinSuperResponse",
     selections: [
       {
         kind: "Linked",
         fieldName: "checkin",
         arguments: null,
+        concreteType: null,
         selections: [
           {
             kind: "InlineFragment",
```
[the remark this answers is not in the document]
PatrykWalach: Right now it's correct, because field "chekin" returns "IChekin", and every interface has a `concrete_type: None`

crates/isograph_schema/src/isograph_schema.rs
```diff
@@ -353,6 +367,7 @@ pub struct SchemaObject {
     pub id_field: Option<ServerStrongIdFieldId>,
     pub encountered_fields:
         BTreeMap<SelectableFieldName, FieldDefinitionLocation<ServerFieldId, ClientFieldId>>,
+    pub concrete_type: Option<IsographObjectTypeName>,
```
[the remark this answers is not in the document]
PatrykWalach: Right now it's always one of these two values, 

we could resolve `interface` implemented by a single `type` to a `concrete_type`,
but then implementing the interface by different type would be a breaking change when it shouldn't be

demos/pet-demo/src/components/__isograph/Pet/PetCheckinsCard/__refetch__0.ts
PatrykWalach: so the `queryType` is generated on every entrypoint in case of an edge case where we need to normalize nested `Query`

it would be much better if there was single place to generate this,
the best I can think of right now is to generate this in `iso.ts`


```tsx
import { _createIsographEnvironment } from "@isograph/react"

export function createIsographEnvironment(...args){
  return _createIsographEnvironment(...args, {
    queryType: "Query"
  })
}

```

libs/isograph-react/src/tests/__isograph/Query/subquery/resolver_reader.ts
PatrykWalach: Yes, this is something I've asked about before, and you wanted to generate it in `readerAst`, it's all about some trade offs

Storing `__typename` in store is much simpler for us, it removes a bunch of code,
but it leads to storing more data, if we have an array of concrete types, we store `__typename` N times, instead of once.

libs/isograph-react/src/core/cache.ts
PatrykWalach: This should be good, both `set1[typeName]` and `set2[typeName]`, happens once per iteration of the outer loop

libs/isograph-react/src/core/cache.ts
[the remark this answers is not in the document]
PatrykWalach: I like these shorthands, because when the value on the left is long, then it takes some time to understand that value on the right is the same
```tsx
  cachedComponentsById[
    fragmentReference.root.__typename + ':' + fragmentReference.root.__link
  ] =
    cachedComponentsById[
      fragmentReference.root.__typename + ':' + fragmentReference.root.__link
    ] ?? {};
```

crates/isograph_schema/src/isograph_schema.rs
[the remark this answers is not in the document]
rbalicki2: Ah I see. Okay, happy to keep it as is (which has landed) and reconsider in the future.

libs/isograph-react/src/tests/__isograph/Query/subquery/resolver_reader.ts
rbalicki2: Oh I'm sorry! I was clearly wrong when I said that. Let's store the typename in the store. Sorry about telling you the wrong thing.

If we want, we can later optimize the concrete type situation by also storing a `{ kind: 'AllOneType, typename, records }` struct.

demos/pet-demo/src/components/__isograph/Pet/PetCheckinsCard/__refetch__0.ts
rbalicki2: Sorry, I don't really remember what the edge case is. If we have `foo: Query`, what goes wrong without this?

demos/pet-demo/src/components/__isograph/Pet/PetCheckinsCard/__refetch__0.ts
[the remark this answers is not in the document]
rbalicki2: #183

demos/pet-demo/src/components/__isograph/Pet/PetCheckinsCard/__refetch__0.ts
PatrykWalach: In this case we want `getDataIdOfNetworkResponse` to return `__ROOT`, so the object gets normalized in `store["Query"]["__ROOT"]`,
Right now it will calculate a different key `ParentObject:123.foo` and normalize in `store["Query"]["ParentObject:123.foo"]`,
Every time we read, we will start from `store["Query"]["__ROOT"]` so we will miss those records

libs/isograph-react/src/tests/__isograph/Query/subquery/resolver_reader.ts
PatrykWalach: If I remove `concreteType` from `readerAst`, then `missingFieldHandler` has to return `__typename`
But we can't calculate `__typename` from `id` to implement `missingFieldHandler` for `Query.node` 😖

PatrykWalach: We need to be able yo store just the `__typename` when normalizing

Something like `Link  | MissingLink` 

```tsx
interface MissingLink {
  __typename: TypeName
}
```

rbalicki2: Ah... How would we ever support the Node missing field handler, in that case, if the ID is not globally unique?

You could for example write `node(id: $id) { ... on Actor { ... } ... on User { ... } }` :/

Incidentally, the use of the node field is being discouraged, because it's costly on the backend. Since you don't know the typename of the ID, you potentially have to look in many databases.

I wonder if we can specially handle this case, e.g. if your query is of the form `interface_field { ... on ConcreteType { ... } }`, we can provide that concrete type as a param to the missing field handler.

In any case, for now, let's just assume that the ID is a Pet and solve this properly later.

demos/pet-demo/src/components/__isograph/Pet/PetCheckinsCard/__refetch__0.ts
rbalicki2: ohhhh, hm. I think the optimal solution here is to have a directive on the type that we parse `@globalSingleton` that controls this behavior. We can skip this for now. If someone runs into this issue, we can reprioritize

libs/isograph-react/src/tests/__isograph/Query/subquery/resolver_reader.ts
PatrykWalach: > Ah... How would we ever support the Node missing field handler, in that case, if the ID is not globally unique?

The `node` field doesn't make sense when `ID` is not globally unique.
With this PR it's hard to support `node` field when `ID` is globally unique.

Having a single layer cache, and `getRecordId` would allow to work great with `unique` and `not unique` `ID`, I'm curios what are the reasons for us to use two layer cache?

libs/isograph-react/src/core/cache.ts
```diff
@@ -562,28 +638,40 @@ function normalizeNetworkResponseObject(
...
 ): DataId /* The id of the modified or newly created item */ {
   const newStoreRecordId = getDataIdOfNetworkResponse(
     targetParentRecordLink,
     networkResponseData,
     astNode,
     variables,
     index,
+    queryType,
   );
+  const __typename =
```
rbalicki2, line 652: since this is the only place that we use `.queryType`, let's get rid of it and just hard code it here, along with 

```
// TODO do not hard code this value here
```

I'm willing to bet that no one will have a nested query type, while *also* renaming the query type, before we have a proper solution. I'll sketch out the proper solution in #209

rbalicki2 (changes requested):

libs/isograph-react/src/core/FragmentReference.ts
```diff
@@ -35,7 +35,7 @@ export type FragmentReference<
 export function stableIdForFragmentReference(
   fragmentReference: FragmentReference<any, any>,
 ): string {
-  return `${fragmentReference.root.__link}/TODO_FRAGMENT_NAME/${serializeVariables(fragmentReference.variables ?? {})}`;
+  return `${fragmentReference.root.__link}/${fragmentReference.root.__typename}/TODO_FRAGMENT_NAME/${serializeVariables(fragmentReference.variables ?? {})}`;
```
rbalicki2, line 38: nit: let's put the typename before the `__link` here, since that reflects how the store is organized

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -155,9 +162,8 @@ export function defaultMissingFieldHandler(
     const value = variables?.[variable];
 
     // TODO can we handle explicit nulls here too? Probably, after wrapping in objects
-    if (value != null) {
-      // @ts-expect-error
-      return { __link: value };
+    if (typeof value === 'string') {
+      return { __link: value, __typename: 'Pet' };
```
rbalicki2, line 166: Let's not have a `defaultMissingFieldHandler` if it requires hard coding `Pet` as a result of this 😂

See #211 for a proper solution. We can also hard code pet in the pet-demo missing field handler.

libs/isograph-react/src/core/cache.ts
```diff
@@ -330,7 +339,24 @@ function callSubscriptions(
...
+    }
+
+    if (isDisjointFrom(set1, set2)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+function isDisjointFrom<T>(set1: Set<T>, set2: Set<T>): boolean {
```
rbalicki2, line 359: can we add a `TODO use a polyfill library`

libs/isograph-react/src/core/cache.ts
```diff
@@ -492,13 +550,26 @@ function normalizeLinkedField(
...
+      root,
     );
 
+    let __typename =
+      astNode.concreteType ?? networkResponseData[TYPENAME_FIELD_NAME];
+
+    if (!__typename) {
+      throw new Error(
+        'Unexpected missing __typename in network response when normalizing a linked field. ' +
+          'This is indicative of bug in Isograph.',
```
rbalicki2, line 562: nit: `of a bug` not `of bug`. I see this in three places.

rbalicki2 (commented):
Partial review. Overall, this looks amazing and I'm super stoked about it! When I have a second wind, I'll continue reviewing

libs/isograph-react/src/core/cache.ts
```diff
@@ -337,7 +344,25 @@ function callSubscriptions(
...
+
+    if (isDisjointFrom(set1, set2)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// TODO use a polyfill library
+function isDisjointFrom<T>(set1: Set<T>, set2: Set<T>): boolean {
```
rbalicki2, line 365: should this be `isNotDisjointFrom?`

libs/isograph-react/src/core/cache.ts
```diff
@@ -412,11 +438,26 @@ function normalizeDataIntoRecord(
...
+      targetParentRecordLink.__typename,
+      new Set(),
+    );
+
+    encounteredRecordsIds.add(targetParentRecordLink.__link);
   }
   return recordHasBeenUpdated;
 }
 
+export function insertIfNotExists<K, V>(map: Map<K, V>, key: K, value: V) {
```
rbalicki2, line 452: nit: can we call this `getOrCreate` and rename `value` to `defaultValue`? That might be more intuitive.

Also, it would be more performant if we didn't create a new `Set` every time this was called. Can we make map `Map<K, Set<V>>`, not have a `value` param, and say `result = new Set()` below?

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -51,31 +54,54 @@ export function garbageCollectEnvironment(environment: IsographEnvironment) {
     recordReachableIds(environment.store, query, retainedIds);
   }
 
-  for (const dataId in environment.store) {
-    if (!retainedIds.has(dataId)) {
-      delete environment.store[dataId];
+  for (const typeName in environment.store) {
+    const dataById = environment.store[typeName];
+    if (dataById == null) continue;
```
rbalicki2, line 59: should we throw (eventually, just in dev) if `dataById == null` and `retainedTypeIds != null ? retainedTypeIds.length > 0 : false`?

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -51,31 +54,54 @@ export function garbageCollectEnvironment(environment: IsographEnvironment) {
...
-  recordReachableIdsFromRecord(
-    store,
-    store[ROOT_ID],
-    mutableRetainedIds,
-    retainedQuery.normalizationAst,
-    retainedQuery.variables,
-  );
+  const record =
+    store[retainedQuery.root.__typename]?.[retainedQuery.root.__link];
+  if (record)
```
rbalicki2, line 91: can we have `{` here? And in general, we should have them everywhere, so as to be clear and idiomatic.

Should we throw if the record is not present? Not quite! The record is allowed to `=== null`, but it cannot `=== undefined` (since it's retained.)

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -85,25 +111,47 @@ function recordReachableIdsFromRecord(
...
           const link = assertLink(linkedFieldOrFields);
           if (link != null) {
-            ids.push(link.__link);
+            links.push(link);
           }
         }
 
-        for (const nextRecordId of ids) {
-          const nextRecord = store[nextRecordId];
+        let typeStore = selection.concreteType && store[selection.concreteType];
```
rbalicki2, line 129: `??` instead of `&&`. We should never use `&&`, it's unclear, and it allows for bugs. For example, `concreteType` is `""`, then `typeStore === ""`. This also leaks into the type, `typeStore` has type
```
typeStore: "" | {
    [index: string]: StoreRecord | null;
} | null
```

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -85,25 +111,47 @@ function recordReachableIdsFromRecord(
...
+
+        if (!typeStore && selection.concreteType) {
```
rbalicki2, line 131: let's also be explicit here, `typeStore != null && ...`

rbalicki2 (changes requested):
some minor nits on the runtime side... I'll continue reviewing the rust side as well

rbalicki2 (commented):
rust side lgtm

libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -42,7 +45,7 @@ export function retainQuery(
 }
 
 export function garbageCollectEnvironment(environment: IsographEnvironment) {
-  const retainedIds = new Set<DataId>([ROOT_ID]);
+  const retainedIds: RetainedIds = { Query: new Set<DataId>([ROOT_ID]) };
```
PatrykWalach, line 48: This is not good, because `Query` is hard coded, can we garbage collect it?

libs/isograph-react/src/core/garbageCollection.ts
PatrykWalach: In both of these cases it would make sense to warn in dev about bug in Isograph with garbage collection

libs/isograph-react/src/core/garbageCollection.ts
rbalicki2: Happy to revisit this in the future. I'm not concerned right now. I don't think we would ever garbage collect `Query` or `Query[ROOT_ID]`, because (in the current version) we keep 10 queries in the GC buffer.

That being said, that all needs to be more customizable, but we can revisit it in the future.

Another thing we could do is to treat `Query[ROOT_ID]` as "special", and treat each field on `Query[ROOT_ID]` as its own garbage-collectable object. Because, right now, since everything goes through `Query[ROOT_ID]` we will eventually have a `Query[ROOT_ID]` object with the root fields of every single query you ever made. Oh well!

rbalicki2 (approved):
🎉🎉🎉 phenomenal work! Thank you for splitting this up into smaller PRs and for making the necessary changes 🎉🎉🎉 

I'm glad this is finally shipping!!

libs/isograph-react/src/core/garbageCollection.ts
PatrykWalach: Right now it will `garbageCollect` if different type than `Query` is used

rbalicki2: Oh do we have a bug with the star wars API example in the quickstart then?

PatrykWalach: I've fixed it

rbalicki2: given the change, can we make `const retainedIds: RetainedIds = {}`?

PatrykWalach: Yes, but this required a test change

rbalicki2: nice lgtm

PR #168, opened 2024-10-10
Use vitest workspaces

PatrykWalach:
So I was looking at the `vitest` documentation and noticed it supports workspaces
it allows to run all tests in parallel, `pnpm -r` runs all commands in a "safe" order, which is slower

But it wasn't fully working, so I've updated `vitest`
This version used newer `vite` so `commonjs` imports broke
This was probably the issue `Lane Sawyer` was getting

After that, `vitest` was running in root directory, but the imports were still not working
it turned out, babel was resolving all imports relative to `cwd` so I've changed it to resolve relative to `isograph.config.json`

So with this, the old commands `pnpm -r test` still work, but `pnpm -w test` can be used to go even faster

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,7 +137,7 @@ jobs:
         #
         # This can be overcome, if necessary, e.g. if "@isograph/react" was a
         # parameter passed to the CLI.
-        run: pnpm -r test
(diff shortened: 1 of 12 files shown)


rbalicki2 (approved):
Holy cow this is so much faster!!

PR #169, opened 2024-10-12
docs: Vite demo

lanesawyer:
# docs: Vite Demo

## What
- Adds a Vite example to the `demos` folder that displays Pokemon information
- Updates Quickstart docs page with a callout pointing to the Vite demo if the user wants an example with Vite

## Demo Screenshot
image


demos/vite-demo/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint .",
+    "preview": "vite preview",
+    "iso": "iso --watch"
+  },
+  "dependencies": {
+    "@isograph/react": "^0.2.0",
```
rbalicki2, line 14: let's install these the same way that they're installed in the other demos, that way we can make local changes and immediately see them reflected here. So:

* in the root package.json, add equivalent `watch-vite-demo` and `build-vite-demo` and adjust the other commands
* add stuff to `ci.yml`, namely ensuring that when we run `build-vite-demo` the working directory is clean
* here, `"@isograph/react": "*",` and installed via pnpm

demos/vite-demo/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "vite-demo",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "vite",
```
rbalicki2, line 7: I see `index.js:2 Uncaught ReferenceError: exports is not defined    at index.js:2:23` when I run `pnpm i && pnpm dev`. Is there a step I'm missing?

demos/vite-demo/package-lock.json
```diff
@@ -0,0 +1,3702 @@
+{
```
rbalicki2, line 1: let's remove this file and `pnpm i`

rbalicki2 (changes requested):
Just a few nits! Amazing!!! So happy to add this to the family of demos

demos/vite-demo/package.json
lanesawyer: Yeah, I'm getting the same thing too as I'm trying to shift away from `npm` to using `pnpm` and referencing the `isograph` libraries locally...  I'll try to see what I can dig up online regarding that error message

lanesawyer: Oooh, that error is coming from `libs/isograph-disposable-types/dist`. So maybe I've set up the Vite stuff in a way that's incompatible with the locally built `libs`?

demos/vite-demo/package.json
lanesawyer: Done!

demos/vite-demo/package-lock.json
lanesawyer: Done!

libs/isograph-compiler/cli.js
lanesawyer, line 1: This is weird, I didn't modify this file. Also it's empty?

lanesawyer:
Typecheck CI step is failing and I also forgot to delete a variety of boilerplate files from the Vite create CLI. Will look into those soon, in addition to the demo not working when linked to the libraries locally.

rbalicki2:
If we change `tsconfig.build.json` to have

```
-    "module": "commonjs",
+    "module": "ESNext",
+    "moduleResolution": "Bundler",
```

then the vite demo doesn't error out, but it causes the other demos to break :/

However, that doesn't seem to be the right thing to do with respect to releasing the package.

demos/vite-demo/vite.config.ts
```diff
@@ -0,0 +1,26 @@
...
+        importRules: 'merge',
+      },
+    }),
+  ],
+  resolve: {
+    alias: {
+      '@iso': path.resolve(__dirname, './src/components/__isograph/iso.ts'),
+      '@iso/*': path.resolve(__dirname, './src/components/__isograph/*'),
+    },
+  },
```
PatrykWalach, line 25: This will make sure, `vite` transpiles isograph from `commonjs` to `esm` with `esbuild`.
```suggestion
  },
  optimizeDeps: {
    include: ['@isograph/react'],
  },
```

demos/vite-demo/src/main.tsx
```diff
@@ -0,0 +1,10 @@
+import { StrictMode } from 'react';
+import { createRoot } from 'react-dom/client';
+import App from './App.tsx';
+import './index.css';
+
+createRoot(document.getElementById('root')!).render(
+  <StrictMode>
```
PatrykWalach, line 7: Also, isograph won't run in `StrictMode`.

demos/vite-demo/package.json
lanesawyer: Should work for you now, thanks to @PatrykWalach!

demos/vite-demo/vite.config.ts
lanesawyer: Done, worked great. Thank you!

lanesawyer:
Alright, everything but the typechecking should be good to go! I'm not sure why the app works, the compiler is happily humming along, but I get some errors on the `data.getAllPokemon` query ☹️

demos/vite-demo/src/main.tsx
lanesawyer: Fixed!

rbalicki2:
Screenshot 2024-10-20 at 12 21 56 PM

I see this 😭

rbalicki2:
closed by 484beaa48ff4a5726f4e7b9a022c4571c38e1a78

PR #171, opened 2024-10-15
fix with_span panic

ch1ffa:
This PR fixes a panic when there is an error inside `with_span` do_stuff function and we bail and return it early


rbalicki2 (approved):
shipitt

PR #184, opened 2024-10-18
fix unsupported token panic

### LINKED ISSUE #58: [bug] Commenting out an iso literal should not cause the compiler to panic
* Maybe we can handle this by adding `/` or `//` to the lexer and then just bailing if we encounter it

ch1ffa:
This PR adds a new error token type to handle unexpected single line comment token `//` and collects all other encountered unexpected tokens.

Closes #58

### DIFF
--- a/crates/isograph_lang_parser/src/token_kind.rs
+++ b/crates/isograph_lang_parser/src/token_kind.rs
@@ -75,6 +75,9 @@ pub enum IsographLangTokenKind {
     // Comments
     // #[regex("#[^\n\r]*")]
     // SingleLineComment,
+    #[regex("//")]
+    ErrWrongSingleLineComment,
+
     // Whitespace
     #[token(",")]
     Comma,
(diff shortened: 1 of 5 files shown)


rbalicki2:
I'm still seeing panics when I comment out a line. Maybe something broke between when you tested and pushed or we're missing some changes?

crates/isograph_lang_parser/src/token_kind.rs
```diff
@@ -75,6 +75,9 @@ pub enum IsographLangTokenKind {
     // Comments
     // #[regex("#[^\n\r]*")]
     // SingleLineComment,
+    #[regex("//")]
```
rbalicki2, line 78: you should be able to use `#[token("//")]`, which presumably is faster than using a regex. Source: https://github.com/maciejhirsz/logos

ch1ffa: Ah, my bad, thanks! Was testing which is better: show a token error of the whole "commented" line

ch1ffa: @rbalicki2 done!

rbalicki2:
closed in favor of 9c01b96f6570a8a5599b65240cc1a18464c8110e and d60ff77a6eb5eba2a5db456d642cbc4e2f524831

Issue #173, opened 2024-10-16
fetchMore should not take args (other than count)


rbalicki2:
- we currently pass additional args to fetchMore. That's incorrect, since conceptually, we would be fetching an entirely new list if we pass different args. Oops. e.g. `fetchMore({ order: 'DESC' }, 5)` followed by `fetchMore({ order: 'ASC'}, 5)` is clearly silly.
- so, instead, those params should be passed to hook and used by the initial call to fetchMore, or should be part of the initial state param, or something.

PR #187, opened 2024-10-19
add initial args to pagination

### LINKED ISSUE #172: pagination hooks should take initial state as a param
- the skip limit and connection spec pagination hooks should accept the initial state (initial skip, initial cursor)
- this will enable the following pattern

```
iso(`field Foo.bar {
  initialData: PaginationComponent(limit: 5, skip: 0)
  PaginationComponent @loadable
}`)(({data}) => {
  const paginationState = useSkipLimitPagination(data.PaginationComponent, { skip: 5 })
  const paginatedResults = [...data.initialData, ...paginationState.records];
})
```

- we should use this in various places, such as:
  - fixing newsfeed (where we pass an additional param... lol)
  - the other places we paginate

PatrykWalach:
Somehow the github demo stopped working

Closes #172, #173


libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -20,10 +20,27 @@ import { useIsographEnvironment } from '../react/IsographEnvironmentProvider';
 import { useSubscribeToMultiple } from '../react/useReadAndSubscribe';
 import { maybeUnwrapNetworkRequest } from '../react/useResult';
 
-type FirstOrAfter = 'first' | 'after';
-type OmitFirstAfter<TArgs> = keyof Omit<TArgs, FirstOrAfter> extends never
-  ? void | Record<string, never>
```
PatrykWalach, line 25: We don't need `Record<string, never>` in `args`, because there's always at least one key `after` or `skip`

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -20,10 +20,27 @@ import { useIsographEnvironment } from '../react/IsographEnvironmentProvider';
...
+  [K in Exclude<keyof TArgs, After>]: TArgs[K];
+};
+
+type Prettify<T> = {
+  readonly [K in keyof T]: T[K];
+} & {};
+
+// if none of the keys are required initialArgs will be optional
+type MakeOptionalArgs<TArgs> = {} extends TArgs
+  ? [initialArgs?: TArgs]
```
PatrykWalach, line 38: `void | TArgs` requires passing `undefined`, `[initialArgs?: TArgs]` allows for no extra argument

rbalicki2: neat trick!

rbalicki2:
Looking into this; usually the reason a demo breaks is due to indefinite suspense. What you can do is load the demo on a commit that works, then checkout the broken commit. Then at least next js will render the page, lol, and you'll see that we get stuck on a suspending screen. I don't know how to turn off this behavior, it's very frustrating because this is a frequent source of bugs!

The underlying issue is (usually) that we write the network response into the store under key A, but then try to read that field from the store using a different key B.

I haven't confirmed that that's what's happening here, but it's my top hypothesis

rbalicki2:
the fix for the github demo is fixed on master

PatrykWalach:
The demos are working, this should be ready to merge

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -179,10 +195,7 @@ export function useConnectionSpecPagination<
...
   const getFetchMore =
     (after: string | null | undefined) =>
-    (
-      args: OmitFirstAfter<TReadFromStore['parameters']>,
-      first: number,
-    ): void => {
+    (first: number): void => {
       // @ts-expect-error
       const loadedField = loadableField({
         ...args,
```
rbalicki2, line 201: Since `fetchMore` is only called once (in response to a click, for example), and not on every render, I'm not super concerned about perf. But, I do think we can probably optimize this by avoiding spreading args unless this is the first call to `fetchMore` (i.e. if `state.length === 0`)

(Feel free to do that as a follow up, as well.)

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -20,10 +20,27 @@ import { useIsographEnvironment } from '../react/IsographEnvironmentProvider';
...
+  readonly [K in keyof T]: T[K];
+} & {};
+
+// if none of the keys are required initialArgs will be optional
+type MakeOptionalArgs<TArgs> = {} extends TArgs
+  ? [initialArgs?: TArgs]
+  : [initialArgs: TArgs];
+
+type UseConnectionSpecArgs<TArgs> = MakeOptionalArgs<
+  Prettify<MakeAfterOptional<Omit<TArgs, First>>>
```
rbalicki2, line 42: concretely, what does `Prettify` get us here?

rbalicki2 (changes requested):
Amazing work! Just a few nits with respect to eking out performance. If you want to land this as-is, I'm happy with that, and we can do all of these as follow ups.

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
PatrykWalach: User won't see type `MakeAfterOptional<Omit<TArgs, First>`, but an "expanded" type, such as:
```tsx
{
  skip: number,
  search: string
}
```
not sure if we care about this

rbalicki2: sgtm

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
PatrykWalach: we can use arguments[1] to not allocate array, and then cast the hook to type the variable 🤔

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -92,7 +106,9 @@ export function useConnectionSpecPagination<
   TItem,
 >(
   loadableField: LoadableField<TReadFromStore, Connection<TItem>>,
+  ...[initialArgs]: UseConnectionSpecArgs<TReadFromStore['parameters']>
 ): UsePaginationReturnValue<TReadFromStore, TItem> {
+  const [args] = useState(initialArgs);
```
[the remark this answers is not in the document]
PatrykWalach: If we pass extra arguments, like filters, then we need them for next pages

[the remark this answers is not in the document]
rbalicki2: Three thoughts:

- Good point, we certainly don't correctly handle other params at the moment. Maybe we punt for now on additional args, and just have you pass additional args via the isograph literal, since we should be recreating everything if the args change, and that requires that `change -> uncommited state` hook.
- Right now, if you pass different variables via the isograph literal, all the state gets recreated, so maybe that is sufficient for now. (I haven't thought this through all the way.)
- I think that this hook should take three params, the loadable field, the initial state, and the additional args that are passed to each `fetchMore`

libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts
```diff
@@ -285,3 +284,35 @@ export function useSkipLimitPagination<
...
+      readonly parameters: {
+        readonly search: string;
+        readonly skip: number;
+        readonly limit: number;
+      };
+    },
+    object[]
+  >;
+  // @ts-expect-error
+  useSkipLimitPagination(advancedLoadable);
```
PatrykWalach, line 317: `ExtractParameters` and `ExtractData` needs to be fixed in seperate PR

libs/isograph-react/src/core/reader.ts
```diff
@@ -144,7 +144,7 @@ type StableId = string;
 export type LoadableField<
   TReadFromStore extends { data: object; parameters: object },
   TResult,
-  TProvidedArgs extends object = object,
+  TArgs = ExtractParameters<TReadFromStore>,
```
PatrykWalach, line 147: So I've this to have more control, now `useSkipLimitPagination` can say it takes
```tsx
type UseSkipLimitPaginationArgs = {
  skip: number;
  limit: number;
};
```
and everything will be type checked nicely

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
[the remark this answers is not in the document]
PatrykWalach: This makes sens to me too, I thought if we go this way, then user will have to add `key={}` to reset components.
Great thing you've noticed it will recreating everything.

libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts
PatrykWalach: This got fixed by changing `LoadableField`

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -36,10 +53,7 @@ type UsePaginationReturnValue<
     }
   | {
       kind: 'Complete';
-      fetchMore: (
-        args: OmitFirstAfter<TReadFromStore['parameters']>,
-        first: number,
-      ) => void;
+      fetchMore: (first: number) => void;
```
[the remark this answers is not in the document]
PatrykWalach: Done

libs/isograph-react/src/core/reader.ts
rbalicki2: nice, what was not type checking before?

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
rbalicki2: sgtm

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -81,17 +73,26 @@ type NonNullConnection<T> = {
...
   TItem,
 >(
-  loadableField: LoadableField<TReadFromStore, Connection<TItem>>,
+  loadableField: LoadableField<
+    TReadFromStore,
+    Connection<TItem>,
+    UseConnectionSpecPaginationArgs
+  >,
+  initialArgs?: {
+    after?: string | void | null;
```
rbalicki2, line 94: let's add `hasNextPage` here as well. Also, is `initialArgs?: { after: string } | null` a better type for this?

rbalicki2 (commented):
Let's just include `hasNextPage` in the initial args and this looks g2g.

Just realized, as a follow up, maybe the return data should have three states: pending, complete and noMoreRecords #198

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
PatrykWalach: Then user has to pass `firstPage.pageInfo.endCursor && { after: firstPage.pageInfo.endCursor }`

libs/isograph-react/src/core/reader.ts
PatrykWalach: This wasn't working without `@ts-expect-error`
https://github.com/isographlabs/isograph/blob/6bd3135f11b67c408bac8ec5118cf5e0a598049e/libs/isograph-react/src/loadable-hooks/useSkipLimitPagination.ts#L162-L167
Typescript said `skip` and `limit` were not existing, although types where
```tsx
(parameter) loadableField: (args: void |
Omit<TReadFromStore["parameters"], keyof
Omit<TReadFromStore["parameters"], "skip" | "limit">>) => [...]
```
and `TReadFromStore["parameters"]` extended `UseSkipLimitPaginationArgs`, so it should have, `skip` and `limit` keys, somehow `Omit` made it impossible to infer the type.
But this wasn't a big deal.

## Bigger issue was Covariance
Now the parameters are `covariant`.
If user tries to pass `LoadableField` that somehow takes
```tsx
{
  skip: number | string | SomeOther,
  limit: number | null
}
```
(creates a wrapper, have custom scalar or widens types?) 

then it's not an issue, because `number` is assignable to `number | string | SomeOther`.

Previously `parameters` would be used in a return type of `loadable`; as a result type of `parameters` was `contravariant`.
We would have to predict all the possible values `skip` and `limit` can take
```tsx
interface UseSkipLimitPaginationArgs {
  skip: number | null | string | AnyOther,
  limit: number | null
}
```
Inside of `useSkipLimitPagination` we don't care directly about the type of parameters returned by `loadableField`.
By splitting `parameters` into arguments and return type in `LoadableField` we can have `contravariant` return type and `covariant` arguments.

With the default `TArgs = ExtractParameters<TReadFromStore>`, none of the call sites had to be changed.
But when we want to provide arguments, the logic have to be done one level higher, that's why change in rust has been required.
```rs
format!(
    ",\n{indent}Omit<ExtractParameters<{}__param>, keyof {}>",
    client_field.type_and_field.underscore_separated(),
    get_loadable_field_type_from_arguments(
        schema,
        provided_arguments
    )
)
```

rbalicki2: Okay, awesome, thanks for digging into that and figuring this out!!

> Previously parameters would be used in a return type of loadable;

where? Is that because parameters was returned as part of `FragmentReference`?

PatrykWalach: Yes, the fragment includes parameters

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
PatrykWalach: I think it's up to you to decide
- a lot of users will want an initial page, so it makes sens to take `pageInfo`,
- tbh. I like what we have now the most, I think it's a good low level api, and to me the argument `after` is more readable then `endCusor`,
- overall it shouldn't matter much what we go for, as long as user can make a custom hook around this, then he can even take an array of initial connections and merge those.

rbalicki2: yeah, I think what draws me to passing `pageInfo` is:
- that is the type of the hooks internal state (in the same way that skip is the state of the use skip limit hook. Skip also happens to be the parameter there, but I'd make the case that it would make sense to also add `hasNextPage` to the state there, corresponding to whether the last request was empty.)
- you can just pass in the `pageInfo` directly, so you don't have to actually type out `endCursor` and `hasNextPage`.
- the hook is lying when it says `hasNextPage: true` if we know that the previous page info had `hasNextPage: false` :/

rbalicki2: In this specific case, it would also work to take the entire connection. Then, the user wouldn't have to concatenate. But that only works because the return value includes the page info, which is the state. In the skip/limit pagination case, we don't return the `fetchedSoFar` field, so that's not possible :/

PatrykWalach: Do we want `Optional<PageInfo>`?

rbalicki2: What would the use case be for that?

rbalicki2 (approved):
Hell yeahhhh shipitttt, thanks for doing this! And for bearing with me through all the unrelated changes leading to merge conflicts and for incorporating all the feedback

PR #189, opened 2024-10-20
publish compiler for win64

PatrykWalach:
Seems like we are building the compiler for win 64, but not publishing it


rbalicki2:
great catch!!! but also lol

PR #196, opened 2024-10-21
refactor handle_compile_command

### LINKED ISSUE #174: break up handle_compile_command
Let's chat about this before you start, I want to make sure we're on the same page.

`handle_compile_command` should be broken up as follows:

- there should be a `create_validated_schema_from_sources` function that accepts the parsed isograph literals, and the parsed graphql schema and extensions (specifically, the "Outcome" object)
- the parsed isograph literals and parsed schema and extensions should be data structures that are wrapped in newtypes, and which contain a `new` method that returns a `Result<Self, BatchCompileError>`. This `new` method should **not** take an `&mut IsographSchema`!!!!
    - the parsed isograph literals struct should probably be `struct ParsedIsographLiterals(HashMap<FilePath, HashSet<ParsedIsographLiterals>>)`. This is because we're going to modify this data structure on a per-file basis in #175 .
- `handle_compile_command` should create these data structures and pass them to the `create_validated_schema_from_sources`, and that validated schema should be used to create and write the artifacts.

ch1ffa:
Closes #174

### DIFF
--- a/crates/isograph_compiler/src/batch_compile.rs
+++ b/crates/isograph_compiler/src/batch_compile.rs
@@ -1,35 +1,21 @@
-use std::{path::PathBuf, str::Utf8Error};
+use std::{collections::HashMap, path::PathBuf, str::Utf8Error};
 
 use colored::Colorize;
-use common_lang_types::{
-    FilePath, Location, SourceFileName, Span, TextSource, WithLocation, WithSpan,
-};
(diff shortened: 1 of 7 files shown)


rbalicki2 (commented):
Just took a quick look, overall, this looks great. Thank you for tackling this!

rbalicki2 (approved):
Amazing, lgtm! Shipittt! Now let's make watch mode fast :)!!!

PR #203, opened 2024-10-23
chore: Move from cargo-watch to bacon

### LINKED ISSUE #195: Migrate off of cargo-watch
https://github.com/watchexec/cargo-watch?tab=readme-ov-file#maintenance

We should use `watch-exec` directly, or `bacon`, or find some other solution.

lanesawyer:
# chore: Move from cargo-watch to bacon

## What
- Closes #195 by moving `cargo-watch` to `bacon`, the tool blessed by the creator of `cargo-watch`
- Slightly changes the developer experience. The console output isn't as pretty for the `pnpm` commands, and `bacon`  _completely_ takes over the terminal while running. All of it seems to work the same, **although I'd appreciate a bit of testing from someone with more experience working on Isograph before it goes in!**
- Updated the documentation to encourage developers to install `bacon`

## How
1. Wrote down what the flags all did in `cargo-watch`:
- `-x` means command to execute on change
- `-d` is file update debounce time in seconds (does not have `bacon` equivalent)
- `-w` is watch directory
- `-s` is shell command to execute on change (does not have flag equivalent, jobs are defined in `bacon.toml`)

2. Found matching ones (where the existed) in `bacon`:
- `-j` is job to launch (rough equivalent of `-x` and `-s`)
- `-p` is path to watch (equivalent of `-w`)

3. Generated the `bacon.toml` with `bacon --init`

4. Added the `build` and `watch-libs` jobs to capture the same functionality as `pnpm run watch-rs` and `pnpm run watch-libs`. There is no native `build` command for `bacon` (it just has `check`) and they [recommended making a custom job](https://github.com/Canop/bacon/issues/238#issuecomment-2407138772)

5. Updated the commands in `package.json`

### DIFF
--- a/bacon.toml
+++ b/bacon.toml
@@ -1,0 +1,120 @@
+# This is a configuration file for the bacon tool
+#
+# Bacon repository: https://github.com/Canop/bacon
+# Complete help on configuration: https://dystroy.org/bacon/config/
+# You can also check bacon's own bacon.toml file
+#  as an example: https://github.com/Canop/bacon/blob/main/bacon.toml
+
+default_job = "check"
+
+# Build and watch-libs are custom jobs, the rest of the file is from the Bacon default configuration
+[jobs.build]
+command = [
+	"cargo", "build",
+	"--color", "always",
+]
+need_stdout = false
+
+[jobs.watch-libs]
+command = ["pnpm", "-r", "compile"]
+need_stdout = true
+
+[jobs.check]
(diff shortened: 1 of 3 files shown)


rbalicki2 (approved):
LGTM! Thank you!!! This is fun, I get to learn about bacon now

rbalicki2:
FWIW my testing involved running it and seeing that the behavior matched what is described in https://dystroy.org/bacon/ :-D

PR #214, opened 2024-10-25
parallelize `build-demos` and `tsc`

PatrykWalach:
I've moved iso build scripts to corresponding packages, and used `--parallel` to run them in parallel, `--aggregate-output` makes sure the logs are grouped and easier to read.


rbalicki2 (approved):
Awww yeah so fast!!! Shipit

PR #215, opened 2024-10-25
fix `tsconfig` exclude

PatrykWalach:


tsconfig.build.json
```diff
@@ -1,5 +1,5 @@
 {
-  "exclude": ["**/*.test.ts", "**/*.stub.ts", "node_modules", "dist"],
+  "exclude": ["**/*.test.ts", "**/*.stub.ts", "**/node_modules", "**/dist"],
```
PatrykWalach, line 2: 1. `dist` means, `dist` directory in root of the project, `**/dist` means any `dist` directory.
Without this running `compile-watch` fails, because it tries to compile from `dist` directory.

libs/isograph-react/tsconfig.pkg.json
```diff
@@ -8,5 +8,5 @@
     "lib": ["es2017", "DOM"]
   },
   "include": ["./**/*.ts", "./**/*.tsx"],
-  "exclude": ["./src/tests/**/*", "./vitest.config.ts"]
+  "exclude": ["./src/tests/**/*", "./vitest.config.ts", "**/node_modules", "**/dist"]
```
PatrykWalach, line 11: 2. We `extend` from base `tsconfig`, but setting `exlude`, completely overrides extended `exclude`,
so we need to add these manually.

rbalicki2 (approved):
SGTM

PR #216, opened 2024-10-25
add `recordLink` argument in `check`

PatrykWalach:


libs/isograph-react/src/core/check.ts
```diff
@@ -151,8 +146,7 @@ function checkFromRecord(
               normalizationAstNode.selections,
               variables,
               linkedRecord,
-              // TODO this seems likely to be wrong
-              backupId + '.' + parentRecordKey,
+              link,
```
PatrykWalach, line 149: here, and in other place `linkedRecord` is `environment.store[link.__link]`, so `link` makes sense here

rbalicki2 (approved):
Good catch, thanks for making this change!

PR #217, opened 2024-10-25
handle nullable lists when normalizing

PatrykWalach:


libs/isograph-react/src/core/cache.ts
```diff
@@ -127,8 +127,8 @@ type NetworkResponseValue =
   | NetworkResponseScalarValue
   | null
   | NetworkResponseObject
-  | NetworkResponseObject[]
-  | NetworkResponseScalarValue[];
+  | (NetworkResponseObject | null)[]
+  | (NetworkResponseScalarValue | null)[];
```
PatrykWalach, line 131: I've noticed we expect lists to be not nullable

libs/isograph-react/src/core/cache.ts
```diff
@@ -596,35 +601,29 @@ function normalizeNetworkResponseObject(
...
-): data is NetworkResponseScalarValue | NetworkResponseScalarValue[] {
-  // N.B. empty arrays count as empty arrays of linked fields.
+function isNullOrEmptyArray(data: unknown): data is never[] | null[] | null {
   if (Array.isArray(data)) {
     if (data.length === 0) {
-      return false;
+      return true;
     }
-    // This is maybe fixed in a new version of Typescript??
-    return (data as any).every((x: any) => isScalarOrEmptyArray(x));
```
PatrykWalach, line 621: Keeping this as is would return true for list of scalars

libs/isograph-react/src/core/cache.ts
```diff
@@ -461,17 +461,24 @@ function normalizeLinkedField(
     return existingValue !== null;
   }
 
-  if (isScalarButNotEmptyArray(networkResponseData)) {
+  if (
+    isScalarOrEmptyArray(networkResponseData) &&
+    !isNullOrEmptyArray(networkResponseData)
```
PatrykWalach, line 466: 3. So I just do two checks here.

rbalicki2 (approved):
LGTM

Issue #53, opened 2024-04-11
[Epic] Client fields that return IDs


rbalicki2:
* Client fields currently are scalars, in the sense that you cannot use `{}`, as in `my_client_field { ... }`.
* Client fields should be able to return IDs, in which case, you would be able to do `best_friend_client_field { name }`
* This needs additional thought:
  * How do we enable lazy fetching? As in, we calculate `best_friend_client_field`, and then execute the operation `node(id: $bestFriendId) { ... on User { name } }`
  * Do we want to force the user to opt into this? (As in the `@waterfall` directive in Relay?)

rbalicki2:
an open question: if we're going to use these for graphql type refinment (which IMO is a good idea and is the right abstraction), then the (auto-generated) pointer would look like this, if it were user written:

```
# note the optional User type here... "to User", not "to User!"
iso(`pointer Node.asUser to User {
  __typename
  id
}`)(({data}) => {
  if (data.__typename === 'User') {
    return id;
  }
})
```

Given this, it would be natural for us to treat `asFoo` as a loadable field (ish), i.e. when it is read, it must be passed to `useClientSideDefer`, then the data is fetched with a separate request.

That's fine! It's a good v1! It's something we should implement anyway. But we need some way to indicate that this field can be merged into the parent (via `... on User`). So, probably, there will be two types of pointers: mergable and non-mergable. All user-written pointers will be non-mergable (because, how would we know what to do?) (Maybe when there is a server component, all pointers can be mergable!)

The user should be able to unmerge a mergable pointer, e.g. with an @annotation on the pointer selection.

rbalicki2:
I think we can probably implement user-generated pointers separately from non-user generated pointers, even though they use the same infra. With that in mind, the following are some ideas for subtasks here:

## Tasks for both generated pointers (i.e. inline fragments) and user-defined pointers:

[ ] add `schema.client_pointers: Vec<ClientPointer>`. `FieldType` should have a new variant: `ClientPointer`. Add `ClientPointerId`.
[ ] modify validation to allow ensure client pointers always have selection sets
[ ] modify validation to ensure that client pointers are selected correctly (same rules as fields)
[ ] generate a reader AST node for client pointers (where it is read)
[ ] generate a `Type/PointerName/resolver_reader.ts` file, or something, that includes the reader AST and the resolver. For auto-generated pointers, the resolver will be inlined, otherwise, it will be imported.
[...]

PR #224, opened 2024-10-27
add `ClientPointer` variant

PatrykWalach:
Part of #53

### DIFF
--- a/crates/isograph_schema/src/isograph_schema.rs
+++ b/crates/isograph_schema/src/isograph_schema.rs
@@ -158,28 +158,16 @@ impl<TSchemaValidationState: SchemaValidationState> Schema<TSchemaValidationStat
 /// an iso field literal. Refetch fields and generated mutation fields are
 /// also local fields.
 #[derive(Debug, Clone, Copy)]
-pub enum FieldType<TServer, TClient, TPointer> {
+pub enum FieldType<TServer, TClient> {
     ServerField(TServer),
     ClientField(TClient),
-    ClientPointer(TPointer),
 }
 
-impl<TFieldAssociatedData, TClientFieldType, TClientPointerType>
-    FieldType<TFieldAssociatedData, TClientFieldType, TClientPointerType>
-{
+impl<TFieldAssociatedData, TClientFieldType> FieldType<TFieldAssociatedData, TClientFieldType> {
     pub fn as_server_field(&self) -> Option<&TFieldAssociatedData> {
         match self {
             FieldType::ServerField(server_field) => Some(server_field),
             FieldType::ClientField(_) => None,
-            FieldType::ClientPointer(_) => None,
-        }
-    }
-
-    pub fn as_client_pointer(&self) -> Option<&TClientPointerType> {
-        match self {
-            FieldType::ServerField(_) => None,
-            FieldType::ClientField(_) => None,
-            FieldType::ClientPointer(client_pointer) => Some(client_pointer),
         }
     }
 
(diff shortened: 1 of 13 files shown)


crates/isograph_schema/src/isograph_schema.rs
```diff
@@ -42,6 +42,17 @@ pub trait SchemaValidationState: Debug {
     ///   i.e. DefinedField<ServerFieldId, ClientFieldId>
     type ClientFieldSelectionScalarFieldAssociatedData: Debug;
 
+    /// The associated data type of scalars in client pointers' selection sets and unwraps
```
rbalicki2, line 45: Let's take a second to reconsider... I think I gave you bad instructions.

1. These comments are somewhat outdated: `DefinedField` no longer exists, it's now `ValidatedScalarFieldAssociatedData` (which contains `FieldType`, which is what `DefinedField` has become). Let's change `ObjectId` (since renamed to `ServerObjectId`) to `ValidatedLinkedFieldAssociatedData`
2. I don't think we need an additional type param here. What do you think?

What we're doing when we're validating is that we're going from something that is selected as a scalar (i.e. has no curlies) or something selected as a linked field (has curlies), and going from no information other than the name of the field to an enum of all the possible things it could be. (Plus some other data that we can only get while validating... but let's ignore that for now.)

So, a scalar selection goes from no info to either a client field or a server scalar field.

A linked selection goes from no info to (currently) just an `ServerObjectId`. But now it should go to an enum with either `ServerObjectId` (i.e. server linked field) or client pointer.

Soooo.... I don't think `FieldType` needs a third variant. Instead, as a first step, we should replace `ServerObjectId` in `ValidatedLinkedFieldAssociatedData` to a `FieldType<ServerObjectId, ClientPointerId>`

----

I think that's the correct move, given our current setup.

crates/isograph_schema/src/unvalidated_schema.rs
```diff
@@ -47,6 +53,11 @@ pub type UnvalidatedVariableDefinition = VariableDefinition<
     <UnvalidatedSchemaState as SchemaValidationState>::VariableDefinitionInnerType,
 >;
 
+pub type UnvalidatedClientPointer = ClientPointer<
+    <UnvalidatedSchemaState as SchemaValidationState>::ClientPointerSelectionScalarFieldAssociatedData,
+    <UnvalidatedSchemaState as SchemaValidationState>::ClientPointerSelectionLinkedFieldAssociatedData,    <UnvalidatedSchemaState as SchemaValidationState>::VariableDefinitionInnerType,
```
rbalicki2, line 58: spacing looks weird. Sometimes when the line is too long, `cargo fmt` fails to pick up on it. Can you fix this manually?

rbalicki2 (commented):
Overall, other than the `FieldType` changes, this looks good. `ClientPointer` looks like what I would expect, etc. I like that we're getting the types right and then putting a bunch of `todo!()`'s throughout the codebase, which we can fill in as we go.

PatrykWalach:
When we generate `client_pointers` for each subtype, we will need to push pointer to `encountered_fields`?
So do we need to change `encountered_fields` as well?
https://github.com/isographlabs/isograph/blob/92dc4eb75457522fdbff1eac55bccc6e768d7309/crates/isograph_schema/src/isograph_schema.rs#L373-L382

rbalicki2:
Yes, we would need to add it to encountered fields

PatrykWalach:
I think we need to make a new `ClientFieldVariant` with `pointer_id`

PR #225, opened 2024-10-27
refactor watch mode

### LINKED ISSUE #175: Use broken up handle_compile_command parts to make watch mode more efficient
After #174, we can modify watch mode as follows:

- don't just recompile everything
- instead, create the data structures **and clone them**, and then pass the cloned versions to `create_validated_schema_from_sources` and so on.
- when we get an event that a path has changed, determine whether the changed paths includes any schema files, in which case, re-parse the schema. Likewise, if a path in the components directory changed, recreate the hashmap of parsed isograph literals.
- pass this to `create_validated_schema_from_sources` and generate artifacts

----

- note: instead of re-creating the hashmap of parsed isograph literals if a single isograph literal changes, instead, we can just mutate the hash map of parsed isograph literals! That can be done as part of this, or as a follow up

ch1ffa:
Closes #175 

Not handled yet:
- [x] Rename/move and delete directories
- [x] Watch isograph config file changes


crates/isograph_compiler/src/watch.rs
```diff
@@ -32,16 +35,17 @@ pub async fn handle_watch_command(
             .expect("Failing when watching schema extension");
     }
 
+    let mut state = CompilerState::new(config);
+    compile_and_print(&mut state, &Vec::new());
```
rbalicki2, line 39: `&[]` instead of `&Vec::new()`, but more to the point, `compile_and_print` should take `Option<changes>`.

In addition, it would be nice to use the type system to ensure that `source_files.read_and_parse_all` is called before `validate_and_create_artifacts` is called. (Maybe we can also ensure that `read_and_parse_all` is called before `update`, but that's a little less clear in the language server case.) Maybe we can try:
- `CompilerState.source_files` is `Option<SourceFiles>`, and the only way to make it `Some` is to call `read_and_parse_all`
- When we get to the language server, we can consider a "better" (but more verbose) solution, `enum SourceFileState { PartiallyLoaded(PartiallyLoadedSourceFiles) FullyLoaded(FullyLoadedSourceFiles) }`, where only `FullyLoadedSourceFiles` has a `.validate_and_generate_artifacts` method.

crates/isograph_compiler/src/batch_compile.rs
```diff
@@ -26,41 +20,7 @@ pub struct CompilationStats {
...
-                format!("Compilation took {}.", pretty_duration(&elapsed_time, None)).bright_red()
-            );
-            Err(err)
-        }
-    }
-}
-
-/// This the "workhorse" command of batch compilation. It is currently run in a loop
-/// in watch mode.
+/// This the "workhorse" command of batch compilation.
```
rbalicki2, line 23: probably worth removing or moving this comment

crates/isograph_compiler/src/watch.rs
```diff
@@ -52,20 +56,102 @@ pub async fn handle_watch_command(
...
+        .collect::<Vec<_>>()
+}
+
+fn process_create_event(
+    config: &CompilerConfig,
+    create_kind: CreateKind,
+    paths: &[PathBuf],
+) -> Option<SourceFileEvent> {
+    match create_kind {
+        CreateKind::File if paths.len() == 1 => changed_file_kind(config, &paths[0])
```
rbalicki2, line 89: what happens if you change branches? presumably there are a bajillion create events at the same time... might be worth noting that as a TODO, or perhaps having some sort of logic where any event we don't handle just results in `read_and_parse_all` getting called.

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,314 @@
...
+impl SourceFiles {
+    pub fn is_empty(&self) -> bool {
+        self.schema.0.is_empty()
+            && self.schema_extensions.is_empty()
+            && self.contains_iso.is_empty()
+    }
+
+    pub fn read_and_parse_all(&mut self, config: &CompilerConfig) -> Result<(), BatchCompileError> {
+        // Schema
+        self.schema = read_and_parse_graphql_schema(&config.schema)?;
```
rbalicki2, line 38: we should probably be creating an entirely new `SourceFiles` here and calling `*self = new_source_files`.

For example, right now, we don't clear out the `schema_extensions` and `contains_iso`.

Here, I am thinking of someone else making a change in the future (e.g. adding a field to `SourceFiles`) and this code note breaking.

crates/isograph_compiler/src/watch.rs
```diff
@@ -96,3 +182,18 @@ fn create_debounced_file_watcher() -> (
...
+pub enum SourceEventKind {
+    CreateOrModify(PathBuf),
+    Rename((PathBuf, PathBuf)),
+    Remove(PathBuf),
+}
+
+pub enum ChangedFileKind {
+    Schema,
+    SchemaExtension,
+    MaybeContainsIso,
```
rbalicki2, line 196: nit: Maybe makes sense to call this `JavaScriptSourceFile`?

rbalicki2 (changes requested):
Love it!!!! From a quick test, incremental mode on the GitHub demo dropped by 50% 🤩 🤩 🤩 🤩  super happy about this.

Once you make these changes, I'll review the `process_X_event` part more closely. But overall, love what I see here, and I'm super stoked about it!

crates/isograph_compiler/src/watch.rs
ch1ffa: Yeah, fallback is a good idea

rbalicki2: This sounds good to me.

On a side note, watchman integrates with source control. But it doesn't look like that's exposed via `notify-debouncer`. Anyway, this sounds good for now and we can always revisit.

crates/isograph_compiler/src/watch.rs
```diff
@@ -52,20 +64,106 @@ pub async fn handle_watch_command(
...
+    match remove_kind {
+        RemoveKind::File | RemoveKind::Folder if paths.len() == 1 => {
+            changed_file_kind(config, &paths[0])
+                .map(|file_kind| (SourceEventKind::Remove(paths[0].clone()), file_kind))
+        }
+        _ => None,
+    }
+}
+
+fn changed_file_kind(config: &CompilerConfig, path: &PathBuf) -> Option<ChangedFileKind> {
```
rbalicki2, line 156: `fn categorize_changed_file_and_filter_changes_in_artifact_directory`

crates/isograph_compiler/src/watch.rs
```diff
@@ -52,20 +64,106 @@ pub async fn handle_watch_command(
...
+    }
+}
+
+fn process_create_event(
+    config: &CompilerConfig,
+    create_kind: CreateKind,
+    paths: &[PathBuf],
+) -> Option<SourceFileEvent> {
+    match create_kind {
+        CreateKind::File if paths.len() == 1 => changed_file_kind(config, &paths[0])
```
rbalicki2, line 101: It's unexpected for `paths.len()` to not equal 1, right? Let's panic in that case. If we get reports of that occurring, we can do something about it.

```
panic!("Unexpected file create event containing multiple files. This is indicative of a bug in Isograph.")
```

Same with the other `process_X_event`

rbalicki2, line 101: can you add a `TODO handle symlink creation`? Likewise, for symlink deletion and renaming? (Those are `CreateKind::Other` or `RemoveKind::Other`). 

Anyway, the compiler does not ignore symlinks, so we should handle them. But we can just note it for now, I don't want to worry about this, since it's obviously silly

crates/isograph_compiler/src/watch.rs
```diff
@@ -32,16 +37,23 @@ pub async fn handle_watch_command(
...
-                    ) {
-                        eprintln!("{}", "File changes detected.".cyan());
-                        let _ = compile_and_print(&config);
+                    if let Some(changes) = changed_files(&events, &state.config) {
+                        eprintln!("{}", "File changes detected. Starting to compile.".cyan());
+                        let _ = print_result(WithDuration::new(|| {
+                            if changes.len() < MAX_CHANGED_FILES {
+                                state.update(&changes)
+                            } else {
+                                state.compile()
```
rbalicki2, line 54: can we add a `eprintln` here indicating that we're falling back to a full compilation? Might be nice info for the user to see when they change branches

crates/isograph_compiler/src/watch.rs
```diff
@@ -52,20 +64,106 @@ pub async fn handle_watch_command(
     .await
 }
 
-fn any_modified_path_is_outside_artifact_directory(
+fn changed_files(
```
rbalicki2, line 67: nit: let's rename this `categorize_and_filter_events`

crates/isograph_compiler/src/watch.rs
```diff
@@ -52,20 +64,106 @@ pub async fn handle_watch_command(
...
+
+fn process_remove_event(
+    config: &CompilerConfig,
+    remove_kind: RemoveKind,
+    paths: &[PathBuf],
+) -> Option<SourceFileEvent> {
+    match remove_kind {
+        RemoveKind::File | RemoveKind::Folder if paths.len() == 1 => {
+            changed_file_kind(config, &paths[0])
+                .map(|file_kind| (SourceEventKind::Remove(paths[0].clone()), file_kind))
```
rbalicki2, line 150: Right now, if you remove a folder containing `iso` literals, it will remove that folder from `source_files`, but not remove any source files within that folder.

Perhaps we should have a `SourceEventKind::RemoveFile` and `SourceEventKind::RemoveFolder` variants, then if we process a `SourceEventKind::RemoveFolder` in `SourceFiles::handle_remove`, we iterate over the source files and remove the ones that start with that folder?

Likewise, we don't handle moves of folders correctly for the same reason

crates/isograph_compiler/src/watch.rs
```diff
@@ -52,20 +64,106 @@ pub async fn handle_watch_command(
...
+                RenameMode::Any if paths.len() == 1 => {
+                    changed_file_kind(config, &paths[0]).map(|file_kind| {
+                        if paths[0].exists() {
+                            (SourceEventKind::CreateOrModify(paths[0].clone()), file_kind)
+                        } else {
+                            (SourceEventKind::Remove(paths[0].clone()), file_kind)
+                        }
+                    })
+                }
+                RenameMode::Both if paths.len() == 2 => {
```
rbalicki2, line 127: maybe unrelated, but if we move an entire folder compilation breaks (once)

```
mkdir folder
touch folder/file
mv ./folder ./demos/pet-demo/src/components
```

```
Unable to read the file at the following path: "/Users/rbalicki/code/isograph/demos/pet-demo/src/components/folder".
Reason: Is a directory (os error 21)
```

fixes itself after modifying a file though

crates/isograph_compiler/src/compiler_state.rs
```diff
@@ -0,0 +1,134 @@
...
+            entrypoint_count: stats.entrypoint_count,
+            total_artifacts_written,
+        })
+    }
+
+    pub fn compile(&mut self) -> Result<CompilationStats, BatchCompileError> {
+        let source_files = SourceFiles::new(&self.config)?;
+        let stats = source_files.contains_iso.stats();
+        self.source_files = Some(source_files.clone());
+        let (_, artifacts) = source_files.validate_and_create_artifacts(&self.config)?;
```
rbalicki2, line 92: should we extract a common fn:
```
fn validate_and_create_artifacts_from_source_files(
  source_files: SourceFiles,
  artifact_directory,
): usize /* artifact count */
```

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,353 @@
...
+    ) -> Result<(), BatchCompileError> {
+        match change_kind {
+            ChangedFileKind::Schema => {
+                self.schema = read_and_parse_graphql_schema(&config.schema)?;
+            }
+            ChangedFileKind::SchemaExtension => {
+                let (file_path, document) = read_and_parse_schema_extensions(path)?;
+                match self.schema_extensions.entry(file_path) {
+                    Entry::Occupied(occupied_entry) => {
+                        let value = occupied_entry.into_mut();
```
rbalicki2, line 156: `mut occupied_entry` and `.get_mut`

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,353 @@
...
+                    }
+                    Entry::Vacant(vacant_entry) => {
+                        vacant_entry.insert(document);
+                    }
+                };
+            }
+            ChangedFileKind::JavaScriptSourceFile => {
+                let canonicalized_root_path = get_canonicalized_root_path(&config.project_root)?;
+                let (path_buf, file_content) =
+                    read_file(path.to_path_buf(), &canonicalized_root_path)?;
```
rbalicki2, line 167: this is probably why we are seeing the issue with the moved folders

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,353 @@
...
+        change_kind: &ChangedFileKind,
+    ) -> Result<(), BatchCompileError> {
+        match change_kind {
+            ChangedFileKind::Schema => {
+                if config.schema != *target_path {
+                    return Err(BatchCompileError::SchemaNotFound);
+                }
+            }
+            ChangedFileKind::SchemaExtension => {
+                if config.schema_extensions.contains(target_path) {
```
rbalicki2, line 202: should we handle schema extension renames? At least right now, they all have to be explicitly named in the config... I haven't thought much about this.

rbalicki2 (changes requested):
Okay, I have a bunch of nits, but TBH none of them would be blockers if we needed to ship now. This is amazing and I'm super excited about this!!!

Building a file watcher is harder than I thought 😭 thanks for all this amazing work

crates/isograph_compiler/src/compiler_state.rs
ch1ffa: Yeah, we can do it for now and change later if needed to keep schema and artifact hashes in the state.

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,353 @@
...
+                    };
+                } else {
+                    let interned_file_path = intern_file_path(source_path);
+                    self.schema_extensions.remove(&interned_file_path);
+                }
+            }
+            ChangedFileKind::JavaScriptSourceFile => {
+                let source_file_path = intern_file_path(source_path);
+                let target_file_path = intern_file_path(target_path);
+                if let Some(entry) = self.contains_iso.remove(&source_file_path) {
```
rbalicki2, line 221: The parsed iso contains a Location, which is to say, it contains the interned file path :/  This isn't ideal, we should fix that. But in the mean time, maybe it makes sense to treat renames as delete 1 + create 2

crates/isograph_compiler/src/source_files.rs
ch1ffa: Yeah, it is safe to rename schema extension if we add watch for config file. In that case an extension stops being watched and will be recreated if user reverts the name or changes the config. And we will see errors if something is used in iso literals. Although it's not consistent with batch mode where we panic if schema extension is absent

crates/isograph_compiler/src/watch.rs
```diff
@@ -1,47 +1,55 @@
...
-                        eprintln!("{}", "File changes detected.".cyan());
-                        let _ = compile_and_print(&config);
+                    if let Some(changes) = categorize_and_filter_events(&events, &state.config) {
+                        let result = if has_config_changes(&changes) {
+                            eprintln!(
+                                "{}",
+                                "Config change detected. Starting a full compilation.".cyan()
+                            );
+                            state.reload_config();
+                            reconfigure_watcher(&mut watcher, &state.config);
```
rbalicki2, line 40: it might be cleaner to not have a `[re]configure_watcher` functions, and instead just do

```
let (new_rx, new_watcher) = create_debounced_file_watcher(&config);
*rx = new_rx;
*watcher = new_watcher;
*state = create_entirely_new_state(); // just reread everything
```

what do you think? IMO this is easier to reason about, instead of methods that take `&mut self`

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,370 @@
...
+
+    pub fn update(
+        &mut self,
+        config: &CompilerConfig,
+        changes: &[SourceFileEvent],
+    ) -> Result<(), BatchCompileError> {
+        let errors = changes
+            .iter()
+            .filter_map(|(event, change_kind)| match change_kind {
+                ChangedFileKind::Config => None,
```
rbalicki2, line 84: should we panic here? Encountering a config change is unexpected.

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,370 @@
...
+        Ok(())
+    }
+
+    fn handle_update_source_folder(
+        &mut self,
+        config: &CompilerConfig,
+        event_kind: &SourceEventKind,
+    ) -> Result<(), BatchCompileError> {
+        match event_kind {
+            SourceEventKind::CreateOrModify(_) => {}
```
rbalicki2, line 177: this isn't quite right. We should read all the files in the folder. If someone moves a folder from outside of the component directory into the component directory, and it contains iso literals, then these would be unnoticed until we restart the compiler

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,370 @@
...
+            SourceEventKind::Rename((source_path, target_path)) => {
+                let path_string = source_path.to_string_lossy().to_string();
+                self.contains_iso
+                    .retain(|file_path, _| !file_path.to_string().starts_with(&path_string));
```
rbalicki2, line 181: TIL about retain

crates/isograph_compiler/src/source_files.rs
```diff
@@ -0,0 +1,370 @@
...
+        }
+        Ok(())
+    }
+
+    fn create_or_update_schema_extension(
+        &mut self,
+        path: &PathBuf,
+    ) -> Result<(), BatchCompileError> {
+        let (file_path, document) = read_and_parse_schema_extensions(path)?;
+        match self.schema_extensions.entry(file_path) {
```
rbalicki2, line 202: since we're not doing anything different in the vacant and occupied cases, this would be cleaner as `self.schema_extensions.insert`. Likewise below for iso literals

rbalicki2 (changes requested):
This looks amazing!!! Just a few stylistic nits and one issue with folders, but otherwise, we should be good to land soon :)

Thanks for handling config changes too :)

crates/isograph_compiler/src/source_files.rs
ch1ffa: yeah, for now it means a bug. I thought eventually we can handle some of the changes without a full compilation but we should panic atm

crates/isograph_compiler/src/source_files.rs
ch1ffa: Indeed, after moving to match file kind first everything became simpler

crates/isograph_compiler/src/watch.rs
ch1ffa: Agreed. I tried to do this first, but had some issues. But yeah, now I see how to implement it, thanks.

crates/isograph_compiler/src/watch.rs
```diff
@@ -1,47 +1,55 @@
...
-                        eprintln!("{}", "File changes detected.".cyan());
-                        let _ = compile_and_print(&config);
+                    if let Some(changes) = categorize_and_filter_events(&events, &state.config) {
+                        let result = if has_config_changes(&changes) {
+                            eprintln!(
+                                "{}",
+                                "Config change detected. Starting a full compilation.".cyan()
+                            );
+                            state = CompilerState::new(state.config.config_location);
+                            watcher.stop();
```
rbalicki2, line 39: just FYI dropping the watcher will also stop it

rbalicki2 (approved):
Lets goooo shipitttt

Amazing work!

PR #226, opened 2024-10-27
fix github-demo project names

ch1ffa:
github-demo compilation was broken in watch mode. pnpm filters projects by a project name from `package.json` rather than a folder name


rbalicki2 (approved):
great catch! shipittt

PR #229, opened 2024-10-28
read client field variant in `validate_field_type_exists_and_is_linked`

PatrykWalach:
Part of #53

### DIFF
--- a/crates/isograph_schema/src/validate_schema.rs
+++ b/crates/isograph_schema/src/validate_schema.rs
@@ -453,12 +453,13 @@ fn validate_and_transform_client_fields(
         },
     ))?;
 
-    get_all_errors_or_all_ok_iter(client_fields.into_iter().map(|client_field| {
+    get_all_errors_or_all_ok_iter(client_fields.iter().map(|client_field| {
         validate_client_field_selection_set(
             schema_data,
             client_field,
             server_fields,
             &client_field_args,
+            &client_fields,
         )
         .map_err(|err| err.into_iter())
     }))
@@ -497,9 +498,10 @@ fn validate_all_variables_are_used(
 
 fn validate_client_field_selection_set(
     schema_data: &ServerFieldData,
-    top_level_client_field: UnvalidatedClientField,
+    top_level_client_field: &UnvalidatedClientField,
     server_fields: &[ValidatedSchemaServerField],
     client_field_args: &ClientFieldArgsMap,
+    client_fields: &[UnvalidatedClientField],
 ) -> Result<ValidatedClientField, Vec<WithLocation<ValidateSchemaError>>> {
     let variable_definitions = client_field_args
         .get(&top_level_client_field.id)
@@ -512,6 +514,7 @@ fn validate_client_field_selection_set(
     let client_field_parent_object = schema_data.object(top_level_client_field.parent_object_id);
     let selection_set_result = top_level_client_field
         .reader_selection_set
+        .as_ref()
         .map(|selection_set| {
             validate_client_field_definition_selections_exist_and_types_match(
                 schema_data,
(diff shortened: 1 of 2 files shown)


crates/isograph_schema/src/validate_schema.rs
```diff
@@ -1055,13 +1071,20 @@ fn validate_field_type_exists_and_is_linked(
...
-                ValidateSelectionsError::FieldSelectedAsLinkedButTypeIsClientField {
-                    field_parent_type_name: field_parent_object.name,
-                    field_name: linked_field_name,
-                },
-                linked_field_selection.name.location,
-            )),
+            FieldType::ClientField(client_field_id) => {
+                let client_field = &client_fields[client_field_id.as_usize()];
+
+                match client_field.variant {
```
PatrykWalach, line 1077: 1. Here given that a selected field is a client pointer we need to validate it's selections

to do it, we have to thread `client_fields` and `client_pointers` through all the functions

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -769,20 +780,21 @@ fn validate_client_field_definition_selections_exist_and_types_match(
     let validated_selection_set_result =
         get_all_errors_or_all_ok(selection_set.into_iter().map(|selection| {
             validate_client_field_definition_selection_exists_and_type_matches(
-                selection,
+                selection.clone(),
```
PatrykWalach, line 783: This seems bad

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -601,7 +609,9 @@ fn validate_use_refetch_field_strategy(
     Ok(ValidatedRefetchFieldStrategy {
         refetch_selection_set,
         root_fetchable_type: use_refetch_field_strategy.root_fetchable_type,
-        generate_refetch_query: use_refetch_field_strategy.generate_refetch_query,
+        generate_refetch_query: use_refetch_field_strategy
+            .generate_refetch_query
+            .clone_box(),
```
PatrykWalach, line 614: I don't know if this can be done better?

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -769,20 +780,21 @@ fn validate_client_field_definition_selections_exist_and_types_match(
...
                 &variable_definitions,
+                client_fields,
             )
         }));
 
     let (validated_selection_set, _) = get_all_errors_or_tuple_ok(
         validated_selection_set_result,
         validate_all_variables_are_used(
-            variable_definitions,
+            variable_definitions.to_vec(),
```
PatrykWalach, line 797: This seems bad too

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -453,12 +453,13 @@ fn validate_and_transform_client_fields(
         },
     ))?;
 
-    get_all_errors_or_all_ok_iter(client_fields.into_iter().map(|client_field| {
+    get_all_errors_or_all_ok_iter(client_fields.iter().map(|client_field| {
```
rbalicki2, line 456: - Sorry about shipping a bunch of stuff that cleaned up `validate_schema` 😭  while you were working on it
- It looks like the reason we are passing references here is so that we can also pass `&ClientFields` down, in order to look up the nested client field's `variant`. I'm not really following. Why?
- I would imagine that what we need is to go from a client field + linked field selection to "is this a pointer or not", right? In which case, in `validate_field_type_exists_and_is_linked`, the `FieldType::ClientField(_) =>` match arm should have an enum which is `ClientFieldId` or `ClientPointerId`.
    - Let's make this enum generic and call it `FieldOrPointer<TField, TPointer>`, since we can use this enum to distinguish between scalar and linked server fields eventually.

Am I understanding correctly?

PR #233, opened 2024-10-30
generate client pointer entrypoint

PatrykWalach:


crates/graphql_artifact_generation/src/reader_ast.rs
```diff
@@ -117,15 +122,63 @@ fn linked_field_ast_node(
...
+
+            // N.B. additional_reader_imports will be empty for now, but at some point, we may have
+            // refetch selection sets that import other things! Who knows!
+            for import in additional_reader_imports {
+                reader_imports.insert(import);
+            }
+
+            return format!(
+                "{indent_1}{{\n\
+        {indent_2}kind: \"LoadablySelectedField\",\n\
```
PatrykWalach, line 161: 1. This replaces `Linked` fields with `LoadablySelectedField`

crates/isograph_schema/src/create_merged_selection_set.rs
```diff
@@ -29,10 +29,10 @@ pub type MergedSelectionMap = BTreeMap<NormalizationKey, MergedServerSelection>;
 
 // Maybe this should be FNVHashMap? We don't really need stable iteration order
 pub type ClientFieldToCompletedMergeTraversalStateMap =
-    BTreeMap<ClientFieldId, ClientFieldTraversalResult>;
+    BTreeMap<FieldType<ServerFieldId, ClientFieldId>, FieldTraversalResult>;
```
PatrykWalach, line 32: 1. Now we need to keep track of encountered `ServerFields`

crates/isograph_schema/src/create_merged_selection_set.rs
```diff
@@ -775,6 +781,27 @@ fn merge_validated_selections_into_selection_map(
                                             global_client_field_map,
                                             variable_context,
                                         );
+                                        let server_field =
+                                            &schema.server_fields[server_field_id.as_usize()];
+
+                                        create_merged_selection_map_for_field_and_insert_into_global_map(
+                                            schema,
```
PatrykWalach, line 788: 2. Then we insert encountered fields

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -90,123 +91,93 @@ pub fn get_artifact_path_and_content(
...
-                    encountered_client_field,
-                    project_root,
-                    artifact_directory,
-                    *info,
-                    &traversal_state.refetch_paths,
-                ));
-
-                if *was_ever_selected_loadably {
-                    path_and_contents.push(generate_refetch_reader_artifact(
-                        schema,
```
PatrykWalach, line 117: 3. All of this gets extracted into `generate_entrypoint_artifacts_with_field_traversal_result`

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -90,123 +91,93 @@ pub fn get_artifact_path_and_content(
...
+
+                // path_and_contents.push(generate_refetch_reader_artifact(
+                //     schema,
+                //     encountered_server_field,
+                //     None,
+                //     &traversal_state.refetch_paths,
+                //     true,
+                // ));
+
+                generate_entrypoint_artifacts_with_field_traversal_result(
```
PatrykWalach, line 115: 4. And then we can generate the entrypoints fo `ServerField`

PatrykWalach:
I think this doesn't fully make sense, we can't have a single `entrypoint.ts` we should have multiple `refetch` queries, but `LoadablySelectedField` requires an `entrypoint`.

demos/github-demo/src/isograph-components/PullRequestLink.tsx
```diff
@@ -11,6 +12,10 @@ export const PullRequestLink = iso(`
       name
       owner {
         login
+        asUser {
```
rbalicki2, line 15: Amazing!!! Thanks for continuing to make progress.

It took me a while to figure out how we should handle this. In particular, I want to make sure what we're building fits into a coherent bigger picture.

## `asUser` vs other types of fields

How does `asUser` differ from `pet(id: 4)`? Both of them can be thought of as client pointers. A client pointer has a `condition` function associated with it. `pet(id: '4')` has a simple condition: `(data: {}) => { __typename: 'Pet', id: '4' }`, and `asUser` has condition `(data: { __typename, id }) => data.__typename === 'User' ? { __typename: 'User', id: data.id } : null`. We aren't actually generating the `pet` function, but that's the implicit behavior.

So, the way I'm thinking about it is that linked fields and pointers are the same thing, except linked fields have a "default" condition.

Okay, with that said, how does `asUser` differ from a proper client pointer. Well, it differs in that it doesn't have a refetch query or entrypoint. The reader AST is sufficient. And since it doesn't have a refetch query or entrypoint, we can just read the result as part of the parent fragment.

## categorizing linked fields

So, for linked fields on the reader side, I see four possible cases:

```
                     no query/entrypoint     has query/entrypoint
unconditional (1)    regular linked field    loadably selected linked field
  conditional        asUser field (2)        client pointer, loadably selected inline fragment
```
- Note that this grid excludes loadably selected client fields and refetch fields, which are scalars!
- (1): the unconditional row just means "default" behavior. You can't have an unconditional client pointer, because we're defining a new field, and so there isn't a link in the store (at that same slot) to use for the default behavior. So, a client pointer which always returns a DataID is still unconditional, in this world.
- (2): also, client pointers that are executed on the server. That's pretty much what `asUser` fields are, with the added twist that GraphQL understands them.

So, I think we can model reader linked fields in all four quadrants as follows:

```diff
export type ReaderLinkedField = {
  readonly kind: 'Linked';
  readonly fieldName: string;
  readonly alias: string | null;
  readonly selections: ReaderAst<unknown>;
  readonly arguments: Arguments | null;
+  readonly condition: LinkedFieldCondition | null // (1)
+  // let's not implement this yet:
+  readonly entrypointOrRefetchQueryIndexOrSomething: Something | null
};
```

Okay, so what is this `LinkedFieldCondition`? I think that this is a generated file, `Node/asUser/resolver_reader.ts`, which we generate for all encountered client pointers and inline fragments. For a client pointer, the condition would be an import of the resolver (i.e. `export const bestFriendClientPointer = ...`). For inline fragment fields, we can hard code the resolver.

An eager reader artifact works as a `LinkedFieldCondition`, and maybe we can reuse that, with the hard coded resolver, and with `TClientFieldValue` (aka the output type) being `Link`.

So, given that, we should actually just have

```diff
+  readonly condition: EagerReaderArtifact<any, Link>
```

(Maybe we can improve upon the `any` type, but I don't think it's important.)

------------

Okay, so that's how the reader AST stuff will have to change. How can we actually do this?

- the schema server field variant `InlineFragment` should have a reader AST containing `__typename` and `id`
    - that should be enough. I think we can just hard code the `({ data }) => data.__typename === ConcreteType ? data.id : null` wherever we generate that for now.
- when generating a merged selection set, that reader AST w/typename and id should additionally be merged into the merged selection set.
- when generating artifacts
    - when we encounter a linked field with an inline fragment, that should be tracked in a data structure similar to `encountered_client_field_map`
    - for each linked field in this data structure, we should generate a resolver reader. That is probably enough. But, the resolver reader references the `param_type.ts` and `output_type.ts`, so maybe it will be easier to just generate those as well. (The output_type.ts can just be `export type Type__fieldName__output_type = Link`, `param_type.ts` will probably be correctly generated, but also doesn't matter.)
    - whenever we encounter an inline fragment, we need to import the generated file, i.e. `PullRequestLink/resolver_resolver` should import `User/asUser/resolver_reader`, just as if it was a client field
- read.ts, when we encounter a linked field, should either read it from the store using `storeRecord[getParentRecordKey(...)]` or by reading `condition.readerAst` and calling `condition.resolver` with that data


Okay, I know that's a lot! We can chat tomorrow if you want about this.

PatrykWalach:
There's a case in the github schema, where `interface Actor` doesn't have `id` but `User implements Actor` does
We try to read from the store `id` and `__typename`, but we end up with `MissingData`, because `id` is not present.
We should just read `null` in this case.
This need to be fixed.

rbalicki2:
Good catch re: actor. Let's select `... on User { id }`. That's kind of annoying, because we're going to have `... on User { id } ... on AnotherConcreteType { id }` etc but I think it's good enough and will unblock us.

Alternatively, perhaps conditions can return `boolean | Link | null`, and thus we won't need to select the `id`.

rbalicki2:
Or perhaps we should add an `__id` field everywhere, which returns the store id...

PatrykWalach:
The issue is, there's no id in the store
> Alternatively, perhaps conditions can return boolean | Link | null, and thus we won't need to select the id.
This could work

Plus the fact, that we require the `id` to use `interfaces` seems silly

rbalicki2:
I'm down for any of these, we can always switch out the implantation details

demos/github-demo/src/isograph-components/UserLink.tsx
```diff
@@ -19,16 +23,27 @@ export const UserLink = iso(`
...
+        style={{ cursor: 'pointer' }}
+      >
+        {children}
+      </Link>
+      {data.asUser?.twitterUsername && (
+        <Link
+          href={`https://x.com/${data.asUser?.twitterUsername}`}
+          target="_blank"
+        >
+          {data.asUser?.twitterUsername && <XIcon> </XIcon>}
```
rbalicki2, line 44: Looks like self-closing tags work just fine here as well

demos/github-demo/src/isograph-components/UserLink.tsx
```diff
@@ -19,16 +23,27 @@ export const UserLink = iso(`
...
+          data.asUser &&
+          setRoute({
+            kind: 'User',
+            userLogin: data.asUser.login,
+          })
+        }
+        style={{ cursor: 'pointer' }}
+      >
+        {children}
+      </Link>
```
rbalicki2, line 38: can we add `&nbsp;` between the username and twitter

crates/isograph_schema/src/add_pointers_to_supertypes.rs
```diff
@@ -17,48 +22,85 @@ impl UnvalidatedSchema {
...
-                    .insert(field_name, FieldType::ServerField(next_server_field_id))
-                    .is_some()
-                {
-                    return Err(WithLocation::new(
-                        ProcessTypeDefinitionError::FieldExistsOnType {
-                            field_name,
-                            parent_type: supertype.name,
+                let server_field = SchemaServerField {
+                    description: Some(
+                        format!("A client poiter for the {} type.", subtype.name)
```
rbalicki2, line 67: typo: `poiter`

demos/github-demo/src/isograph-components/__isograph/Query/RepositoryPage/entrypoint.ts
```diff
@@ -25,6 +25,12 @@ const queryText = 'query RepositoryPage ($repositoryName: String!, $repositoryOw
           author {\
             __typename,\
```
rbalicki2, line 26: kind of cool that this has the correct behavior (selecting `id` only if `author` is a `User`)

libs/isograph-react/src/core/read.ts
```diff
@@ -209,6 +209,43 @@ function readData<TReadFromStore>(
...
+              reason:
+                'Missing data for ' +
+                storeRecordName +
+                ' on root ' +
+                root.__link,
+              nestedReason: data,
+              recordLink: data.recordLink,
+            };
+          }
+          const condition = field.condition.resolver({
```
rbalicki2, line 236: just amazing that we have the correct typing for condition here 🙌

rbalicki2 (changes requested):
buncha small changes, overall, this looks like it can be shipped without much additional work.

Amazing job!!! I'm super pumped that this will land, and with so much less work than I expected it to require 🤩 🙌

demos/github-demo/src/isograph-components/UserLink.tsx
```diff
@@ -1,12 +1,16 @@
...
 import { Link } from '@mui/material';
 import { ReactNode } from 'react';
 import { Route } from './GithubDemo';
 
 export const UserLink = iso(`
   field Actor.UserLink @component {
-    login
+    asUser {
+      id
+      login
```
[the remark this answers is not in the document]
PatrykWalach: If the `Actor` is a `Bot` the app will break

rbalicki2:
Re: app breaking if the user is a bot. That's a pre existing issue with the app. I'll fix that on main

crates/graphql_artifact_generation/src/reader_ast.rs
```diff
@@ -117,12 +122,35 @@ fn linked_field_ast_node(
...
     let indent_1 = "  ".repeat(indentation_level as usize);
     let indent_2 = "  ".repeat((indentation_level + 1) as usize);
+
+    let condition = match linked_field.associated_data.variant {
+        SchemaServerFieldVariant::InlineFragment(_) => {
+            let object = schema
+                .server_field_data
+                .object(linked_field.associated_data.parent_object_id);
+
+            let type_and_field: ObjectTypeAndFieldName = ObjectTypeAndFieldName {
```
rbalicki2, line 132: nit: no need to annotate the type here.

crates/isograph_schema/src/add_pointers_to_supertypes.rs
```diff
@@ -1,10 +1,15 @@
+use std::vec;
```
rbalicki2, line 1: nit: this is already part of the prelude, i.e. doesn't need to be explicitly `use`d

rbalicki2 (approved):
LFGGGG shipittttt

PR #237, opened 2024-11-01
add config for file extensions

### LINKED ISSUE #236: Option to generate .ts extensions for imports in generated files

PatrykWalach:
Closes #236


crates/isograph_config/src/compilation_options.rs
```diff
@@ -178,9 +234,23 @@ impl Default for ConfigFileOptionalValidationLevel {
     }
 }
 
+#[derive(Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+enum ConfigFileOptionalGenerateFileExtensions {
+    Yes(HashMap<String, String>),
```
PatrykWalach, line 240: I've considered two use cases,
- keeping extensions ".tsx" for ".tsx" file
- mapping to javascript ".jsx" for ".tsx" file

I've made it a `HashMap`, so it's possible to use more file extensions, like `.mjs`, `.cjs`, or not standard ones like `.es6` or whatever.

crates/isograph_config/src/compilation_options.rs
```diff
@@ -23,9 +23,64 @@ pub struct CompilerConfig {
...
+            None => "",
+            Some(v) => v,
+        }
+    }
+}
+
+impl Default for OptionalGenerateFileExtensions {
+    fn default() -> Self {
+        let mut map = HashMap::new();
+        map.insert("js".to_string(), "".to_string());
```
PatrykWalach, line 68: Instead of hardcoding this I should replace `OptionalGenerateFileExtensions` with interface, and two implementation 
- for `yes` current implementation - hashmap with warnings
- for `no` method that always returns empty string `""`

crates/isograph_config/src/compilation_options.rs
rbalicki2: When would you need to generate anything besides the name of the actual file?

In addition, if you remap `ts` to `js`, should we also modify `write_artifacts` to generate files with different suffixes?

Also, is the best practice to include extensions, and if so, should we default to including extensions?

crates/isograph_config/src/compilation_options.rs
```diff
@@ -23,9 +23,47 @@ pub struct CompilerConfig {
...
+    pub fn get(&self, extension: &str) -> &str {
+        match self {
+            OptionalGenerateFileExtensions::No => "",
+            OptionalGenerateFileExtensions::Yes(extensions) => match extensions.get(extension) {
+                Some(mapped_extension) => mapped_extension,
+                None => {
+                    eprintln!(
+                        "{}\nmissing .{} file extension in options.generate_file_extensions\n",
+                        "Warning:".yellow(),
+                        extension
```
rbalicki2, line 48: we probably want to panic here, right?

crates/isograph_config/src/compilation_options.rs
PatrykWalach: I'm not sure what best practise here is.

It all depends on tsconfig, and what output user wants, native esm module require explicit ".js" file extensions, I don't know if commonjs allows any extension.

- in case application is transpiled with typescript then it probably runs with `moduleResolution` of `node16` or `nodeNext` where file extensions are required (not sure which .js or .ts are allowed)
- for bundler like vite with `"moduleResolution": "bundler"` imports do not require file extension
- deno has builtin typescript support so it expects `.ts` extensions, and there's an option in typescript for custom file extensions, like `.es6` etc. 



For me most common config is `"moduleResolution": "bundler"` with vite which works without any file extensions. 



I think, theoreticly there should be a different option to customize generated artifacts extension to ".mts" so they can be imported in commonjs modules, not sure if anyone would need this

PatrykWalach: I think there are cases when for `.ts` files `.ts` extension is not allowed, only `.js`

PatrykWalach: Need to check all this to make sure
| File | typescript ESM | typescript CJS | bundler | bundler + `allowImportingTsExtensions` | deno |
|------------------|------------------|------------------|-----------|-----|--|
| .tsx   |  ? |  ?  | `.js` `.jsx` `none` |  `.js` `.jsx` `.tsx` `none`   |?|
| .ts   | ?  |  ?  | `.js` `none` |  `.js` `.ts` `none`   |?|

<details>
  <summary>bundler</summary>

   ```tsx
  {
    "module": "ESNext", 
    "moduleResolution": "Bundler",
  }		
  ```
</details>

PatrykWalach: Different extensions allow for coexistence of CommonJS and ESM files (`.mjs` `.cjs`), explicit file extensions allow for easier transitions between module systems

crates/graphql_artifact_generation/src/eager_reader_artifact.rs
```diff
@@ -303,10 +317,8 @@ fn generate_function_import_statement(
         // Anyway, TODO do better.
         pathdiff::diff_paths(path_to_client_field, artifact_directory.join("Type/Field"))
             .expect("Relative path should work");
-    let extension_char_count_including_dot =
-        relative_path.extension().map(|x| x.len() + 1).unwrap_or(0);
     let complete_file_name = relative_path.to_str().expect(
-        "This path should be stringifiable. This probably is indicative of a bug in Relay.",
+        "This path should be stringifiable. This probably is indicative of a bug in Isograph.",
```
rbalicki2, line 321: 😭  the truth comes out

crates/isograph_config/src/compilation_options.rs
```diff
@@ -28,6 +28,23 @@ pub struct CompilerConfig {
 #[derive(Default, Debug, Clone, Copy)]
 pub struct ConfigOptions {
     pub on_invalid_id_type: OptionalValidationLevel,
+    pub generate_file_extensions: OptionalGenerateFileExtensions,
+}
+
+#[derive(Default, Debug, Clone, Copy)]
+pub enum OptionalGenerateFileExtensions {
+    Yes,
```
rbalicki2, line 36: Can we make this a boolean please? Otherwise lgtm!! Thank you for tackling this

crates/isograph_config/src/compilation_options.rs
```diff
@@ -28,6 +28,23 @@ pub struct CompilerConfig {
 #[derive(Default, Debug, Clone, Copy)]
 pub struct ConfigOptions {
     pub on_invalid_id_type: OptionalValidationLevel,
+    pub generate_file_extensions: OptionalGenerateFileExtensions,
```
rbalicki2, line 31: nit: `include_file_extensions_in_import_statements` is probably clearer

crates/isograph_config/src/compilation_options.rs
PatrykWalach: `boolean` in `isograph.config.json` and map it to `enum` in code?

rbalicki2 (approved):
Looks great! Shipittt

PR #242, opened 2024-11-03
Enable project references

PatrykWalach:
I've fixed `pnpm -r compile`, the cache was not created in `dist`, but in `../dist` so I've used `--force` to treat it as stale
Also there are changes to CI that should cache between `jobs` across a single `commit`, as there's no possibility of src files being changed.

I think it's still not ideal running `pnpm -r compile` is probably slower than before,
I need to experiment with doing something like `pnpm -r clean && tsc --build --force`


rbalicki2:
https://github.com/isographlabs/isograph/actions/runs/11730017827/job/32676927319 looks like we run some build issues

PatrykWalach:
I was looking more into this and I think this might not be what we want, these project references are meant to be used instead of pnpm workspace dependencies.

But I also couldn't find a reasonable alternative for watch mode across whole monorepo.

ch1ffa:
We used turborepo for one of our monorepos. It worked quite well with pnpm workspace and watch mode.

PatrykWalach:
I think turbo repo is a better option for caching. It's just that running multiple watch commands makes them `override` each others output, but it's not a big deal.

PR #246, opened 2024-11-04
add tracing

### LINKED ISSUE #239: logLevel parameter to cli
Show timing info of the various substeps

ch1ffa:
Closes #239

This PR adds `tracing` to the compiler and exposes its `LevelFilter` enum as `--log-level` parameter with possible values: `trace`, `debug`, `info`, `warn`, `error`, `off`.

It can be later configured to measure spans duration or export data in various formats if needed.

Now we can use `debug!` or `trace!` macros to provide additional diagnostics.

### DIFF
--- a/crates/isograph_cli/src/main.rs
+++ b/crates/isograph_cli/src/main.rs
@@ -5,6 +5,8 @@ use colored::Colorize;
 use isograph_compiler::{compile_and_print, handle_watch_command};
 use isograph_config::create_config;
 use opt::{Command, CompileCommand, LspCommand, Opt};
+use std::io;
+use tracing::{error, info, level_filters::LevelFilter};
 
 #[tokio::main]
 async fn main() {
@@ -21,6 +23,7 @@ async fn main() {
 }
 
 async fn start_compiler(compile_command: CompileCommand) {
+    configure_logger(compile_command.log_level);
     let config_location = compile_command
         .config
         .unwrap_or("./isograph.config.json".into());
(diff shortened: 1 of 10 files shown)


crates/isograph_config/src/compilation_options.rs
```diff
@@ -49,9 +48,7 @@ impl OptionalValidationLevel {
             OptionalValidationLevel::Ignore => Ok(()),
             OptionalValidationLevel::Warn => {
                 let warning = on_error();
-                // TODO pass to some sort of warning gatherer, this is weird!
```
rbalicki2, line 52: nice

crates/isograph_cli/src/main.rs
```diff
@@ -60,13 +55,27 @@ async fn start_language_server(lsp_command: LspCommand) {
...
+        error!(
             "{}",
             "Error encountered when running language server.".bright_red(),
             // TODO derive Error and print e
         );
         std::process::exit(1);
     }
 }
+
+// TODO: Configure parameters based on provided log_level.
```
rbalicki2, line 69: what is this TODO referring to? This looks like it is using log_level

rbalicki2 (approved):
LGTM, just lemme know if that TODO is outdated and we'll land

crates/isograph_cli/src/main.rs
ch1ffa: That is for the future me :) I want to provide different tracing parameters, like function name, line number, etc. for debug and trace levels. But I can add it in this PR as well.

rbalicki2: sgtm

PR #250, opened 2024-11-05
add `options.on_missing_babel_transform`

### LINKED ISSUE #235: Option to not throw in iso function for those not using babel plugin

PatrykWalach:
Closes #235


rbalicki2 (approved):
LGTM, let's land it. Thanks for tackling this so fast!

PR #253, opened 2024-11-06
make `client_fields` an enum

PatrykWalach:


crates/graphql_artifact_generation/src/iso_overload_file.rs
```diff
@@ -260,10 +261,12 @@ fn user_written_fields(
     schema
         .client_fields
         .iter()
-        .filter_map(|client_field| match client_field.variant {
-            ClientFieldVariant::UserWritten(info) => {
-                Some((client_field, info.user_written_component_variant))
-            }
-            ClientFieldVariant::ImperativelyLoadedField(_) => None,
+        .filter_map(|client_field| match client_field {
```
rbalicki2, line 264: lgtm, this is used for generating the iso.ts file, and this makes sense there.

crates/graphql_artifact_generation/src/reader_ast.rs
```diff
@@ -37,7 +37,7 @@ fn generate_reader_ast_node(
                         indentation_level,
                         initial_variable_context,
                     ),
-                    FieldType::ClientField(client_field_id) => {
+                    FieldType::ClientField(ClientType::ClientField(client_field_id)) => {
```
rbalicki2, line 40: (1) hm... we'd have to panic here, since a scalar field cannot be a client pointer. I think in `validate_schema`, it doesn't make sense for scalar field selections to be `ClientType`. instead, it makes sense for linked fields to have that enum.

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -52,7 +52,7 @@ pub type ValidatedRefetchFieldStrategy = UseRefetchFieldRefetchStrategy<
 >;
 
 /// The validated defined field that shows up in the TScalarField generic.
-pub type ValidatedFieldDefinitionLocation = FieldType<ServerFieldId, ClientFieldId>;
+pub type ValidatedFieldDefinitionLocation = FieldType<ServerFieldId, ClientType<ClientFieldId>>;
```
rbalicki2, line 55: (2) let's make this part of the `ClientFieldSelectionLinkedFieldAssociatedData` not part of `ClientFieldSelectionScalarFieldAssociatedData`

crates/isograph_compiler/src/refetch_fields.rs
```diff
@@ -43,9 +43,11 @@ fn add_refetch_field_to_object(
         Entry::Vacant(vacant_entry) => {
             let next_client_field_id = client_fields.len().into();
 
-            vacant_entry.insert(FieldType::ClientField(next_client_field_id));
+            vacant_entry.insert(FieldType::ClientField(ClientType::ClientField(
+                next_client_field_id,
+            )));
 
-            client_fields.push(ClientField {
+            client_fields.push(ClientType::ClientField(ClientField {
```
rbalicki2, line 50: Not in this PR, but long term, perhaps we can get rid of variant by making refetch (etc) fields `FieldType::Server`? Or something. Because `FieldType::Client` implies that there's a selection set, it's user generated, etc

crates/isograph_schema/src/isograph_schema.rs
```diff
@@ -191,7 +198,9 @@ impl<TSchemaValidationState: SchemaValidationState> Schema<TSchemaValidationStat
         TSchemaValidationState::ClientFieldSelectionLinkedFieldAssociatedData,
         TSchemaValidationState::VariableDefinitionInnerType,
     > {
-        &self.client_fields[client_field_id.as_usize()]
+        match &self.client_fields[client_field_id.as_usize()] {
+            ClientType::ClientField(client_field) => client_field,
+        }
```
rbalicki2, line 203: This gives me a bit of pause, but I suppose it's fine to panic here if the invariant is violated (i.e. the client field ID refers to a pointer)

crates/isograph_schema/src/isograph_schema.rs
```diff
@@ -363,7 +372,8 @@ pub struct SchemaObject {
     /// TODO remove id_field from fields, and change the type of Option<ServerFieldId>
     /// to something else.
     pub id_field: Option<ServerStrongIdFieldId>,
-    pub encountered_fields: BTreeMap<SelectableFieldName, FieldType<ServerFieldId, ClientFieldId>>,
+    pub encountered_fields:
+        BTreeMap<SelectableFieldName, FieldType<ServerFieldId, ClientType<ClientFieldId>>>,
```
rbalicki2, line 376: seems fine here

crates/isograph_schema/src/validate_client_field.rs
```diff
@@ -14,53 +14,63 @@ use lazy_static::lazy_static;
...
+    UnvalidatedVariableDefinition, ValidateSchemaError, ValidateSchemaResult, ValidatedClientField,
+    ValidatedIsographSelectionVariant, ValidatedLinkedFieldAssociatedData,
+    ValidatedLinkedFieldSelection, ValidatedRefetchFieldStrategy,
+    ValidatedScalarFieldAssociatedData, ValidatedScalarFieldSelection, ValidatedSchemaServerField,
+    ValidatedSelection, ValidatedVariableDefinition,
 };
 
 type UsedVariables = BTreeSet<VariableName>;
-type ClientFieldArgsMap = HashMap<ClientFieldId, Vec<WithSpan<ValidatedVariableDefinition>>>;
+type ClientFieldArgsMap =
```
rbalicki2, line 28: lgtm, makes sense

crates/isograph_schema/src/unvalidated_schema.rs
```diff
@@ -36,7 +36,7 @@ pub type UnvalidatedSchema = Schema<UnvalidatedSchemaState>;
 /// On unvalidated schema objects, the encountered types are either a type annotation
 /// for server fields with an unvalidated inner type, or a ScalarFieldName (the name of the
 /// client field.)
-pub type UnvalidatedObjectFieldInfo = FieldType<ServerFieldId, ClientFieldId>;
+pub type UnvalidatedObjectFieldInfo = FieldType<ServerFieldId, ClientType<ClientFieldId>>;
```
rbalicki2, line 39: lgtm

rbalicki2 (changes requested):
let's not making `ClientType` part of the scalar field selection. Instead, we should have an enum on the linked field side that distinguishes between server defined linked fields and client pointers. Can you make a one-variant enum there and I'll re-review?

crates/isograph_schema/src/isograph_schema.rs
```diff
@@ -150,13 +150,15 @@ pub enum FieldType<TServer, TClient> {
...
-    ClientField(TField),
+pub enum ClientType<TClient> {
+    ClientField(TClient),
+    // ClientPointer(TPointer)
 }
 
 #[derive(Debug, Clone, Copy, Ord, PartialOrd, PartialEq, Eq)]
-pub enum LinkedType<TServerObject> {
-    ServerObject(TServerObject),
+pub enum ObjectType<TServer> {
```
PatrykWalach, line 159:
```rs
pub enum LinkedType<TServerObject> {
    LinkedServerObject(TServerObject),
    // LinkedPointer(TPointer)
}
```
is this a better name?

rbalicki2: yes, this is good and explicit

crates/graphql_artifact_generation/src/reader_ast.rs
```diff
@@ -125,21 +125,23 @@ fn linked_field_ast_node(
 
     let condition = match linked_field.associated_data.variant {
         SchemaServerFieldVariant::InlineFragment(_) => {
-            let object = schema
-                .server_field_data
-                .object(linked_field.associated_data.parent_object_id);
+            match linked_field.associated_data.parent_object_id {
+                LinkedType::LinkedServerObject(parent_object_id) => {
```
rbalicki2, line 129: Hm. We shouldn't have an inline variant and a linked client field. Happy to panic when necessary, for now, but if convenient, maybe we put the variant inside of the `TServerObject`? e.g.
```
pub struct ValidatedLinkedFieldAssociatedData {
    pub parent_object_id: LinkedType<StructWithServerObjectIdAndVariant>,
```

thoughts?

note that in reader_ast, we check the variant inside of the `LinkedServerObject` case, so that would work there, too

rbalicki2 (changes requested):

crates/graphql_artifact_generation/src/reader_ast.rs
PatrykWalach: Since `SchemaServerFieldInlineFragmentVariant` has `pub server_field_id: ServerFieldId` already
I've removed `LinkedType` variant and added `field_id` to `SchemaServerFieldLinkedFieldVariant`
```rust
#[derive(Debug, Clone)]
pub enum SchemaServerFieldVariant {
    InlineFragment(SchemaServerFieldInlineFragmentVariant),
    LinkedField(SchemaServerFieldLinkedFieldVariant),
}

#[derive(Debug, Clone)]
pub struct SchemaServerFieldLinkedFieldVariant {
    pub field_id: FieldType<ServerFieldId, ()>,
}
```
This way in both cases we have access to the `field` which has access to `parent_object_id`

rbalicki2 (approved):
lgtm, edit: maybe not haha, sorry let me think about this for a minute

crates/isograph_schema/src/process_type_definition.rs
```diff
@@ -626,7 +626,11 @@ fn get_field_objects_ids_and_names(
                         .map(graphql_input_value_definition_to_variable_definition)
                         .collect::<Result<Vec<_>, _>>()?,
                     is_discriminator: false,
-                    variant: SchemaServerFieldVariant::LinkedField,
+                    variant: SchemaServerFieldVariant::LinkedField(
+                        SchemaServerFieldLinkedFieldVariant {
+                            field_id: FieldType::ServerField(next_server_field_id),
```
rbalicki2, line 631: why are we storing the id in the variant but also in the `id` property?

rbalicki2:
Can you investigate https://github.com/isographlabs/isograph/actions/runs/11740976157/job/32708819908?pr=253 ?

```
-      ... on Checkin {\
+      ... on Pet {\
         __typename,\
         id,\
         location,\
@@ -32,7 +32,7 @@ const normalizationAst: NormalizationAst = [
         selections: [
           {
             kind: "InlineFragment",
-            type: "Checkin",
+            type: "Pet",
```

**If this has to do with the changes I suggested, I think we can probably revert to that state and ship as it was** (and just panic in a few places).

----

IMO there's a better refactor, which can be done as part of this PR, but I'm also happy to do it separately. Some background:

- You cannot have a client pointer/inline fragment unless the field is linked
- We don't know, until validating, whether a field (e.g. `node: Node`) is linked or scalar, because `Node` might not have been defined.
- But we do know that certain fields are inline fragments. (And possibly that certain fields are ClientPointers)

So, I think the correct way to model this is through the `associated_data` field. Pre-validation, the associated data field will contain:

- whether it is a linked field/client pointer/etc
- the unvalidated type name

Post-validation, this becomes an enum, and we emit an error if the unvalidated type name is a scalar type and we tried to create a linked field/client pointer.

rbalicki2 (changes requested):

crates/isograph_schema/src/validate_server_field.rs
```diff
@@ -62,11 +76,20 @@ fn validate_and_transform_server_field(
                 description: empty_field.description,
                 name: empty_field.name,
                 id: empty_field.id,
-                associated_data: field_type,
+                associated_data: match field_type.inner_non_null() {
+                    SelectionType::Scalar(scalar_id) => {
+                        SelectionType::Scalar(field_type.map(&mut |_| scalar_id))
```
rbalicki2, line 81: Can we panic if the variant is an `InlineFragment` here? (Maybe that will eventually be an error, but for now it's a bug so we should panic)

PR #258, opened 2024-11-09
Cache tsc with turbo

PatrykWalach:


PatrykWalach:
So I need to figure out if using cache multiple times is merging the results or overriding them 🤔 

```yml
      - name: Cache turbo
        uses: actions/cache@v4
        with:
          path: .turbo
          key: ${{ runner.os }}-turbo-${{ github.sha }}
          restore-keys: |
            ${{ runner.os }}-turbo-
  ```

**Comment by PatrykWalach** [comment]
Ok, these jobs are sequential so there are no overrides

File: turbo.json [hunk 1]
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "tasks": {
+    "compile": {
+      "dependsOn": ["^compile"],
```
rbalicki2, line 5: is this right? Does this mean that we're going to require the Rust compiler to be built before we build the JS?

turbo.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "tasks": {
+    "compile": {
```
rbalicki2, line 4: can we name this something like `compileTypescript` or something?

(Perhaps we should also rename the other ones to indicate that they refer to the Rust side)

rbalicki2 (approved):
rebase and lgtm

turbo.json
PatrykWalach: The arrow ^ has the same meaning as pnpm -r - dependencies of each package need to be compiled first

rbalicki2: Okay TIL

rbalicki2 (approved):
lgtm shipitttt

PR #261, opened 2024-11-11
cache `rust build` with `turbo`

### LINKED ISSUE #19: Re-use cached artifacts from other builds
* Right now, if you don't modify any Rust code, we still rebuild the compiler binary from scratch for all platforms. This can take up to 5 minutes.
* It would be nice to e.g. get a hash of the crates folder (and maybe others?) and re-use uploaded artifacts from previous builds, if the hash matches.

PatrykWalach:
Closes #19

### DIFF
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -50,7 +50,7 @@ jobs:
           command: install
           args: cross
       - name: 'Build isograph_cli with cargo (${{inputs.target}})'
-        run: pnpm exec turbo ${{ inputs.cross && 'cross' || 'build' }} -- -- --target ${{ inputs.target }} --release
+        run: pnpx turbo ${{ inputs.cross && 'cross' || 'build' }} -- -- --target ${{ inputs.target }} --release
       - uses: actions/upload-artifact@v4
         with:
           name: ${{ inputs.artifact-name }}
--- a/.github/workflows/build-cli.yml
+++ b/.github/workflows/build-cli.yml
@@ -27,6 +27,13 @@ jobs:
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v2
+      - name: Cache turbo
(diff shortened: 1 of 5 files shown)


PatrykWalach:
These are in parallel so I've added `artifact-name` to cache key so there's no race conditions

.github/workflows/build-cli.yml
```diff
@@ -27,6 +27,21 @@ jobs:
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v2
+      - name: Cache turbo
+        uses: actions/cache@v4
+        with:
+          path: .turbo
+          key: ${{ runner.os }}-turbo-${{ inputs.artifact-name }}-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-turbo-${{ inputs.artifact-name }}-
```
rbalicki2, line 36: TBH I don't really understand what's happening here, so apologies for the basic question, but should we have `${{ inputs.artifact-name }}-${{ github.sha }}` or the like here?

turbo.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "tasks": {
+    "//#build": {
+      "inputs": ["crates/**", "Cargo.*"],
```
rbalicki2, line 5: Perhaps we should add `relay-crates/**` as well?

rbalicki2 (changes requested):
can you please rebase this on main and I'll review it again?

.github/workflows/build-cli.yml
PatrykWalach: This restores `.turbo` directory that contains all the `hashed` `inputs` and `outputs`.

The `actions/cache` will lookup `key` first, then if it's not available it will look for entry starting with the patterns provided in `restore-keys`.
So this let's us restore cache from previous commit.

Issue #252, opened 2024-11-05
[epic] Isograph router integration


rbalicki2:
This is a sketch of how routing in Isograph might work.

- Certain entrypoints are associated with URLs, e.g. `entrypoint Query.PetDetailPage($id: ID!)` associated with `pet/:id`.
- At build time, a type is generated, such as:
```
type AnyRoute = {
  kind: 'Query.PetDetailPage',
  params: { id: number },
} | /* other routes */
```
creating an instance of this type is assumed to be cheap. i.e. doing so doesn't send a network request!

- Client fields have a `navigateToEntrypoint` function, e.g.:

```
function MyClientField({ data, navigateToEntrypoint, preloadEntrypoint }) {
  const onClick = () => navigateToEntrypoint({ kind: 'Query.PetDetailPage', params: { id: 123 } });
}
```

- The router has a generated function that turns an AnyRoute object into what we need:

```
function preloadRoute(route: AnyRoute, shouldFetch: ShouldFetch): RouteState {
  switch (route.kind) {
    case 'Query.PetDetailPage': {
      // pseudo code here
      const entrypointImport = import('../__isograph/Query/PetDetailPage/entrypoint.tsx')
      entrypointImport.then(value => makeNetworkRequest(value.entrypoint);
      return entrypointImport;
    }
  }
}
```

- Certain entrypoints can be pre-loaded (i.e. hard required) in the router if you choose
- In particular, we can probably pre-fetch: nothing, just the NetworkRequestInfo (with or without normalization Ast), the NetworkRequestInfo and the readerWithRefetchQueries, etc.

THIS WILL BE SO GOOD

rbalicki2:
> Edit: I think this should be its own epic... will create one

I think routing is actually a specific application of state transitions.

# State transitions

## Intro

There are three types of state transitions I'm thinking of:

- changing variables within a mounted query
- mounting a new query, and/or unmounting an existing query

These two encompass routing. The second one is pretty similar to how pointers/loadable queries work.

## State transition function

If we embed a state transition into an Isograph literal, then we know two things:
- what fields must have been fetched for the "state transition function" to be reachable
- what fields must be fetched after the transition completes

So, we should be able to take the diff, and construct a query for exactly the new fields.

## What we can achieve with state transitions

- Routing/navigation, and fetching just the new fields. e.g. if you navigate from list view to detail view, constructing a query that doesn't include any fields already fetched by the list view
[...]

PR #264, opened 2024-11-12
fix error with hot module replacement

PatrykWalach:
When hot module replacement happens the `parentCache` doesn't change but `useEffect` is rerun (don't know what the exact cause of this is), when `useEffect` is rerun it crashes, because it can't access `cacheItem` anymore.

The extra check makes sure effect is run once per `parentCache` (maybe we are compatible with `StrictMode` now?)


libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
```diff
@@ -51,6 +51,10 @@ export function useCachedResponsivePrecommitValue<T>(
   const lastCommittedParentCache = useRef<ParentCache<T> | null>(null);
 
   useEffect(() => {
+    if (lastCommittedParentCache.current === parentCache) {
```
rbalicki2, line 54: Screenshot 2024-11-13 at 6 10 17 PM
I now see this, btw... I'll investigate what is going on.

My suspicion is that the component cache is caching the old component. I haven't looked into this.

I think in terms of strict mode, we're still likely incompatible. Basically, `useLazyReference` will uncommit, dispose, and re-commit, which does not re-issue the query.

I am philosophically opposed to that — queries aren't always re-issuable! But we can probably warn loudly and reissue queries. Just sucks if you're lazily doing a mutation, which is possible :-P

Once #252 lands, there probably won't be a huge need for lazy loaded queries though

rbalicki2:
https://isograph.dev/docs/faq/#why-does-isograph-not-support-strict-mode

PatrykWalach:
Now `useLazyDisposableState` should work correctly, it will setup cleanup once for every `parentCache`

libs/isograph-react-disposable-state/src/useDisposableState.ts
```diff
@@ -47,14 +46,23 @@ export function useDisposableState<T = never>(
...
+        return;
       }
-    };
-  }, []);
+      lastCommittedParentCache.current = parentCache;
+      // capture last set pair in a variable
+      const current = itemCleanupPairRef.current;
+      return () => {
+        // current is a stale variable
+        current?.[1]();
```
PatrykWalach, line 61: This is wrong, maybe
```suggestion
        if (itemCleanupPairRef.current !== null) {
          current?.[1]();
        }
```
would be better, need to think about that

PatrykWalach:
This is not erroring, but the environment cache will indeed produce stale components

rbalicki2:
It's interesting that `iso(entrypoint)` calls don't get HMR'd, because (I think) that works with Relay, and we don't do much that's different`

PatrykWalach:
If you place `console.log` after a resolver it will retrigger on update but if you put it in any artifact importing the resolver it will not retrigger.

rbalicki2:
weird

PatrykWalach:
[This](https://gist.github.com/gaearon/a4d9fb3e6ea487a9296a8d2d9a6e3bf2) post from Dan Abramov explains pretty much exactly how HMR works. The identity of function rendered `<Component />` matters, the way we wrap components in `getOrCreateCachedComponent` we lose their identity. For it to work we would need to `render` resolvers:
```diff
- readerWithRefetchQueries.readerArtifact.resolver(firstParam, additionalRuntimeProps)
+ <readerWithRefetchQueries.readerArtifact.resolver {...firstParameter} />
```
We can't do that because we will lose the second parameter.

## Solutions:
1. We could change `iso` to have runtime impact and wrap the `resolver`
    ```tsx
    let firstParam = Symbol() // maybe symbol works with react?
    const iso = Component => (props) => {
      return Component(props[firstParam], props)
    }
    ```
    then HMR registers wrapped component
    ```tsx
    const Pokemon = iso(({ data, props }) => /* ... */)
    
    __register__(Pokemon, "Poekemon")
    ```
    and we can `render` it as mentioned above.
2. We can hook into the `react-refresh` and rely on its implementation details or create our version.

rbalicki2:
Great find. The gist describes several possible approaches. Do you know which one is actually used by HMR? Is the HMR code adding `__register__(PetUpdater, PetUpdater.ts$PetUpdater)`?

- Will the HMR code understand to add that, even though our components don't look like components? They're wrapped in `iso`, for example.
- I don't understand what's going on. But it makes sense that the caching in the component cache is problematic.

### What I see

On `localhost:3000` and on main:
- If I modify `HomeRoute`, the specific error that I see is that the cache item is accessed before initialization.
- If I modify `PetSummaryCard`, I see that HMR occurs, but it doesn't reflect any changes.
- This is different than the error you see, which I saw in the past

### Random thoughts about two params

- My instinct is to keep the two parameters, because
    - That's what makes it easiest to provide types to the second parameter
    - It means data, etc. aren't special props within the same namespace
    - We preserve our ability to add future props, like `parameters`, `refetch` etc
- But I'm certainly open to the consider making these regular ol' components
- It might be worth experimenting with clobbering everything into a single parameter and seeing if that fixes the HMR issues

### Our own integration with HMR

- It 100% makes sense that the cache in componentCache is the problem. That's global state that's not reset when we do HMR.
- I was thinking that we should move that caching into the Isograph store, so that we get garbage collection and other such benefits for free. But we should consider whether that is compatible with HMR.
- Given all of that, maybe it makes sense for our "HMR integration" to be just deleting items in the Isograph store/out of the component cache? Just a thought. I have no idea if that will work

rbalicki2:
We see that `ReferenceError: Cannot access 'cacheItem' before initialization`, which means that the effect is called but we returned early (ie. `(lastCommittedParentCache.current === parentCache)`).

Okay, just noticing that. I don't know what that implies

PatrykWalach:
> It 100% makes sense that the cache in componentCache is the problem. That's global state that's not reset when we do HMR.

Keeping stale references to resolvers is not an issue.
> I don't understand what's going on. But it makes sense that the caching in the component cache is problematic.

The way HMR works is every component refers to `stale` `<SubComponent />` but react under the hood has a `Map<StaleComponent, LatestComponent>`, whenever a file is modified it updates the entry in the map with `__register__` and rerenders with react.

By calling as a function `StaleComponent()` instead of rendering `<StaleComponent />` we skip the step of `react` reading from the `Map`

PatrykWalach:
> We see that ReferenceError: Cannot access 'cacheItem' before initialization, which means that the effect is called but we returned early (ie. (lastCommittedParentCache.current === parentCache)).

This means you probably didn't run `pnpm -r compile` because this is an error that should be fixed here.

rbalicki2:
Oh sorry, I was on main. Maybe that's why we were seeing different errors

rbalicki2:
> By calling as a function StaleComponent() instead of rendering <StaleComponent /> we skip the step of react reading from the Map

Ohhh sorry. I hadn't read the latest changes in your PR. That's a helpful description.

So HMR is "working", and that's re-running the effect, and the effect is being run despite the component returning early, so we throw in an effect.

That makes sense. 

Can you explain why `useDisposableState` is different than the rest?

I'll try to re-review holistically. Great work!!! And great detective work!!!

PatrykWalach:
HMR seems to work with `iso(function Component { /*...*/ })`
but not with `iso('field ...')(function Component { /*...*/ })`
so it would not work without babel transform 🤔

PatrykWalach:
For no babel transform HMR will revalidate all the files upwards, but we will still rerender with stale components from the cache. To cover this case, We can add the `WeakMap<Resolver, ...>` layer to the cache. Then the only downside is that components will lose their current state.

If the users don't want to use babel transform but want HMR, splitting `hoc` would work but I'm not sure if we can recommend doing that.
```tsx
const isoFoo = iso(`field Foo ...`)

export const Foo = isoFoo(function Foo(){/* ... */})
```
The reason this works is because react has to support builtin `hoc` (`memo` and `forwardRef`)

rbalicki2:
I don't think we need to extensively support the non-babel transform case

rbalicki2:
Can they do

```
export const isoFoo=  iso(`...`)(Component)

function Component() { ... }
```

?

PatrykWalach:
They can't because this double function call is not recognized as `hoc`. Only a single call is recognized. (`const Foo = memo(function Foo ...)`)

PatrykWalach:
If someone doesn't use `babel` they still might use HMR.
They might use something like vite-swc-react and don't want to install babel. For those cases it'd be reasonable to add this extra layer of cache.

PR #265, opened 2024-11-13
chore: Fix vite-demo type checking

lanesawyer:
# chore: Fix vite-demo type checking

## What
- Fixes the `vite-demo` `tsc` command

## How
- Added the `--noEmit` flag. The `tsc` error was about an output file not matching an input file, but that's not really something we're concerned about. So by using the `--noEmit` flag, no compiler output is created, which is what was causing the problem! Types within the demo are still fully checked, so we lose nothing from this tweak

### DIFF
--- a/demos/vite-demo/package.json
+++ b/demos/vite-demo/package.json
@@ -8,7 +8,7 @@
     "build": "tsc -b && vite build",
     "lint": "eslint .",
     "preview": "vite preview",
-    "tsc": "echo TODO figure out why this fails in CI",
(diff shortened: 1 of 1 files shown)


lanesawyer:
Welp, that didn't work lol

I reproduced the errors locally and then fixed it locally, but then it failed on CI 😭

lanesawyer:
`tsc -b` worked, and that would also fail on type errors, so let's go with that! Updating PR description now

PR #281, opened 2024-11-15
fix: Add noUncheckedIndexAccess to tsconfig

### LINKED ISSUE #213: [typescript] Turn on all of the recommended config options
* The stricter the better

lanesawyer:
# fix: Add noUncheckedIndexAccess to tsconfig

## What
- Adds the `noUncheckedIndexAccess` option to the base `tsconfig.json` file
- Fixes the errors that appeared
- Resolves #213 

## How
- Added line to the file
- Fixed issues primarily by throwing errors, as each area of the code expects it to exist and something is very wrong if it doesn't
- Loosened equality checks in some places so all falsey values are caught
- Changed a few `null`s to `undefined`s because accessing an index that doesn't exist (or accessing an array with `undefined` as an index value returns `undefined`)

### DIFF
--- a/libs/isograph-react/src/core/read.ts
+++ b/libs/isograph-react/src/core/read.ts
@@ -643,11 +643,11 @@ function generateChildVariableMap(
   const childVars: Writable<Variables> = {};
   for (const [name, value] of fieldArguments) {
     if (value.kind === 'Variable') {
-      const hting = variables[value.name];
-      if (hting == null) {
-        throw new Error('Variable ' + value.name + ' is not defined');
+      const variable = variables[value.name];
+      if (variable == null) {
+        throw new Error('Variable ' + value.name + ' is not missing');
       }
-      childVars[name] = hting;
+      childVars[name] = variable;
     } else {
       childVars[name] = value.value;
     }
--- a/libs/isograph-react/src/core/read.ts
+++ b/libs/isograph-react/src/core/read.ts
@@ -374,15 +374,13 @@ function readData<TReadFromStore>(
       }
       case 'Resolver': {
         const usedRefetchQueries = field.usedRefetchQueries;
-        const resolverRefetchQueries = usedRefetchQueries.map(
-          (index) => {
-            const resolverRefetchQuery = nestedRefetchQueries[index];
-            if (resolverRefetchQuery == null) {
-                throw new Error('resolverRefetchQuery is null in Resolver');
-            }
-            return resolverRefetchQuery;
-          },
-        );
+        const resolverRefetchQueries = usedRefetchQueries.map((index) => {
+          const resolverRefetchQuery = nestedRefetchQueries[index];
+          if (resolverRefetchQuery == null) {
+            throw new Error('resolverRefetchQuery is null in Resolver');
+          }
+          return resolverRefetchQuery;
+        });
 
         switch (field.readerArtifact.kind) {
           case 'EagerReaderArtifact': {
@@ -645,7 +643,11 @@ function generateChildVariableMap(
   const childVars: Writable<Variables> = {};
   for (const [name, value] of fieldArguments) {
     if (value.kind === 'Variable') {
-      childVars[name] = variables[value.name];
+      const hting = variables[value.name];
+      if (hting == null) {
+        throw new Error('Variable ' + value.name + ' is not defined');
+      }
+      childVars[name] = hting;
     } else {
       childVars[name] = value.value;
     }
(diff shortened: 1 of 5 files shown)


lanesawyer:
@rbalicki2 Did my best on the error throwing messages, let me know if you have more informative ones in mind!

libs/isograph-react/src/core/read.ts
```diff
@@ -371,9 +374,13 @@ function readData<TReadFromStore>(
       }
       case 'Resolver': {
         const usedRefetchQueries = field.usedRefetchQueries;
-        const resolverRefetchQueries = usedRefetchQueries.map(
-          (index) => nestedRefetchQueries[index],
-        );
+        const resolverRefetchQueries = usedRefetchQueries.map((index) => {
+          const resolverRefetchQuery = nestedRefetchQueries[index];
+          if (resolverRefetchQuery == null) {
+            throw new Error('resolverRefetchQuery is null in Resolver');
```
rbalicki2, line 380: Can you please add `this is indicative of a bug in Isograph`?

libs/isograph-react/src/core/read.ts
```diff
@@ -343,9 +343,12 @@ function readData<TReadFromStore>(
         } else {
           const refetchQueryIndex = field.refetchQuery;
           if (refetchQueryIndex == null) {
```
rbalicki2, line 345: oops, this can never be null... the field is not optional. This is probably a holdover from a previous time. perhaps something like https://typescript-eslint.io/rules/no-unnecessary-condition/ can prevent this in the future

libs/isograph-react/src/core/read.ts
```diff
@@ -343,9 +343,12 @@ function readData<TReadFromStore>(
...
-            throw new Error('refetchQuery is null in RefetchField');
+            throw new Error('refetchQueryIndex is null in RefetchField');
           }
           const refetchQuery = nestedRefetchQueries[refetchQueryIndex];
+          if (refetchQuery == null) {
+            throw new Error('refetchQuery is null in RefetchField');
```
rbalicki2, line 350: Can you add that this is indicative of a bug in Isograph

libs/isograph-react/src/core/read.ts
```diff
@@ -636,7 +643,11 @@ function generateChildVariableMap(
   const childVars: Writable<Variables> = {};
   for (const [name, value] of fieldArguments) {
     if (value.kind === 'Variable') {
-      childVars[name] = variables[value.name];
+      const variable = variables[value.name];
+      if (variable == null) {
+        throw new Error('Variable ' + value.name + ' is not missing');
```
rbalicki2, line 648: s/not missing/missing

But also, this breaks if we have a default value for a variable and the variable is not provided, e.g. `field Query.PetDetailRoute($id: ID = 4) @component {` with no variable provided.

I think the appropriate behavior is to continue the loop, i.e. we don't (currently) distinguish between missing and null variables, so skipping the variable is appropriate (and based on my limited testing, works)

libs/isograph-react/src/loadable-hooks/useConnectionSpecPagination.ts
```diff
@@ -114,8 +114,13 @@ export function useConnectionSpecPagination<
         fragmentReference.readerWithRefetchQueries,
       );
 
+      const data = readOutDataAndRecords[i]?.item;
+      if (data == null) {
+        throw new Error('Parameter data is unexpectedly null');
```
rbalicki2, line 119: indicative of a bug in isograph, likewise below and in `useSkipLimit`. Could you also add a comment `// invariant: readOutDataAndRecords.length === completedReferences.length` somewhere near here as well?

rbalicki2 (changes requested):
just a few nits, overall, amazing!!! Love it. Thank you for making the codebase stricter and saving my future self from my now self

libs/isograph-react/src/core/read.ts
lanesawyer: Oh ChatGPT, thank you for being awful lol. I did not even see the "not" lol

And that logic makes sense! Will just allow the loop to continue.

libs/isograph-react/src/core/read.ts
lanesawyer: Done!

libs/isograph-react/src/core/read.ts
lanesawyer: Done!

rbalicki2 (approved):
Looks amazing!! Shipitttt

PR #282, opened 2024-11-15
add magic `link` field

PatrykWalach:
This is a bit awkward at places because
- I didn't extended `Loadability`, should we do that? 
- I don't know what will happen when we select `__link` \@loadably, we should validate it as part of this PR
- `type Link` is not imported right know, but it's not an issue since client pointers are not implemented yet

other then that It seems to work

### DIFF
--- a/crates/isograph_schema/src/add_link_fields.rs
+++ b/crates/isograph_schema/src/add_link_fields.rs
@@ -26,7 +26,7 @@ impl UnvalidatedSchema {
                     variant: ClientFieldVariant::Link,
                     type_and_field: ObjectTypeAndFieldName {
                         field_name,
-                        type_name: "ID".intern().into(),
+                        type_name: object.name,
                     },
                     refetch_strategy: None,
                 }));
--- a/crates/isograph_schema/src/add_link_fields.rs
+++ b/crates/isograph_schema/src/add_link_fields.rs
@@ -1,0 +1,61 @@
+use common_lang_types::{Location, WithLocation};
+use intern::string_key::Intern;
+
+use crate::{
+    ClientField, ClientFieldVariant, ClientType, FieldType, ObjectTypeAndFieldName,
+    ProcessTypeDefinitionError, ProcessTypeDefinitionResult, UnvalidatedSchema,
+};
+
+impl UnvalidatedSchema {
+    /// For each supertype (e.g. Node), add the fields defined on it (e.g. Node.MyComponent)
+    /// to subtypes (e.g. creating User.MyComponent).
+    ///
+    /// We do not transfer server fields (because that makes no sense in GraphQL, but does
+    /// it make sense otherwise??) and refetch fields (which are already defined on all valid
+    /// types.)
+    ///
+    /// TODO confirm we don't do this for unions...
+    pub fn add_link_fields(&mut self) -> ProcessTypeDefinitionResult<()> {
+        for object in &mut self.server_field_data.server_objects {
+            let field_name = "__link".intern().into();
+            let next_client_field_id = self.client_fields.len().into();
+            self.client_fields
+                .push(ClientType::ClientField(ClientField {
+                    description: Some(
+                        format!("A store Link for the {} type.", object.name)
+                            .intern()
+                            .into(),
+                    ),
+                    id: next_client_field_id,
+                    name: field_name,
+                    parent_object_id: object.id,
+                    variable_definitions: vec![],
+                    reader_selection_set: Some(vec![]),
+                    variant: ClientFieldVariant::Link,
+                    type_and_field: ObjectTypeAndFieldName {
+                        field_name,
+                        type_name: "ID".intern().into(),
+                    },
+                    refetch_strategy: None,
+                }));
+
+            if object
+                .encountered_fields
+                .insert(
+                    field_name,
+                    FieldType::ClientField(ClientType::ClientField(next_client_field_id)),
+                )
+                .is_some()
+            {
+                return Err(WithLocation::new(
+                    ProcessTypeDefinitionError::FieldExistsOnType {
+                        field_name,
+                        parent_type: object.name,
+                    },
+                    Location::generated(),
+                ));
+            }
+        }
+        Ok(())
+    }
+}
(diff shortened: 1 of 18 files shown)


rbalicki2:
should we name the field `link`? I think `__link` looks ugly... :-D

PatrykWalach:
And do we fail if there's a link already, or do we ignore?
Ignoring will panic when adding server pointer, but we can warn when that's an issue.

PatrykWalach:
> type Link is not imported right know, but it's not an issue since client pointers are not implemented yet
And this is an issue, because `__link` can be selected manually so I guess I should fix this too

rbalicki2:
I think we should panic with a helpful message. Eventually we can allow the user to provide a directive in a schema extension to rename the `link` field if they have a clash

rbalicki2:
And maybe once we have namespaces, we can have an "internal" namespace, and the user can selected `internal.link` and `internal.asUser` etc or something

No more conflicts! Idk

crates/isograph_schema/src/process_client_field_declaration.rs
```diff
@@ -199,6 +199,7 @@ pub struct UserWrittenClientFieldInfo {
 pub enum ClientFieldVariant {
     UserWritten(UserWrittenClientFieldInfo),
     ImperativelyLoadedField(ImperativelyLoadedFieldVariant),
+    Link,
```
rbalicki2, line 202: Thinking out loud... yes, a link field is a client field. The id aspect of it is magical, though, so yes, it sounds like it should be a separate `ClientFieldVariant`. Okay. Sgtm.

crates/isograph_schema/src/add_link_fields.rs
```diff
@@ -0,0 +1,61 @@
+use common_lang_types::{Location, WithLocation};
+use intern::string_key::Intern;
+
+use crate::{
+    ClientField, ClientFieldVariant, ClientType, FieldType, ObjectTypeAndFieldName,
+    ProcessTypeDefinitionError, ProcessTypeDefinitionResult, UnvalidatedSchema,
+};
+
+impl UnvalidatedSchema {
+    /// For each supertype (e.g. Node), add the fields defined on it (e.g. Node.MyComponent)
```
rbalicki2, line 10: this comment was copy pasted over, but doesn't apply here

crates/isograph_schema/src/add_link_fields.rs
```diff
@@ -0,0 +1,61 @@
...
+                    ),
+                    id: next_client_field_id,
+                    name: field_name,
+                    parent_object_id: object.id,
+                    variable_definitions: vec![],
+                    reader_selection_set: Some(vec![]),
+                    variant: ClientFieldVariant::Link,
+                    type_and_field: ObjectTypeAndFieldName {
+                        field_name,
+                        type_name: "ID".intern().into(),
```
rbalicki2, line 37: `type_name` here should be the parent object's typename.

crates/isograph_schema/src/add_pointers_to_supertypes.rs
```diff
@@ -56,7 +56,31 @@ impl UnvalidatedSchema {
                     Span::todo_generated(),
                 );
 
-                let condition_selection_set = vec![typename_selection];
+                let link_selection = WithSpan::new(
```
rbalicki2, line 59: note to self got to here

rbalicki2 (changes requested):
Some nits, overall looks great and I really like how this is turning out! Sorry that it took me a few days to review this.

crates/isograph_schema/src/add_pointers_to_supertypes.rs
```diff
@@ -56,7 +56,31 @@ impl UnvalidatedSchema {
...
+                    ServerFieldSelection::ScalarField(ScalarFieldSelection {
+                        arguments: vec![],
+                        associated_data: ValidatedScalarFieldAssociatedData {
+                            location: FieldType::ClientField(
+                                match *subtype
+                                    .encountered_fields
+                                    .get(&"link".intern().into())
```
rbalicki2, line 66: nit: can `"link".intern().into()` be defined in a `lazy_static` block and we use that everywhere?

crates/isograph_schema/src/add_pointers_to_supertypes.rs
```diff
@@ -56,7 +56,31 @@ impl UnvalidatedSchema {
...
+                                    .expect("Expected link to exist")
```
rbalicki2, line 67: I don't think it's ideal that `.encountered_fields` is just a hash map. IMO it would be better to have a struct with special slots for all the special fields: `__typename`, an optional `id`, `link`, etc. and a hash map for all of the regular fields. that way, we don't have to do more assertions here.

#295

crates/isograph_schema/src/create_merged_selection_set.rs
```diff
@@ -713,16 +713,22 @@ fn merge_validated_selections_into_selection_map(
...
-                                schema,
-                                parent_map,
-                                merge_traversal_state,
-                                newly_encountered_scalar_client_field,
-                                encountered_client_field_map,
-                                variable_context,
-                                &scalar_field_selection.arguments,
-                            ),
+                            None => match newly_encountered_scalar_client_field.variant {
+                                ClientFieldVariant::Link => {}
```
rbalicki2, line 717: should we merge the `link` field's selection set here? I can't actually get it to exhibit behavior in which the typename is not selected but should be, so I think that would be unnecessary. But can we rely on that behavior?

Anyway, I don't know the right answer here, happy to go with whatever you think is best

rbalicki2 (changes requested):
just one nit and an open question, but otherwise, lgtm and we should land this!!!!

rbalicki2 (changes requested):
just one nit and an open question, but otherwise, lgtm and we should land this!!!!

crates/isograph_schema/src/create_merged_selection_set.rs
PatrykWalach: I think we don't need to merge here, we are guaranteed to have `__typename` since cache uses two layers now

rbalicki2: Ok Sgtm

rbalicki2 (approved):
Shipittt

PR #283, opened 2024-11-19
adjusts vsCode extension to use npm instead of pnpm

saminoorsyed:
Adjusts VS Code extension scripts to use npm instead of pnpm


rbalicki2 (approved):
lgtm

PR #285, opened 2024-11-19
replace std::Error with String

ch1ffa:
For incremental compiler we need errors to implement Clone, so we have to replace `std::io::Error` with `String`


rbalicki2 (approved):
lgtm shipittt

PR #286, opened 2024-11-27
no_babel_transform option & change entrypoint iso overloads to return void

PatrykWalach:


crates/graphql_artifact_generation/src/iso_overload_file.rs
```diff
@@ -13,29 +13,45 @@ use crate::generate_artifacts::ISO_TS;
...
         validated_client_field.type_and_field.field_name
     );
-    s.push_str(&format!(
-        "
+    match on_missing_babel_transform {
+        OptionalValidationLevel::Ignore => (
+            None,
+            format!(
+                "
+export function iso<T>(
```
rbalicki2, line 28: great catch

rbalicki2 (commented):
This makes me think that we should not be using error|ignore|warning for this, but instead have a `"no_babel_transform": bool` option or something. Thoughts?

Anyway, happy to land this as is to unblock Randall, but if you're up for making ☝️ change, we can skip a step

PatrykWalach:
Makes sense, I can make this change

crates/isograph_config/src/compilation_options.rs
```diff
@@ -207,9 +207,7 @@ impl Default for ConfigFileOptionalValidationLevel {
 fn create_options(options: ConfigFileOptions) -> ConfigOptions {
     ConfigOptions {
         on_invalid_id_type: create_optional_validation_level(options.on_invalid_id_type),
-        on_missing_babel_transform: create_optional_validation_level(
-            options.on_missing_babel_transform,
-        ),
+        no_babel_transform: options.no_babel_transform,
```
rbalicki2, line 210: Let's rebase this on main and regenerate the schema as well

crates/graphql_artifact_generation/src/iso_overload_file.rs
```diff
@@ -159,20 +177,13 @@ export function iso(_isographLiteralText: string):
...
 "  throw new Error('iso: Unexpected invocation at runtime. Either the Babel transform ' +
       'was not set up, or it failed to identify this call site. Make sure it ' +
       'is being used verbatim as `iso`.');"
         }
-        OptionalValidationLevel::Warn => {
-
-            "  console.warn('iso: Unexpected invocation at runtime. Either the Babel transform ' +
-      'was not set up, or it failed to identify this call site. Make sure it ' +
-      'is being used verbatim as `iso`.');
-  return (clientFieldResolver: any) => clientFieldResolver;"
```
rbalicki2, line 173: can we modify this error message too? E.g. add "If you cannot use the babel transform, set `options.no_babel_transform` to `true` in your Isograph config."

rbalicki2 (changes requested):
Looks great to me, thanks for identifying this and getting it done. Just a few nits

rbalicki2 (approved):
LGTM shipittt

Issue #181, opened 2024-10-17
Use TypeAnnotation instead of GraphQLTypeAnnotation in VariableDefinition


rbalicki2:
- in VariableDefinition, we use GraphQLTypeAnnotation. This is (probably) the last use of GraphQLTypeAnnotation in IsographSchema.
- It will probably make it easier to do type checking of variables if both the variables and the object types use TypeAnnotation

PR #291, opened 2024-12-20
Type check variables

PatrykWalach:
I've used `GraphQLTypeAnnotation` instead of `TypeAnnotation` suggested in #181

I am not checking `enums` and `object literals` yet.

Object literals require checking `extraneous` and `missing` `fields` just like the `arguments` maybe it's possible to make a single function that can be used for both of these.

### DIFF
--- a/crates/graphql_lang_types/src/graphql_type_annotation.rs
+++ b/crates/graphql_lang_types/src/graphql_type_annotation.rs
@@ -9,6 +9,25 @@ pub enum GraphQLTypeAnnotation<TValue> {
     NonNull(Box<GraphQLNonNullTypeAnnotation<TValue>>),
 }
 
+impl<TValue> From<GraphQLTypeAnnotation<TValue>> for GraphQLNonNullTypeAnnotation<TValue> {
+    fn from(value: GraphQLTypeAnnotation<TValue>) -> Self {
+        match value {
+            GraphQLTypeAnnotation::Named(named) => GraphQLNonNullTypeAnnotation::Named(named),
+            GraphQLTypeAnnotation::List(list) => GraphQLNonNullTypeAnnotation::List(*list),
+            GraphQLTypeAnnotation::NonNull(non_null) => *non_null,
+        }
+    }
+}
+
+impl<TValue> From<GraphQLNonNullTypeAnnotation<TValue>> for GraphQLTypeAnnotation<TValue> {
+    fn from(value: GraphQLNonNullTypeAnnotation<TValue>) -> Self {
+        match value {
+            GraphQLNonNullTypeAnnotation::Named(named) => GraphQLTypeAnnotation::Named(named),
+            GraphQLNonNullTypeAnnotation::List(list) => GraphQLTypeAnnotation::List(Box::new(list)),
+        }
+    }
+}
+
 impl<TValue> GraphQLTypeAnnotation<TValue> {
     pub fn inner(&self) -> &TValue {
         match self {
--- a/crates/isograph_schema/src/isograph_schema.rs
+++ b/crates/isograph_schema/src/isograph_schema.rs
@@ -89,16 +89,6 @@ pub struct Schema<TSchemaValidationState: SchemaValidationState> {
     pub entrypoints: TSchemaValidationState::Entrypoint,
     pub server_field_data: ServerFieldData,
 
-    // Well known types
-    pub id_type_id: ServerScalarId,
-    pub string_type_id: ServerScalarId,
-    pub float_type_id: ServerScalarId,
-    pub boolean_type_id: ServerScalarId,
-    pub int_type_id: ServerScalarId,
-    // TODO restructure UnionTypeAnnotation to not have a nullable field, but to instead
-    // include null in its variants.
-    pub null_type_id: ServerScalarId,
-
     /// These are root types like Query, Mutation, Subscription
     pub fetchable_types: BTreeMap<ServerObjectId, RootOperationName>,
 }
@@ -185,6 +175,16 @@ pub struct ServerFieldData {
     pub server_objects: Vec<SchemaObject>,
     pub server_scalars: Vec<SchemaScalar>,
     pub defined_types: HashMap<UnvalidatedTypeName, SelectableServerFieldId>,
+
+    // Well known types
+    pub id_type_id: ServerScalarId,
+    pub string_type_id: ServerScalarId,
+    pub float_type_id: ServerScalarId,
+    pub boolean_type_id: ServerScalarId,
+    pub int_type_id: ServerScalarId,
+    // TODO restructure UnionTypeAnnotation to not have a nullable field, but to instead
+    // include null in its variants.
+    pub null_type_id: ServerScalarId,
 }
 
 impl<TSchemaValidationState: SchemaValidationState> Schema<TSchemaValidationState> {
--- a/crates/isograph_schema/src/validate_client_field.rs
+++ b/crates/isograph_schema/src/validate_client_field.rs
@@ -1,14 +1,23 @@
-use std::collections::{BTreeSet, HashMap};
+use std::{
+    collections::{BTreeSet, HashMap},
+    vec,
+};
 
 use common_lang_types::{
-    FieldArgumentName, Location, SelectableFieldName, UnvalidatedTypeName, VariableName,
-    WithLocation, WithSpan,
+    EnumLiteralValue, FieldArgumentName, Location, SelectableFieldName, VariableName, WithLocation,
+    WithSpan,
+};
+
+use graphql_lang_types::{
+    GraphQLListTypeAnnotation, GraphQLNamedTypeAnnotation, GraphQLNonNullTypeAnnotation,
+    GraphQLTypeAnnotation,
 };
 use intern::{string_key::Intern, Lookup};
 use isograph_lang_types::{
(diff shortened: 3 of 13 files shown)


crates/isograph_schema/src/isograph_schema.rs
```diff
@@ -185,6 +175,16 @@ pub struct ServerFieldData {
     pub server_objects: Vec<SchemaObject>,
     pub server_scalars: Vec<SchemaScalar>,
     pub defined_types: HashMap<UnvalidatedTypeName, SelectableServerFieldId>,
+
+    // Well known types
+    pub id_type_id: ServerScalarId,
```
rbalicki2, line 180: all of these types implement copy, they might not need to be moved. I don't really have an opinion on where they go, just commenting

crates/graphql_lang_types/src/graphql_type_annotation.rs
```diff
@@ -9,6 +9,25 @@ pub enum GraphQLTypeAnnotation<TValue> {
     NonNull(Box<GraphQLNonNullTypeAnnotation<TValue>>),
 }
 
+impl<TValue> From<GraphQLTypeAnnotation<TValue>> for GraphQLNonNullTypeAnnotation<TValue> {
+    fn from(value: GraphQLTypeAnnotation<TValue>) -> Self {
+        match value {
+            GraphQLTypeAnnotation::Named(named) => GraphQLNonNullTypeAnnotation::Named(named),
```
rbalicki2, line 15: This converting `Foo` to `Foo!`... if we need this, we should not use `From` for this, since it will be easy to unintentionally do this

rbalicki2 (commented):
partial review

crates/graphql_lang_types/src/graphql_type_annotation.rs
PatrykWalach: I want to avoid copying here, I could use something like this instead, do you have any better ideas?
```rust
    pub fn fold_non_null<OnNamed, OnList, TNewValue>(
        &self,
        on_named: OnNamed,
        on_list: OnList,
    ) -> TNewValue
    where
        OnNamed: FnOnce(&GraphQLNamedTypeAnnotation<TValue>) -> TNewValue,
        OnList: FnOnce(&GraphQLListTypeAnnotation<TValue>) -> TNewValue,
    {
        match self {
            GraphQLTypeAnnotation::Named(named) => on_named(named),
            GraphQLTypeAnnotation::List(list) => on_list(list),
            GraphQLTypeAnnotation::NonNull(non_null) => match non_null {
                GraphQLNonNullTypeAnnotation::Named(named) => on_named(named),
                GraphQLNonNullTypeAnnotation::List(list) => on_list(list),
            },
        }
    }
    ```

Comment (PatrykWalach) [reply crates/graphql_lang_types/src/graphql_type_annotation.rs:15]: With this I'd copy only in the case I encounter two `non_null` types because the function takes two nullable types, and the only way I can think of solving this is to make a new enum
```rust
enum GraphQLMaybeNullTypeAnnotation<T>{
  NonNull(GraphQLNonNullTypeAnnotation<T>),
  Null(GraphQLTypeAnnotation<T>)
}
```

rbalicki2: I'll think about this the next time I review this... which will most likely be in 2025 🎉, for now I need to pass out!

Though it's not a huge deal if we have to do some cloning here, these are really small objects

crates/isograph_schema/src/isograph_schema.rs
PatrykWalach: I need types of scalars to type check `number` and `string` literals. But `validate_and_transform_client_fields` doesn't have access to the invalidated schema, just to `ServerFieldData`. I can reverse this and pass `UnvalidateSchema` if that's what you prefer.

crates/isograph_schema/src/validate_client_field.rs
```diff
@@ -619,43 +632,352 @@ fn get_missing_arguments_and_validate_argument_types<'a>(
...
+        None => Err(WithLocation::new(
+            ValidateSchemaError::UsedUndefinedVariable {
+                undefined_variable: *variable_name,
+            },
+            location,
+        )),
+    }
+}
+
+pub fn get_missing_arguments_and_validate_types(
```
rbalicki2, line 940: doesn't need to be pub

Also this is confusingly similarly named to `get_missing_arguments_and_validate_argument_types` :/ maybe this would be clearer if we had `fn get_missing_and_provided_arguments() -> MissingAndProvidedArguments` and `struct MissingAndProvidedArguments { missing: Vec<...>, provided: Vec<...> }` or something. Then, we can call that, and validate the provided args separate.

(Bonus points if we return an iterator of `EitherProvidedOrMissingArg` 😂)

This, like the renaming variables, could be separate PRs

crates/isograph_schema/src/validate_client_field.rs
```diff
@@ -619,43 +632,352 @@ fn get_missing_arguments_and_validate_argument_types<'a>(
...
+    schema_data: &ServerFieldData,
+    argument_definitions: &[&ValidatedVariableDefinition],
     arguments: &[WithLocation<SelectionFieldArgument>],
     include_optional_args: bool,
-) -> Vec<ValidatedVariableDefinition> {
+    variable_definitions: &[WithSpan<ValidatedVariableDefinition>],
+) -> ValidateSchemaResult<Vec<ValidatedVariableDefinition>> {
     argument_definitions
+        .iter()
         .filter_map(|definition| {
```
rbalicki2, line 949: nit: it's kind of hard to track what's going on here. Can we (throughout this file, and ideally as a separate PR) take a pass to rename variables to clearer?

e.g. `argument_definitions` -> `field_argument_definitions`, `argument_definitions_vec` -> `field_argument_definitions_vec`, `argument` -> `selection_supplied_arguments`, `user_supplied_argument` -> `selection_supplied_argument`

crates/isograph_schema/src/validate_client_field.rs
```diff
@@ -619,43 +632,352 @@ fn get_missing_arguments_and_validate_argument_types<'a>(
...
+    Err(WithLocation::new(
+        ValidateSchemaError::ExpectedType { expected, actual },
+        location,
     ))
 }
 
-pub fn get_missing_arguments<'a>(
-    argument_definitions: impl Iterator<Item = &'a ValidatedVariableDefinition> + 'a,
+fn value_satisfies_type(
+    value: &WithLocation<NonConstantValue>,
```
rbalicki2, line 770: nit: `selection_supplied_argument_value` and `field_argument_definition_type` would be clearer

crates/isograph_schema/src/validate_client_field.rs
```diff
@@ -619,43 +632,352 @@ fn get_missing_arguments_and_validate_argument_types<'a>(
...
+                    // or Value satisfies Value
+                    if named_variable_type.item == named_type.item {
+                        return Ok(());
+                    }
+                }
+                GraphQLNonNullTypeAnnotation::List(_) => (),
+            }
+        }
+        GraphQLTypeAnnotation::NonNull(non_null_argument_type) => match variable_type {
+            // Value! satisfies Value!
```
rbalicki2, line 745: ¡Olé!

crates/isograph_schema/src/validate_client_field.rs
```diff
@@ -619,43 +632,352 @@ fn get_missing_arguments_and_validate_argument_types<'a>(
...
+                GraphQLNonNullTypeAnnotation::Named(named_type) => enum_satisfies_type(
+                    enum_literal_value,
+                    &named_type,
+                    schema_data,
+                    value.location,
+                ),
+            }
+        }
+        NonConstantValue::Null => match type_ {
+            GraphQLTypeAnnotation::NonNull(_) => Err(WithLocation::new(
```
rbalicki2, line 824: `type_.is_nullable()`

rbalicki2 (changes requested):
Great work. I don't think there's anything majorly substantial with respect to the type checking.

Let's either split this up into up to three PRs:
- renaming variables
- the fn to split the vector into unused and used variables
- the validation work

Or if you prefer, and we can land this and have a cleanup pass afterward

PatrykWalach:
Right now if we have two types `[Foo]` and `[Bar]` we will get error: "expected `Foo`, received `Bar`" because of recursion, should we display the whole types?

rbalicki2:
Yeah, it would be great if we displayed the whole type. `GraphQLTypeAnnotation` implements `Display`, so we should be able to do that.

This probably (annoyingly) requires passing down the argument definition and provided argument all the way down; perhaps those can be wrapped in some sort of immutable "context" struct for clarity. It would be confusing to pass down `argument_definition` and `original_argument_definition_for_error` 😭

crates/isograph_schema/src/validate_schema.rs
```diff
@@ -386,6 +387,40 @@ pub enum ValidateSchemaError {
         argument_type: UnvalidatedTypeName,
     },
 
+    #[error("Expected input of type {expected_type}, found variable {variable_name} of type {variable_type}")]
```
rbalicki2, line 390: Can we add the argument_name here?
```
The argument `${argument_name}` has type `${expected_type}`, but the variable $${variable_name} has type `${variable_type}`
```

Also notice the backticks and double $

We're not very consistent about the backticks :/

rbalicki2 (approved):
LGTM, just a nit for a future PR about the error messages

PR #290, opened 2024-12-12
Generate schema for `isograph.config.json`

PatrykWalach:

PR #292, opened 2024-12-21
Generate schema for `isograph.config.json`

PatrykWalach:
Supersedes #290


crates/isograph_config/src/compilation_options.rs
```diff
@@ -203,6 +209,10 @@ impl Default for ConfigFileOptionalValidationLevel {
         Self::Error
     }
 }
+pub fn generate_json_schema_config() {
```
rbalicki2, line 212: Can you move this to `main.rs`? Seems weird to have side effectful functions here

.github/workflows/ci.yml
```diff
@@ -348,6 +348,8 @@ jobs:
       #   # TODO do this less hackily
       - name: Build
         run: pnpm -r compile
+      - name: Build json schema
+        run: pnpm build-json-schema
```
rbalicki2, line 352: - let's instead have a step where we ensure that the checked in `isograph-config-schema.json` file is unchanged after we run `build-json-schema`, and have this run as a dependency of `all-checks-passed`

rbalicki2 (changes requested):
Looks amazing, love it, just a few nits

crates/isograph_config/src/compilation_options.rs
```diff
@@ -80,9 +81,13 @@ impl Default for OptionalValidationLevel {
     }
 }
 
-#[derive(Deserialize)]
+#[derive(Deserialize, JsonSchema)]
 #[serde(deny_unknown_fields)]
-struct ConfigFile {
+pub struct ConfigFile {
```
rbalicki2, line 86: this causes `title: "ConfigFile"` in the schema file, can we change that to IsographProjectConfig or something?

rbalicki2 (changes requested):
How are you able to get validation to work with a path that starts with `..`? If I copy-paste isograph-config-schema.json into the same folder as the isograph.config.json file, I can get vscode to validate it correctly, but if I have it in `../../whatever`, it shows `Unable to parse content from '/Users/rbalicki/code/isograph/libs/isograph-compiler/isograph-config-schema.json': Parse error at offset 0.`

Do you see this?

PatrykWalach:
I do get `Unable to parse content` error but it's not limited to isograph, I get it in my relay project where I use `./node_modules/relay-compiler/relay-compiler-config-schema.json` so maybe it's a vs code error. It used to work just fine.

rbalicki2:
re: vscode not accepting it, sgtm

.prettierignore
```diff
@@ -5,3 +5,4 @@
 **/.docusaurus/**
 **/dist/**
 **/relay-crates/**
+isograph-config-schema.json
```
rbalicki2, line 8: Is there a reason we should ignore the json file? It seems like we should always check in the latest version, just like we ensure that the demo artifacts are current (etc)

Wdyt?

PatrykWalach: This is prettier ignore, we don't format any of the artifacts.

rbalicki2: oops, you're right

rbalicki2: So I guess, why are we uploading the artifact (on linux)?

The way I see it, we have the artifact checked in, so instead of uploading and downloading it, we can just use the checked in version?

PatrykWalach: To check if it's been built before check-in

rbalicki2 (approved):
LFGGG looks great

PR #293, opened 2024-12-22
add pico and pico_macros crates

ch1ffa:
This PR adds 2 new crates `pico` and `pico_macros` that implement building blocks for a new Isograph incremental compiler:

the `Database` struct that holds all information about the compilation state:

```rs
pub struct Database {
    pub current_epoch: u64,
    pub dependency_stack: Vec<Vec<(u64, Dependency)>>,
    pub nodes: LruCache<NodeId, DerivedNode>,
    pub sources: LruCache<NodeId, SourceNode>,
    pub values: LruCache<NodeId, Box<dyn DynEq>>,
    pub params: LruCache<ParamId, Box<dyn Any>>,
}
```

and 2 procedural macros:

1. `#[memo]`
2. `#[source]`

The `#[memo]` macro should be applied to functions we want to memoize. The first argument of the memoized function must be a mutable reference to the `Database` structure. The macro generates code that:

- wraps the function call and memoizes the result in the `values` field.
- stores arguments in the `params` field.
- creates a `DerivedNode` structure to track dependencies and the epoch number when the function was last updated/verified.

The `#[source]` macro should be applied to a struct that represents a source value. The macro generates `get`, `set`, and `remove` methods:

- Use the `set` and `remove` methods to update or remove source values. These actions increment `current_epoch` internally.
- Use the `get` method to retrieve a value from the database. This ensures proper downcasting and tracks the `get` call as a dependency.

Nodes, parameters, and values are stored in an LRU Cache to restrict memory usage and remove unused nodes. Cache eviction may lead to a compiler panic if a required node is unexpectedly missing but most such cases seems to be covered:

- Before using memoization, we check if the required node and value exist in the cache and recalculate the function if something is missing.
- If a cache miss occurs during the dependency check stage, we can treat the dependency as changed and call the node's inner function, recalculating the subtree and recreating evicted nodes.
- db.memo function internally returns NodeId and retrieves references or owned values when needed, but it is safe because we have a check before memo call
- It is safe to get parameters from the cache inside `inner_fn` since parameters are recalculated before each memoized function call. If parameters are missing during dependency checks (without a memoized function call), we can assume the dependency has changed.


crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,134 @@
...
+        block,
+        attrs,
+    } = parse_macro_input!(item as ItemFn);
+
+    let fn_name = sig.ident.clone().to_string();
+
+    let inputs = sig.inputs.iter().collect::<Vec<_>>();
+
+    if inputs.is_empty() {
+        return syn::Error::new_spanned(&sig, "Function must have at least one argument")
```
rbalicki2, line 18: nit: maybe we can specify that the first arg must be the db (though I don't know whether we can enforce that?)

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,134 @@
...
+            .to_compile_error()
+            .into();
+    }
+
+    let db_arg = match &inputs[0] {
+        FnArg::Typed(PatType { pat, .. }) => pat.clone(),
+        _ => {
+            return syn::Error::new_spanned(inputs[0], "Expected a typed argument")
```
rbalicki2, line 26: is this also `unreachable`, as below?

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,134 @@
...
+                    .as_any()
+                    .downcast_ref::<#return_type>()
+                    .expect("unexpected return type. This is indicative of a bug in Isograph.")
+                    .clone()
+            },
+            return_type,
+        )
+    };
+
+    let output = if inputs.len() == 2 {
```
rbalicki2, line 84: do we need to special case this? In the non-length-two case, we wrap the other args in a tuple. Can we do the same here (i.e. a length-one tuple)?

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,134 @@
...
+                .into()
+        }
+    };
+
+    let other_args = inputs.iter().skip(1).map(|arg| match arg {
+        FnArg::Typed(PatType { pat, .. }) => pat,
+        _ => unreachable!(),
+    });
+
+    let argument_types = inputs.iter().skip(1).map(|arg| match arg {
```
rbalicki2, line 37: nit: if we make this mutable, we can avoid a clone below. Likewise for `let other_args`

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,134 @@
...
+        FnArg::Typed(PatType { ty, .. }) => ty,
+        _ => unreachable!(),
+    });
+
+    let return_type = match &sig.output {
+        ReturnType::Type(_, ty) => ty.clone(),
+        ReturnType::Default => {
+            return syn::Error::new_spanned(
+                &sig.output,
+                "Memoized function must have a return type",
```
rbalicki2, line 47: We should be able to relax this. If a memoized function returns `()` and that isn't specified, we will run into this. I mean, returning `()` isn't very useful TBH, but we already support that (if the user specifies `()`) so we should probably also support the anonymous return type too

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,134 @@
...
+            )
+            .to_compile_error()
+            .into()
+        }
+    };
+
+    let (return_expr, return_type) = if let syn::Type::Reference(type_ref) = *return_type {
```
rbalicki2, line 54: Some questions: what if the return type is aliased, as in `type Foo = &Bar`? Is whether it is a reference evaluated at the syntactic level? Will the resulting code will fail to compile, because `downcast_ref::<&T>` doesn't work? (TBH I have no idea!)

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,134 @@
...
+                #db_arg
+                    .values
+                    .get(&node_id)
+                    .expect("value should exist. This is indicative of a bug in Isograph.")
+                    .as_any()
+                    .downcast_ref::<#base_type>()
+                    .expect("unexpected return type. This is indicative of a bug in Isograph.")
+            },
+            base_type,
+        )
```
rbalicki2, line 67: can you add a comment/explain what's going on here?

AFAICT if you return a `&Foo`, we use the base type (`Foo`) later when we call `let value: Foo = (|| #block)();`. But would't value have type `&Foo` here?

crates/pico_macros/src/source.rs
```diff
@@ -0,0 +1,58 @@
...
+use syn::{parse_macro_input, DeriveInput};
+
+pub(crate) fn source(_args: TokenStream, item: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(item as DeriveInput);
+
+    let struct_name = input.clone().ident;
+    let key = struct_name.to_string();
+
+    let output = quote! {
+        #[derive(Debug, Clone, PartialEq, Eq)]
```
rbalicki2, line 12: I think it's common practice to have the user derive Debug, Clone, etc. themselves, in case they require a custom impl. Wdyt?

crates/pico_macros/src/source.rs
```diff
@@ -0,0 +1,58 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{parse_macro_input, DeriveInput};
+
+pub(crate) fn source(_args: TokenStream, item: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(item as DeriveInput);
+
+    let struct_name = input.clone().ident;
```
rbalicki2, line 8: nit: `input.ident.clone()`

crates/pico_macros/src/source.rs
```diff
@@ -0,0 +1,58 @@
...
+    let key = struct_name.to_string();
```
rbalicki2, line 9: nit can we rename this something clearer, like `struct_name_static_str`? I looked at this and thought that `struct_name` and `key` were redundant at first

crates/pico_macros/src/source.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    let output = quote! {
+        #[derive(Debug, Clone, PartialEq, Eq)]
+        #input
+
+        impl #struct_name {
+            pub fn set(
+                self,
+                db: &mut pico::database::Database,
+                static_key: &'static str,
```
rbalicki2, line 19: can you explain what `static_key` is doing? Is this some stable key for a given source node?

I would hope that this can be transparent somehow. Perhaps require that `Input` derive a trait, and have a `static_key` fn on that trait?

rbalicki2 (changes requested):
Amazing work!!!! I can't wait to incorporate this into the compiler.

So far, I've just started reviewing the macros. But I will finish reviewing at some point! It may take me a few days :)

crates/pico/tests/integration_tests.rs
```diff
@@ -0,0 +1,25 @@
...
+    let mut input = Input {
+        value: "2 + 2 * 2".to_string(),
+    };
+    input.set(&mut db, key);
+    let mut result = evaluate_input(&mut db, key);
+    assert_eq!(result, 6);
+
+    input = Input {
+        value: "(2 + 2) * 2".to_string(),
+    };
```
rbalicki2, line 21: can we add some tests that ensure that:

- if the overall expression hasn't changed, it isn't re-evaluated?
- if a subexpression hasn't changed, it isn't re-evaluated? e.g. if we have (1 + 2) * 3 and later calculate (1 + 2) * 4, we do not recalculate (1 + 2)?

This might require some restructuring of the calc to e.g. call a `evaluate_input` recursively

crates/pico/src/database.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        dependencies
+            .into_iter()
+            .any(|(dep_node_id, time_verified)| match dep_node_id.kind {
+                NodeKind::Source => self.source_node_changed(&dep_node_id, time_verified),
+                NodeKind::Derived => self.derived_node_changed(&dep_node_id),
+            })
+    }
+
+    fn source_node_changed(&mut self, node_id: &NodeId, time_verified: u64) -> bool {
```
rbalicki2, line 111: nit: can we either use a newtype around epochs, or at least a type alias?

crates/pico_macros/src/memo.rs
ch1ffa: yeah, I think we can check that type is `&mut Database`

crates/pico_macros/src/memo.rs
ch1ffa: yeah, forgot to cleanup!

crates/pico_macros/src/memo.rs
ch1ffa: I thought what it would be a better solution here and went for a special case. But length-one tuple could simplify code, let me try it and see how it goes

crates/pico_macros/src/source.rs
ch1ffa: for a source it could be a filename or iso literal name. Salsa allows to put #[id] macro on some unique field in the struct. We can go this way as well and require static_key to be a part of a struct

crates/pico/tests/integration_tests.rs
ch1ffa: I need to think how to expose this info. Currently we don't know how exactly we got the value outside of memo function

crates/pico_macros/src/source.rs
rbalicki2: IMO a method is the best in the long run (or if there are non-Isograph users of pico), so that it can be derived from other fields, but yeah, if we want to make it a required field, that's fine too! We can always change it later

crates/pico_macros/src/memo.rs
rbalicki2: It should generate the same code... I think! A length-one tuple is the same at runtime (in terms of memory etc) and just differs at the type level. e.g. it's a newtype

crates/pico_macros/src/memo.rs
rbalicki2: Would that mean this would fail if you imported Database as `PicoDb` or somethig?

rbalicki2: (And if we *can* enforce that, hot damn, that opens up a bunch of possibilities)

crates/pico_macros/src/source.rs
ch1ffa: Yeah, let's begin with a trait. It would be easier to intern keys if needed. And we can replace it later if we find out that most of the source structs are simple - then we will be able to generate new() method by a macro that could intern keys and implement this trait automatically.

rbalicki2: sgtm

crates/pico_macros/src/memo.rs
ch1ffa: Imagine we have a function that returns a value. In our current implementation it cannot be a reference, we store owned values by cloning it to the database. But if we memoize this function we can return a reference if needed. I implemented it just by analysing function return type, so if we put `&` on return type, it means we want to return a reference from the database (but we still store cloned value). Yeah, aliases would not work here, but we should not use an alias in memoized function because we don't store references.

rbalicki2: Hm. 

```
#[memo]
fn first_user(db: Database<'a>): &'a User { return db.list_of_users.iter().next() }
```

I don't think that we can memoize that by cloning the user and returning the reference, right? That defeats the purpose, right? What if the user is a very large object?

crates/pico/src/database.rs
```diff
@@ -0,0 +1,173 @@
...
+        param_id: ParamId,
+        inner_fn: impl Fn(&mut Database, ParamId) -> Box<dyn DynEq>,
+    ) -> (
+        Box<dyn DynEq>,  /* value */
+        Vec<Dependency>, /* dependencies */
+        u64,             /* time_calculated */
+    ) {
+        self.dependency_stack.push(vec![]);
+        let value = inner_fn(self, param_id);
+        let registred_dependencies = self
```
rbalicki2, line 140: typo: `registred`

Also, perhaps we should make a separate function that does something like

```
let (result, dependencies) = with_additional_item_on_dependency_stack(
  || do_something()
);
```

crates/pico/src/database.rs
```diff
@@ -0,0 +1,173 @@
...
+                deps.push(dep);
+                max_time_calculated = std::cmp::max(max_time_calculated, time_calculated);
+                (deps, max_time_calculated)
+            },
+        );
+        (value, dependencies, time_calculated)
+    }
+
+    pub fn register_dependency(&mut self, node_id: NodeId, time_calculated: u64) {
+        if let Some(dependencies) = self.dependency_stack.last_mut() {
```
rbalicki2, line 157: should we panic if none is found?

rbalicki2 (commented):
just a few small things I noticed

crates/pico/src/database.rs
ch1ffa: Yeah, like it!

crates/pico/src/database.rs
ch1ffa: I'll add it just in case. `inner_fn` is always called after `push(vec![])` but it is not guaranteed if we change something accidentally

ch1ffa: Oh, actually I made it on purpose to avoid unnecessary register call on a root memoized function. This is the only place where the dependency stack can be empty as all other inner_fn call wrapped in `dependency_stack.push(vec![])`

crates/pico/tests/integration_tests.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Source)]
+struct Input {
+    #[key]
+    pub key: &'static str,
+    pub value: String,
+}
+
+#[memo]
+pub fn parse_ast(db: &mut Database, id: SourceId<Input>) -> Result<Program> {
```
rbalicki2, line 49: Can you add a really basic, easy-to-follow test like:

```
#[memo]
function both_sides(db: &mut Database, left_id: SourceId<Side>, right_id: SourceId<Side>) -> String {
  both_sides_counter.increment();
  let left = calc_side(db, left_id)
  let right = calc_side(db, right_id)
  format!("{} - {}", left, right)
}

function calc_side(&mut Database, id: SourceId<Side>) {
  side_counter_map.get(id).increment();
  let side = db.get(id);
  format!("side text {}", side.text)
}

#[source]
struct Side { text: String }
```

I mean this is a really silly example with no semantic meaning, so maybe it could be something slightly more meaningful, like formatting parts of an address? Idk

crates/pico_core/src/container.rs
```diff
@@ -0,0 +1,9 @@
+use std::fmt::Debug;
+
+pub trait Container<K, V>: Debug {
+    fn contains_key(&self, k: &K) -> bool;
+    fn get(&self, k: &K) -> Option<&V>;
+    fn get_mut(&mut self, k: &K) -> Option<&mut V>;
+    fn insert(&mut self, k: K, v: V) -> Option<V>;
```
rbalicki2, line 7: can these be commented? e.g. does `insert` return the value that was replaced?

crates/pico/tests/integration_tests.rs
```diff
@@ -0,0 +1,62 @@
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use calc::{ast::Program, error::Result, eval::eval, lexer::Lexer, parser::Parser};
+use pico::storage::DefaultStorage;
+use pico_core::{database::Database as Db, source::SourceId};
```
rbalicki2, line 5: looks like we're clobbering the import of `Database` here, is that by accident? I suppose it can be in scope and clobbered, but that's probably unintended!

crates/pico_macros/src/db.rs
```diff
@@ -0,0 +1,70 @@
...
+                use ::pico_core::storage::StorageMut;
+                self.storage_mut().increment_epoch()
+            }
+
+            fn get<T: Clone + 'static>(&mut self, id: ::pico_core::source::SourceId<T>) -> T {
+                use ::pico_core::{storage::Storage, container::Container};
+                let time_calculated = self.storage()
+                    .sources()
+                    .get(&id.key)
+                    .expect("node should exist. This is indicative of a bug in Isograph.")
```
rbalicki2, line 34: if we expect others to use pico (we should! they should!), maybe we should say `bug in Pico`

crates/pico_macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "pico_macros"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
```
rbalicki2, line 5: nit: we need some way (not necessarily in this PR) to start publishing crates in CI

crates/pico_macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[dependencies]
+pico_core = { path = "../pico_core" }
+proc-macro2 = "1.0"
+syn = "2.0.87"
+quote = "1.0"
+darling = "0.20.10"
```
rbalicki2, line 12: can we try to do `workplace = true` for all dependencies of non-iso non-relay crates?

rbalicki2 (commented):
very partial review

rbalicki2 (commented):
moar incremental reviews

crates/pico_core/src/params.rs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        self.get_ref(db).cloned()
+    }
+
+    pub fn get_ref<'db, T, Db>(&self, db: &'db mut Db) -> Option<&'db T>
+    where
+        T: Hash + Clone + 'static,
+        Db: Database,
+    {
+        db.storage_mut().params().get(self)?.downcast_ref::<T>()
```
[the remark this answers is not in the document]
ch1ffa: tbh, this method is not used outside of param_id, I've added it to think later where and how we can use parameters without cloning. In get() method downcasting is made by macro so should be safe (but we use expect() on option there)

crates/pico_core/src/container.rs
```diff
@@ -0,0 +1,9 @@
+use std::fmt::Debug;
+
+pub trait Container<K, V>: Debug {
```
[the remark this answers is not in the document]
ch1ffa: Not sure where to put it.

In `db.memo()` function we always have a fallback to `inner_fn` call whenever we've lost something. And for the 2 places where we still have `expect()` there is a check if node exists in the storage (we can't use `Some(node)` due to a mutable borrow later).

Source must exist before any root function call because `db.get(id)` could be called inside some memoized function. But it should exist because that's how we get SourceId<T> by calling `db.set(source)`.

In general, it is safe to remove derived or source node if it is not a root function or not a part of any dependency.

crates/pico_core/src/node.rs
```diff
@@ -0,0 +1,61 @@
...
+string_key_newtype!(NodeKey);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct NodeId {
+    pub key: NodeKey,
+    pub param_id: ParamId,
+}
+
+impl NodeId {
+    pub fn new(key: &'static str, param_id: ParamId) -> Self {
```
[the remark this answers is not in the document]
ch1ffa: Oh, `NodeKey` is known at compile time - it is a function name. So maybe we should just calculate its hash and use NodeKey(u64)?

crates/pico_core/src/container.rs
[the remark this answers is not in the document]
rbalicki2: Awesome, that's so good. So safe. Okay, then let's add the docs later

crates/pico_core/src/node.rs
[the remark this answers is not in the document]
ch1ffa: I've just realised that fn name is not unique. And we don't have source file name for proc macro in stable rust yet, so probably I have to use full fn signature hash as a key instead of fn name... Anyway it's all compile time stuff

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,180 @@
...
+        if *value != new_value {
+            *value = new_value;
+            has_changed = true;
+        }
+    } else {
+        db.storage_mut().values().insert(*node_id, new_value);
+        has_changed = true;
+    }
+    let current_epoch = db.current_epoch();
+    if let Some(node) = db.storage_mut().nodes().get_mut(node_id) {
```
rbalicki2, line 124: why do we call `get_mut` twice here? Can we return a mutable reference from the above if block, or move the `node.dependencies = ...` into both branches of the if block?

crates/pico_macros/src/db.rs
```diff
@@ -0,0 +1,71 @@
...
+                    .downcast_ref::<T>()
+                    .expect("unexpected struct type. This is indicative of a bug in Pico.")
+                    .clone()
+            }
+
+            fn set<T>(&mut self, source: T) -> ::pico_core::source::SourceId<T>
+            where T: ::pico_core::source::Source + ::pico_core::dyn_eq::DynEq
+            {
+                use ::pico_core::{storage::StorageMut, container::Container};
+                let current_epoch = self.increment_epoch();
```
rbalicki2, line 52: I don't think we need to increment the epoch unless we modify an existing source.

That will mean that we can be more efficient in the following situation:

- call memoized function foo
- insert a new source
- call foo again

Now we can avoid iterating over foo's dependencies, since it was verified at the current epoch.

Thoughts? This could easily be a follow up as well

crates/pico/src/storage/default.rs
```diff
@@ -0,0 +1,45 @@
...
+    params::ParamId,
+};
+
+use crate::container::DefaultContainer;
+
+#[derive(Debug, Storage)]
+pub struct DefaultStorage<Db: Database> {
+    pub current_epoch: Epoch,
+    pub dependency_stack: Vec<Vec<(Epoch, Dependency)>>,
+    pub nodes: DefaultContainer<NodeId, DerivedNode<Db>>,
```
rbalicki2, line 20: nit: can this be `derived_nodes`?

crates/pico_core/src/node.rs
```diff
@@ -0,0 +1,54 @@
...
+use std::hash::Hash;
+
+use crate::database::Database;
+use crate::dyn_eq::DynEq;
+use crate::epoch::Epoch;
+use crate::key::Key;
+use crate::params::ParamId;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct NodeId {
```
rbalicki2, line 11: nit: can this be `DerivedNodeId`?

crates/pico_core/src/key.rs
```diff
@@ -0,0 +1,18 @@
+use std::hash::{DefaultHasher, Hash, Hasher};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Key(u64);
```
rbalicki2, line 4: can we rename this something more intuitive? I don't yet understand what it is doing. It's an id of a memoized function, ...right?

crates/pico/src/storage/default.rs
rbalicki2: And likewise, could `values` be `derived_values`?

crates/pico_core/src/dyn_eq.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl PartialEq<&Self> for Box<dyn DynEq> {
+    fn eq(&self, other: &&Self) -> bool {
+        <Self as PartialEq>::eq(self, *other)
+    }
+}
+
+impl core::fmt::Debug for dyn DynEq {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "DynEq {{ .. }}")
```
rbalicki2, line 43: This might sound annoying, but can we require Any + Debug? If we don't want that in general, can we require it in `cfg[test]`, and have this fallback debug impl in non-test?

It's really hard to know what's going on from 
```
            NodeId {
                key: Key(
                    17133476938307633808,
                ),
                param_id: ParamId(
                    6585809115041605177,
                ),
            }: DynEq { .. },
```

rbalicki2: this is surprisingly hard

crates/pico/tests/memoization.rs
```diff
@@ -0,0 +1,93 @@
...
+
+mod calc;
+
+static SUM_COUNTER: AtomicUsize = AtomicUsize::new(0);
+static EVAL_COUNTER: LazyLock<Mutex<HashMap<SourceId<Input>, usize>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+#[test]
+fn memoization() {
+    let mut state = TestDatabase::default();
```
rbalicki2, line 22: test lgtm

can we also add an analogous test for when we update `left` and give it a new value, and assert that therefore `sum` was recalculated?

crates/pico/src/storage/default.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub values: DefaultContainer<NodeId, Box<dyn DynEq>>,
+    pub sources: DefaultContainer<Key, SourceNode>,
```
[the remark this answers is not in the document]
ch1ffa: Probably. It doesn't have inner_fn so it's easier to do for source nodes

crates/pico/tests/return_reference.rs
```diff
@@ -0,0 +1,45 @@
...
+#[memo]
+pub fn parse_ast(db: &mut TestDatabase, id: SourceId<Input>) -> Result<Program> {
+    let source_text = db.get(id);
+    let mut lexer = Lexer::new(source_text.value);
+    let mut parser = Parser::new(&mut lexer)?;
+    parser.parse_program()
+}
+
+#[memo(reference)]
+pub fn evaluate_input(db: &mut TestDatabase, id: SourceId<Input>) -> i64 {
```
rbalicki2, line 42: afaict the difference between `#[memo]` and `#[memo(reference)]` is that the reference one doesn't clone. Can we return a reference to an object with a custom clone impl that increments a `AtomicUSize` and assert the clone didn't happen?

crates/pico_core/src/key.rs
rbalicki2: sgtm. Let's add that as a comment!

rbalicki2 (commented):
again, partial review. The tests really helped, thank you for those

crates/pico/tests/memoization.rs
ch1ffa: Maybe add to the same test as a third step?

rbalicki2: yeah sgtm

crates/compiler-playground/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
rbalicki2, line 1: oops — looks leftover and maybe unnecessary?

crates/pico/tests/memoization.rs
```diff
@@ -0,0 +1,93 @@
...
+    });
+
+    let right = state.set(Input {
+        key: "right",
+        value: "(2 + 2) * 2".to_string(),
+    });
+
+    let result = sum(&mut state, left, right);
+    assert_eq!(result, 14);
+    assert_eq!(state.storage().current_epoch(), 3.into());
```
rbalicki2, line 36: though I understand why `.current_epoch()` needs to be exposed on the `trait Storage` (because it's used by `trait Db`), it feels to me like an implementation detail from the perspective of the developer. Perhaps we should avoid making assertions about `current_epoch` here?

crates/pico/tests/memoization.rs
```diff
@@ -0,0 +1,93 @@
+use std::{
```
rbalicki2, line 1: I think it would be helpful to have some tests related to eviction. Namely, my understanding is that you are always allowed to evict derived node values from the cache and get the same results, with the only difference being that we call some of the internal functions more.

I don't think we need any special storage, we can probably just call `db.storage_mut().derived_nodes.remove(...)` or whatnot

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+    params::ParamId,
+    storage::{Storage, StorageMut},
+};
+
+pub enum MemoState {
+    Memoized,
+    Computed,
+}
+
+#[allow(clippy::map_entry)]
```
rbalicki2, line 17: what is this triggered on? Maybe I'm not seeing it but if I run `cargo clippy` or `cargo clippy -p pico`, it doesn't trigger

The only one I get is

```
warning: this `map_or` is redundant
  --> crates/pico_macros/src/source.rs:29:17
   |
29 | /                 attr.path()
30 | |                     .segments
31 | |                     .last()
32 | |                     .map_or(false, |segment| segment.ident == "key")
   | |____________________________________________________________________^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or
   = note: `#[warn(clippy::unnecessary_map_or)]` on by default
help: use is_some_and instead
   |
29 ~                 attr.path()
30 +                     .segments
31 +                     .last().is_some_and(|segment| segment.ident == "key")
   |
```

which is triggered on the latest stable rust, and thus our CI will complain about

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+pub fn memo<Db: Database>(
+    db: &mut Db,
+    node_id: DerivedNodeId,
+    inner_fn: fn(&mut Db, ParamId) -> Box<dyn DynEq>,
+) -> MemoState {
+    let mut state = MemoState::Memoized;
+    let current_epoch = db.current_epoch();
+    let time_calculated = if db.storage().derived_nodes().contains_key(&node_id) {
```
rbalicki2, line 25: nit: `let (time_calculated, state) = ...` might be cleaner

crates/compiler-playground/src/lib.rs
ch1ffa: I'm going to delete the compiler-playground crate anyway. I was needed for development only, having tests is enough now

crates/pico/src/memo.rs
ch1ffa: Oh, probably a leftover

rbalicki2: sgtm

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+        NodeKind::Derived(node_id),
+        time_calculated,
+        current_epoch,
+    );
+    state
+}
+
+fn any_dependency_changed<Db: Database>(
+    db: &mut Db,
+    node_id: &DerivedNodeId,
```
rbalicki2, line 87: `DerivedNodeId` implements Copy, we can just accept the param directly. (This is true for a all the param/key/id structs)

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+#[allow(clippy::map_entry)]
+pub fn memo<Db: Database>(
+    db: &mut Db,
+    node_id: DerivedNodeId,
+    inner_fn: fn(&mut Db, ParamId) -> Box<dyn DynEq>,
+) -> MemoState {
+    let mut state = MemoState::Memoized;
+    let current_epoch = db.current_epoch();
+    let time_calculated = if db.storage().derived_nodes().contains_key(&node_id) {
+        if any_dependency_changed(db, &node_id, current_epoch) {
```
rbalicki2, line 26: Can we restructure these lines to be like

```
match db.storage().derived_nodes().get(&node_id) {
  Some(x) => if any_dependency_changed(db, x, current_epoch)
}
```

So that we can avoid the `expect` in `any_dependency_changed`?

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+            time_calculated,
+            Dependency {
+                node_to: node,
+                time_verified_or_calculated: current_epoch,
+            },
+        ));
+    }
+}
+
+fn with_dependency_tracking<Db>(
```
rbalicki2, line 172: love this function

crates/pico_core/src/node.rs
```diff
@@ -0,0 +1,57 @@
...
+            .field("time_calculated", &self.time_calculated)
+            .field("dependencies", &self.dependencies)
+            .field("value", &self.value)
+            .finish()
+    }
+}
+
+#[derive(Debug)]
+pub struct SourceNode {
+    pub time_calculated: Epoch,
```
rbalicki2, line 43: Hm... I think this isn't ideal naming. Maybe `time_updated` or `time_set`? A source isn't really calculated.

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+        .into_iter()
+        .any(|(dep_node, time_verified)| match dep_node {
+            NodeKind::Source(key) => source_node_changed(db, &key, time_verified),
+            NodeKind::Derived(dep_node_id) => derived_node_changed(db, dep_node_id),
+        })
+}
+
+fn source_node_changed<Db: Database>(db: &Db, key: &Key, time_verified: Epoch) -> bool {
+    match db.storage().source_nodes().get(key) {
+        Some(source) => source.time_calculated > time_verified,
```
rbalicki2, line 116: nit: I think it would be clearer for the signature to be 

```
fn source_node_changed_since<Db: Database>(db: &Db, key: &Key, since: Epoch) -> bool {
```

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+        None => true,
```
rbalicki2, line 117: should we panic here? Or, even better, perhaps `source_nodes` should have a `.get` method that returns `T` instead of `Option<T>`, since it's always a bug to have a missing source? (Right?!)

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+}
+
+fn derived_node_changed<Db: Database>(db: &mut Db, node_id: DerivedNodeId) -> bool {
+    let inner_fn = if let Some(node) = db.storage().derived_nodes().get(&node_id) {
+        node.inner_fn
+    } else {
+        return true;
```
rbalicki2, line 125: can you comment on what this means? e.g. it means that this is the first time we're calculating the derived node (maybe?), or the derived node has been GC'd, so we don't have the value so we can't be sure it hasn't changed.

Or, is it always a bug if we're missing the inner fn?

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+    };
+    if !db.storage().params().contains_key(&node_id.param_id) {
```
rbalicki2, line 127: If we are returning true, we're going to call `memo(...)`, right? Should we call `memo` here and compare results?

In particular, if we have `expensive_calc` which depends on `foo`, and `foo` has been GC'ed, we can recalculate it, and see that it is the same, and thus avoid calling `expensive_calc`.

crates/pico_core/src/storage.rs
```diff
@@ -0,0 +1,25 @@
...
+    container::Container,
+    database::Database,
+    epoch::Epoch,
+    key::Key,
+    node::{Dependency, DerivedNode, DerivedNodeId, SourceNode},
+    params::ParamId,
+};
+
+pub trait Storage<Db: Database + ?Sized> {
+    fn derived_nodes(&self) -> &impl Container<DerivedNodeId, DerivedNode<Db>>;
```
rbalicki2, line 13: might be worth commenting that it is safe to GC these *as long as we're not currently calling a memoized function*

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+        |(mut deps, mut max_time_calculated), (time_calculated, dep)| {
+            deps.push(dep);
+            max_time_calculated = std::cmp::max(max_time_calculated, time_calculated);
+            (deps, max_time_calculated)
+        },
+    );
+    (value, dependencies, time_calculated)
+}
+
+pub fn register_dependency<Db: Database>(
```
rbalicki2, line 155: Maybe we can make the fn name clearer, e.g. `register_dependency_in_parent_memoized_fn`

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,188 @@
...
+) {
+    if let Some(dependencies) = db.storage_mut().dependency_stack().last_mut() {
+        dependencies.push((
+            time_calculated,
+            Dependency {
+                node_to: node,
+                time_verified_or_calculated: current_epoch,
+            },
+        ));
+    }
```
rbalicki2, line 169: can you add an `else { // do nothing. We just finished the outermost memoized function }` or something for clarity?

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,136 @@
...
+
+pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
+    let ItemFn {
+        sig,
+        vis,
+        block,
+        attrs,
+    } = parse_macro_input!(item as ItemFn);
+
+    let fn_hash = hash(&sig);
```
rbalicki2, line 28: Can we detect collisions? (I don't care right now, to be fair)

e.g. if we have a `lazy_static!( I_HAVE_BEEN_SET: false )` which we set to true when the memoized function is called initially. And the we should only store an `inner_fn` under that `fn_hash` when `I_HAVE_BEEN_SET` is false.

Probably worth not doing in v1, I don't think we'll personally ever encounter collisions

crates/pico_macros/src/memo.rs
```diff
@@ -0,0 +1,136 @@
...
+            .expect("value should exist. This is indicative of a bug in Pico.")
+            .value
+            .as_any()
+            .downcast_ref::<#return_type>()
+            .expect("unexpected return type. This is indicative of a bug in Pico.")
+    };
+
+    let mut new_sig = sig.clone();
+
+    if args_.reference && !default {
```
rbalicki2, line 82: I would expect that `#[memo(reference)] fn foo(db) {}` would return `&()` lol

rbalicki2 (commented):
I think I'm quite close to finishing this review!

Absolutely amazing work. Let's get this out!!!

crates/pico/src/memo.rs
ch1ffa: I'll keep a reference in the container trait though to mimic HashMap interface

crates/pico_core/src/node.rs
ch1ffa: Although I like `time_set`, but it might be better to rename `time_calculated` to `time_updated` so we can use it for derived_node as well

crates/pico_macros/src/memo.rs
ch1ffa: I believe we can find something smarter than just a fn body hash, kind of immutable map that we build at compile time. Just haven't figured out it yet.

crates/pico_macros/src/memo.rs
ch1ffa: haha, I thought it doesn't make sense, but maybe remove this check indeed and simplify the macro

crates/pico/src/memo.rs
ch1ffa: `db` is borrowed as mutable in `any_dependency_changed` but we've already borrowed it in `db.storage().derived_nodes().get(&node_id)`

crates/pico/src/memo.rs
ch1ffa: node_id here is a dependency, so we can't be sure it's not GC'd.

crates/pico/src/memo.rs
ch1ffa: if we remove source node intentionally we need a way to force parent derive_node to recalculate its `inner_fn` to remove this source node from dependencies

crates/pico/src/memo.rs
ch1ffa: wdym? We need to know `inner_fn` and `param_id` to call `memo` for a dependent node. If we don't have it in the storage, the only way to restore it is to call parent `inner_fn` (e.g. return true here)

crates/pico_core/src/source.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn get_key(&self) -> Key;
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SourceId<T> {
+    pub key: Key,
+    phantom: PhantomData<T>,
+}
+
+impl<T: Clone> Copy for SourceId<T> {}
```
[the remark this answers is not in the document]
ch1ffa: That's a weird stuff with derive Copy for PhantomData. It requires T to implement Copy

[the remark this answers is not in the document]
rbalicki2: interesting, derive(Copy) compiles for me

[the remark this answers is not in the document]
ch1ffa: Oh, nice. Maybe the latest version fixed it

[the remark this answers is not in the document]
ch1ffa: ah, ok, derive is not very smart in that case, I need to implement Clone manually as well and we can remove `T: Clone`

crates/pico/src/memo.rs
```diff
@@ -0,0 +1,189 @@
...
+    Computed,
+}
+
+pub fn memo<Db: Database>(
+    db: &mut Db,
+    node_id: DerivedNodeId,
+    inner_fn: fn(&mut Db, ParamId) -> Box<dyn DynEq>,
+) -> MemoState {
+    let current_epoch = db.current_epoch();
+    let (time_updated, state) = if db.storage().derived_nodes().contains_key(&node_id) {
```
rbalicki2, line 23: sorry if I've asked this before, but is this more idiomatically done via `match derived_nodes.entry(...)`? In particular, that might allow us to avoid the `expect` in `any_dependency_changed`.

Maybe it's the fact that we're calling `call_inner_fn` etc that prevents us from keeping a mutable reference to the item around...

crates/pico/src/memo.rs
rbalicki2: I see. Maybe it's clearer if the fn is named `derived_node_must_be_recalculated`?

crates/pico/src/memo.rs
rbalicki2: What I mean is this... and I could be wrong...

- `expensively_calculate_full_name` is called a second time. It is a memoized function, so we call `memo`
- `any_dependency_changed` is called
- we call `derived_node_changed` for `first_name`. `db.storage().params().contains_key(&node_id.param_id)` is false, so we `return true` here
- so `any_dependency_changed` is true and we recalculate `expensively_calculate_full_name`

*but* instead, we could:

- we call `derived_node_changed` for `first_name`. `db.storage().params().contains_key(&node_id.param_id)` is false, so we call
```
let state = memo(db, node_id, inner_fn);
matches!(state, MemoState::Computed)
```
- it is not computed, i.e. it had the same value as last time, so we don't have to recalculate `expensively_calculate_full_name`.

I'm honestly not sure if that's possible... does the fact that the param ID is missing mean that we'll always return `MemoState::Computed`? (I don't think so...???)

rbalicki2 (commented):
a few minor questions about `memo`. Now... I only have to review the proc macro 😭  so close

crates/pico/src/memo.rs
ch1ffa: yeah, the problem is we can't hold reference to node and use `&mut db` for `call_inner_fn` the same time

crates/pico/src/memo.rs
ch1ffa: We are in the `any_dependency_changed` fn, we check what dependencies must be verified and then for every found dependency we check if `this` dependency is changed or not. I'm worried if `derived_node_must_be_recalculated` fn with arguments `db` and `node_id` sounds like its own node must be recalculated, but it rather return a `bool` that indicates that a parent node must be recalculated (we've already done recalculation of dependent node inside `derived_node_changed` fn)

crates/pico/src/memo.rs
ch1ffa: afaict `db.storage().params().contains_key(&node_id.param_id)` should not be normally false inside dependency check. The only way it could be missing is being GC'd, because we saved params of a child memoized fn during previous call. 

Without param_id check here we will get a panic inside `inner_fn` during param lookup. We assume that params were interned right before `memo` call, so there is no way it's gonna be GC'd so fast. But if it was GC'd some time ago and there is no such param_id anymore,  let's just call parent `memo` and put it to the storage again.

crates/pico/tests/remove_node.rs
```diff
@@ -0,0 +1,128 @@
...
+use pico_macros::{memo, Db, Source};
+
+mod calc;
+
+static SUM_COUNTER: AtomicUsize = AtomicUsize::new(0);
+static EVAL_COUNTER: LazyLock<Mutex<HashMap<SourceId<Input>, usize>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+#[test]
+fn memoization() {
```
rbalicki2, line 21: nit: `remove_node`

crates/pico/tests/remove_node.rs
```diff
@@ -0,0 +1,128 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Source)]
+struct Input {
+    #[key]
+    pub key: &'static str,
+    pub value: String,
+}
+
+#[memo]
+pub fn parse_ast(db: &mut TestDatabase, id: SourceId<Input>) -> Result<Program> {
```
rbalicki2, line 108: in your tests, you have a bunch of pub fn's that don't need to be pub

crates/pico/tests/shared_dependency.rs
```diff
@@ -0,0 +1,67 @@
...
+#[test]
+fn shared_dependency() {
+    let mut state = TestDatabase::default();
+
+    let input = state.set(Input {
+        key: "input",
+        value: "2 + 2 * 2".to_string(),
+    });
+
+    let result = evaluate(&mut state, input);
```
rbalicki2, line 19: can we add the same `SUM_COUNTER`-type stuff to assert that we're calling the functions the expected number of times?

crates/pico/tests/remove_node.rs
```diff
@@ -0,0 +1,128 @@
...
+        .collect();
+    // derived_node should be removed
+    assert_ne!(
+        derived_nodes_count,
+        state.storage.derived_nodes.iter().count()
+    );
+    let result = sum(&mut state, left, right);
+    assert_eq!(result, 14);
+    // removed derived_node should be restored
+    assert_eq!(
```
rbalicki2, line 53: can we add an assertion that we `SUM_COUNTER` has gone up?

crates/pico/tests/remove_node.rs
```diff
@@ -0,0 +1,128 @@
...
+
+use calc::{ast::Program, error::Result, eval::eval, lexer::Lexer, parser::Parser};
+use pico::storage::DefaultStorage;
+use pico_core::{container::Container, database::Database, dyn_eq::DynEq, source::SourceId};
+use pico_macros::{memo, Db, Source};
+
+mod calc;
+
+static SUM_COUNTER: AtomicUsize = AtomicUsize::new(0);
+static EVAL_COUNTER: LazyLock<Mutex<HashMap<SourceId<Input>, usize>>> =
```
rbalicki2, line 17: let's either add assertions about these (🤩) or remove these statics (😢)

crates/pico/tests/return_reference.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    let value_ref = evaluate_input_ref(&mut state, input);
+    assert_eq!(*value_ref, Value(6));
+    // assert that the value was not cloned
+    assert_eq!(COUNTER.load(Ordering::SeqCst), 0);
+
+    let value = evaluate_input(&mut state, input);
+    assert_eq!(value, Value(6));
+    // assert that the value was cloned this time
+    assert_eq!(COUNTER.load(Ordering::SeqCst), 1);
```
rbalicki2, line 29: nice

rbalicki2 (commented):
a few nits about tests

crates/pico_core/src/storage.rs
```diff
@@ -0,0 +1,25 @@
...
+    fn source_nodes(&self) -> &impl Container<Key, SourceNode>;
+    fn params(&self) -> &impl Container<ParamId, Box<dyn Any>>;
+    fn current_epoch(&self) -> Epoch;
+}
+
+pub trait StorageMut<Db: Database + ?Sized> {
```
rbalicki2, line 19: I think this should require `Storage` and have differently named methods (`derived_nodes_mut` etc)

PR #294, opened 2024-12-25
Fix `isograph-config-schema` being different in ci

PatrykWalach:


PatrykWalach:
The `isograph-config-schema.json` file generated on windows and linux differ in `git status` it doesn't seem to be bacause of permissions or line endings, the file from linux is displayed as binary in `github` 😕

PatrykWalach:
The windows is using `text/plain; charset=utf-16le` but  linux is using `application/json; charset=us-ascii` based on `file -bi libs/isograph-compiler/isograph-config-schema.json` linux command

rbalicki2:
whoa, very cool detective work

PR #296, opened 2024-12-26
cache `rust build` with `turbo`

### LINKED ISSUE #19: Re-use cached artifacts from other builds

PatrykWalach:
Closes #19
Supersedes #261


turbo.json
```diff
@@ -0,0 +1,13 @@
+{
+  "$schema": "https://turbo.build/schema.json",
+  "tasks": {
+    "//#build": {
```
rbalicki2, line 4: Is there a reason these tasks are named "//..." but the task in https://github.com/isographlabs/isograph/pull/258/files is named "compile"?

rbalicki2 (approved):
lgtm shipping

turbo.json
PatrykWalach: Name without # refers to script in every package, for single package it can be vite-demo#build. // Is special alias for root package.

PR #271, opened 2024-11-13
add `NormalizationAstLoader`

PatrykWalach:
#297 should be merged first

Part of #33, #43

Issue #43, opened 2024-04-08
[epic] Deferred normalization ASTs 


rbalicki2:
* Depends on #33

## Background: entrypoint artifacts and normalization ASTs

* For each `` iso(`entrypoint Type.Field`) `` literal that the compiler encounters, it will generate a `Type/Field/entrypoint.ts` file.
* This file currently (i.e. pre #33) contains a query text and a normalization AST.
* The query text is used when making the query
* The normalization AST is (currently) only used to write the network response into the key-value store and for garbage collection (which is also only done after the network response is received).
     * They will (after #42) be used to determine whether a network request needs to be made.
* Because the normalization AST is only used after a network response is received, we can fetch the normalization AST later (e.g. when making the network request).

## This epic

* This epic is to load normalization ASTs when the network request is created, via https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import or something.
* This should be controllable with a directive on the entrypoint
    * So, the iso literal parser should be modified to include support for parsing directives
* Whether the normalization AST is hard-loaded should be reflected in the type of the entrypoint, since the entrypoint **cannot** be used with an `if-necessary` fetch policy if the normalization AST is not present.
    * One should be able to eventually call `entrypoint.loadNormalizationAst().then(entrypointWithNormalizationAst => fetchIfNecessary(...))` or the like

rbalicki2:
I think the remaining tasks are:
- [x] move normalization ASTs into their own files
- [x] add support for a @lazyLoadAst directive (or something) to entrypoint literals
- [ ] replacing the hard require of the normalization AST file with `() => import('./normalizationAst')` when that directive is encountered

XiNiHa:
I'd like to work on this one!

Issue #33, opened 2024-03-20
Split `entrypoint.ts` artifact into two artifacts: query text and normalization AST


rbalicki2:
## Problem and solution

* An entrypoint is two things: query text and normalization AST
* We do not need the normalization AST to make the network request, just the query text
* We should split this into two files, with the entrypoint (containing the query text) importing the file with the normalization AST.
* The files should be named `Type/field/entrypoint.ts` (containing the query text) and `/normalization.ts`, containing the normalization AST.

## Follow up:

This unblocks the following:
* This will make it possible to asynchronously load the normalization AST, for example right after the query text is loaded, when the network request is made or when the network response is received
* We do not want to asynchronously load the normalization AST, as not having the normalization AST means that we cannot determine whether the network request can be completed from the store, so there is a legitimate use case for bundling the normalization AST (i.e. with a regular import) with the query text

ismajl-ramadani:
Can I work on this as well?

rbalicki2:
Closed as duplicate #394

PR #297, opened 2024-12-26
Add kind to normalization ast

PatrykWalach:
This converts `NormalizationAst` from `NormalizationAstNode[]` to object with `kind: "NormalizationAst"` so we can make it a union of `NormalizationAstLoader` and `NormalizationAst` in #271 

Part of #33, #43


rbalicki2 (approved):
LGTM

PR #305, opened 2025-01-04
Fix components not refreshing during hmr

PatrykWalach:
Fixes HMR by method described in #264
Follow up to #264


PatrykWalach:
To fix hmr for no babel user we don't have to use `WeakMap` we have to store `resolver` next to the wrapped component so we can compare if `resolver` has changed.

PR #306, opened 2025-01-06
feat: Wrap components in HOC to support hmr 

PatrykWalach:
This PR makes babel/swc transform change `iso(`..`)(Field)` into `hmr(Field)`.
The `hmr` is a HOC which transforms the resolver to accept a single argument `(firstArgument, additionalProps) => ...` => `({ firstArgument, additionalProps }) => ...`. 
This makes react HMR work in all modes: with swc, babel or no transform.

Hast to be merged locally with #308 to test.

Fixes HMR by method described in #264
Follow up to #264
Rebase of #305


PatrykWalach:
We can consider importing different function from `@isograph/react` instead of changing `iso` definition, this way we can still error if babel is not setup correctly.

PatrykWalach:
Ideally we should support the new [Activity](https://react.dev/blog/2025/04/23/react-labs-view-transitions-activity-and-more#activity) component

PatrykWalach:
I had an idea that I need to think over but instead of introducing arbitrary timeout to dispose, maybe we can dispose in the nextTick/nextAnimationFrame or something like that

rbalicki2:
The activity component (which I believe was called the offscreen API at some point in time in the past) is going to be tough, because if a component goes off and onscreen, React will:

- call the effect disposal function when the component turns invisible, then
- "restore the state" when the component becomes visible again. (Perhaps it will also call the effect, I don't remember.)

But for something like `useLazyReference`, that would result in the state containing a fragment reference that is disposed. I guess the usual thing to do is to refetch in a situation like this, and we should probably just do that.

The additional problem is that you might invoke a mutation in `useLazyReference`. (Should you? I don't think it's necessarily a bad idea, if you can't avoid fetching during render. You might imagine that navigating to https://docs.new creates a new document during render. Or, you might imagine a modal that opens idiopathically, and that modal executes a mutation during initial render.) So, we should either:

- discourage that pattern and let folks be aware that that's the risk, or
- find some way to force folks to store that state somewhere else (e.g. tag some fields as safely refetchable?)

But yeah!

PR #308, opened 2025-01-06
[FIX] error occurring in `useEffect` on `hmr`/`strict mode`

PatrykWalach:
This PR fixes runtime errors thrown during hmr/strict mode.

This on it's own is not enough to make hmr work. Hast to be merged locally with #306 to test hmr.

Rebase of #264

### DIFF
--- a/demos/pet-demo/next.config.js
+++ b/demos/pet-demo/next.config.js
@@ -2,6 +2,7 @@ const path = require('path');
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  reactStrictMode: true,
   experimental: {
     swcPlugins: [
       [
--- a/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
+++ b/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
@@ -74,6 +74,7 @@ export function useCachedResponsivePrecommitValue<T>(
   }, []);
 
   useEffect(() => {
+    lastCommittedParentCache.current = parentCache;
     // On commit, cacheItem may be disposed, because during the render phase,
     // we only temporarily retained the item, and the temporary retain could have
     // expired by the time of the commit.
--- a/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
+++ b/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
@@ -64,16 +64,6 @@ export function useCachedResponsivePrecommitValue<T>(
   }, []);
 
   useEffect(() => {
-    return () => {
-      // Attempt to detect if the component was
-      // hidden (by Offscreen API), or fast refresh occured;
-      // Only in these situations would the effect cleanup
-      // for "unmounting" run multiple times.
-      lastCommittedParentCache.current = null;
-    };
-  }, []);
-
-  useEffect(() => {
     lastCommittedParentCache.current = parentCache;
     // On commit, cacheItem may be disposed, because during the render phase,
     // we only temporarily retained the item, and the temporary retain could have
--- a/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
+++ b/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
@@ -64,15 +64,16 @@ export function useCachedResponsivePrecommitValue<T>(
   }, []);
 
   useEffect(() => {
-    // react reruns all `useEffect` in HMR since it doesn't know if the
-    // code inside of useEffect has changed. Since this is a library
-    // user can't change this code so we are safe to skip this rerun.
-    // This also prevents `useEffect` from running twice in Strict Mode.
-    if (lastCommittedParentCache.current === parentCache) {
-      return;
-    }
+    return () => {
+      // Attempt to detect if the component was
+      // hidden (by Offscreen API), or fast refresh occured;
+      // Only in these situations would the effect cleanup
+      // for "unmounting" run multiple times.
+      lastCommittedParentCache.current = null;
+    };
+  }, []);
 
-    lastCommittedParentCache.current = parentCache;
+  useEffect(() => {
     // On commit, cacheItem may be disposed, because during the render phase,
     // we only temporarily retained the item, and the temporary retain could have
     // expired by the time of the commit.
--- a/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
+++ b/libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
@@ -50,6 +50,10 @@ export function useCachedResponsivePrecommitValue<T>(
   const lastCommittedParentCache = useRef<ParentCache<T> | null>(null);
 
   useEffect(() => {
+    // react reruns all `useEffect` in HMR since it doesn't know if the
+    // code inside of useEffect has changed. Since this is a library
+    // user can't change this code so we are safe to skip this rerun.
+    // This also prevents `useEffect` from running twice in Strict Mode.
     if (lastCommittedParentCache.current === parentCache) {
       return;
     }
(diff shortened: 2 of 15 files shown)


rbalicki2 (approved):
lgtm, just rewrote the comment slightly for grammar and clarity

libs/isograph-react-disposable-state/src/useCachedResponsivePrecommitValue.ts
```diff
@@ -51,6 +51,14 @@ export function useCachedResponsivePrecommitValue<T>(
   const lastCommittedParentCache = useRef<ParentCache<T> | null>(null);
 
   useEffect(() => {
+    // react reruns all `useEffect` in HMR since it doesn't know if the
```
rbalicki2, line 54:
```
    // Since React doesn't know if the code inside of a useEffect has changed,
    // it unmounts and remounts all useEffect calls when doing HMR. This breaks
    // this hook's assumptions! The `onCommit` is called twice.
    //
    // Since this is a library, the user can't change this code, so we are safe
    // to skip this rerun.
    //
    // This also prevents the useEffect from running twice in Strict Mode.
```

libs/isograph-react-disposable-state/src/useLazyDisposableState.ts
```diff
@@ -23,16 +23,24 @@ export function useLazyDisposableState<T>(
...
     parentCache,
     (pair) => {
-      itemCleanupPairRef.current?.[1]();
       itemCleanupPairRef.current = pair;
     },
   );
 
+  const lastCommittedParentCache = useRef<ParentCache<T> | null>(null);
   useEffect(() => {
+    // react reruns all `useEffect` in HMR since it doesn't know if the
```
rbalicki2, line 35:
```
    // Since React doesn't know if the code inside of a useEffect has changed,
    // it unmounts and remounts all useEffect calls when doing HMR. This breaks
    // this hook's assumptions! The cleanupFn is called unexpectedly during HMR,
    // and the item is not recreated. (Even recreating the item would be a bad outcome,
    // as that would reissue network requests.)
    //
    // Since this is a library, the user can't change this code, so we are safe
    // to skip this rerun.
    //
    // This also prevents the useEffect from running twice in Strict Mode.
```

libs/isograph-react-disposable-state/src/useDisposableState.ts
```diff
@@ -47,14 +46,23 @@ export function useDisposableState<T = never>(
...
-  useEffect(function cleanupItemCleanupPairRefIfSetStateNotCalled() {
-    return () => {
-      if (itemCleanupPairRef.current !== null) {
-        itemCleanupPairRef.current[1]();
-        itemCleanupPairRef.current = null;
+  const lastCommittedParentCache = useRef<ParentCache<T> | null>(null);
+
+  useEffect(
+    function cleanupItemCleanupPairRefIfSetStateNotCalled() {
+      if (lastCommittedParentCache.current === parentCache) {
```
rbalicki2, line 53:
```
    // Since React doesn't know if the code inside of a useEffect has changed,
    // it unmounts and remounts all useEffect calls when doing HMR. This breaks
    // this hook's assumptions! The item in the `itemCleanupPairRef` will be disposed.
    //
    // Since this is a library, the user can't change this code, so we are safe
    // to skip this rerun.
    //
    // This also prevents the useEffect from running twice in Strict Mode.
```

Do we also need to have this check above, in the other `useEffect`?

rbalicki2 (changes requested):
Sorry, I forgot to do a thorough review before accepting 😂  Do we also need this in `useUpdatableDisposableState`?

If so, there are probably enough places where we do this that it makes sense to have a `useEffectIgnoringHmrAndStrictMode` hook.

It might return the `lastCommittedParentCache`, in which case we don't need to do anything extra in `useCachedResponsivePreCommitValue`

libs/isograph-react-disposable-state/src/useDisposableState.ts
PatrykWalach: The second one will not throw because it sets the ref to null. However, because of this, we will lose the current state, which is not perfect.

rbalicki2: yeah, seems better to not set it to null!

PatrykWalach: Oh wait, this should be good, because it will cleanup only when we set with a function and when the ref is still not cleaned up, hmr shouldn't be able to trigger it twice

rbalicki2: wdym? Which useEffect are you referring to?

PatrykWalach:
```tsx
useEffect(
    function cleanupItemCleanupPairRefAfterSetState() {
      if (stateFromDisposableStateHook !== UNASSIGNED_STATE) {
        if (itemCleanupPairRef.current !== null) {
          itemCleanupPairRef.current[1]();
          itemCleanupPairRef.current = null;
        } else {
          throw new Error(
            'itemCleanupPairRef.current is unexpectedly null. ' +
              'This indicates a bug in react-disposable-state.',
          );
        }
      }
    },
    [stateFromDisposableStateHook],
  );
```
This one should cleanup only when these two conditions are met, reruning this on hmr has no effect.

PatrykWalach:
The solution used here is not good
- on the first render `useEffect` will `commit` and return `cleanup`
- on `hmr` it will `cleanup`, then the guard will prevent another `commit` and `cleanup` return 

so we end up in a cleaned-up state

The reason it doesn't cause any issues now is probably because of `gcBuffer`.

rbalicki2:
Ah yeah, you're totally right. Good call re: gcBuffer. I'm going to think about this. I wonder if we need to try to detect HMR/double mounting more directly.

PatrykWalach:
Now the cycle is 
- commit -> cache is populated
- cleanup -> set maybe `maybeHiddenOrFastRefresh` to `true`, cache is cleaned up
if component is unmounted, that's the end, if it's hmr:
- commit -> when populating cache we read `maybeHiddenOrFastRefresh` and skip `cacheItem.permanentRetainIfNotDisposed` so there's no error

so at the end we are in the populated state.


we don't need to set `itemCleanupPairRef` to `null` because `cleanup` runs just before cache is populated which will override `itemCleanupPairRef` and there are no renders in between.


`useUpdatableDisposableState` still needs to be fixed and it can't be fixed in the same way, as there's no way to repopulate 🤔

libs/isograph-react-disposable-state/src/useUpdatableDisposableState.ts
```diff
@@ -109,8 +109,10 @@ export function useUpdatableDisposableState<
   useEffect(() => {
     return function disposeAllRemainingItems() {
       for (const undisposedICI of undisposedICIs.current) {
+        undisposedICIs.current.delete(undisposedICI);
         undisposedICI.cleanup();
       }
+      setStateICI(UNASSIGNED_STATE);
```
PatrykWalach, line 115: I think it's fair to set this to `UNASSIGNED_STATE` to prevent mutations from firing twice. This was the last piece preventing this PR from being merged @rbalicki2 let me know what you think

demos/pet-demo/next.config.js
```diff
@@ -2,6 +2,7 @@ const path = require('path');
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  reactStrictMode: true,
```
PatrykWalach, line 5: Not sure if strict mode is working, setting `gcBufferSize = 0` doesn't unretain the query on initial render, but it works as expected during hmr so it should be fine

PatrykWalach: This works as expected.

`StrictMode` runs render twice, discard first render and runs useEffect for second render. So we create two temporary retains, then first one is not destroyed in useEffect, we have to wait for timeout.

libs/isograph-react-disposable-state/src/useUpdatableDisposableState.ts
rbalicki2: wdym prevent mutations from firing twice? maybe prevent mutations from being cleaned up twice

this makes sense to me, but I don't understand your comment

rbalicki2: lol >9 months later I notice this comment

Issue #312, opened 2025-01-12
Label fields as updatable


rbalicki2:
Given 
```
iso(`field Type.name { name @updatable }`)(({ data }) => {})
```
it would be great if we could update `data.name` somehow.

Needed:

- a way to batch updates, so e.g. data.name = foo cannot be the interface
- generated types for the updatable object

So e.g. a good interface might be:

```
({ data, startUpdate }) => {
  startUpdate((updatableData) => {
    updatableData.name = 'Mr. Isograph'
  })
});
```

PR #314, opened 2025-01-15
add `startUpdate` to `TReadFromStore`

PatrykWalach:
This will let use infer type of `startUpdate` inside of `resolver`.
The type will be generated in coresponding `param_type.ts` file.

Part of #312

### DIFF
--- a/libs/isograph-react/src/core/componentCache.ts
+++ b/libs/isograph-react/src/core/componentCache.ts
@@ -51,6 +51,7 @@ export function getOrCreateCachedComponent(
           {
             data,
             parameters: fragmentReference.variables,
+            startUpdate: () => {},
           },
           additionalRuntimeProps,
         );
--- a/libs/isograph-react/src/core/componentCache.ts
+++ b/libs/isograph-react/src/core/componentCache.ts
@@ -5,6 +5,7 @@ import { useReadAndSubscribe } from '../react/useReadAndSubscribe';
 import { NetworkRequestReaderOptions } from './read';
 import { readPromise } from './PromiseWrapper';
 import { logMessage } from './logging';
+import { startUpdate } from './startUpdate';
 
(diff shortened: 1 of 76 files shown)


PatrykWalach:
Should we keep `startUpdate` in client pointers?

libs/isograph-react/src/core/componentCache.ts
```diff
@@ -51,6 +51,7 @@ export function getOrCreateCachedComponent(
           {
             data,
             parameters: fragmentReference.variables,
+            startUpdate: () => {},
```
rbalicki2, line 54: `startUpdate` is optional. I think it should not be provided if you have no updatable fields, right? If that's the case, maybe what we need is:

- an `updatableReader: Reader | null` field, *or* `updatable: boolean; fetchable: boolean` on individual nodes.
- based on this, we can determine whether to pass a `startUpdate` function or null, and this will match the type in param_type
So basically, I think a good default is to have `startUpdate` be missing from the object here (and elsewhere)

Thoughts?

rbalicki2 (changes requested):
LGTM overall, just a minor nit

libs/isograph-react/src/core/componentCache.ts
PatrykWalach: I don't think this can cause any errors, but providing this based on a boolean in ast shouldn't be a problem

rbalicki2:
I'm a little torn about the `hasUpdatable: bool` flag. I'm inclined to have a single source of truth: the reader AST. But this is probably a good optimization, since:
- we may have some null/missing fields, so we may not have read the entire reader AST
- so we (usually) can't conclusively say a reader fragment is not updatable
- but most fragments won't have any updatable fields
- and it seems wasteful to create an updatable data proxy every time we read, just to most often throw it away

PR #320, opened 2025-01-21
make raw network response error handling graphql spec compliant in demos

### LINKED ISSUE #318: Throw on deviations from spec: if errors array is empty
If json.errors.length == 0, it should not throw. But we do!

adrtivv:
Fixes #318

### DIFF
--- a/demos/vite-demo/src/App.tsx
+++ b/demos/vite-demo/src/App.tsx
@@ -20,17 +20,24 @@ function makeNetworkRequest<T>(
     const json = await response.json();
 
     if (response.ok) {
-      if (json.errors != null) {
+      /**
+       * Enforce that the network response follows the specification:: {@link https://spec.graphql.org/draft/#sec-Errors}.
+       */
+      if (Object.hasOwn(json, 'errors')) {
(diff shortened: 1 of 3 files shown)


demos/vite-demo/src/App.tsx
```diff
@@ -20,17 +20,24 @@ function makeNetworkRequest<T>(
     const json = await response.json();
 
     if (response.ok) {
-      if (json.errors != null) {
+      /**
+       * The `errors` field in a graphql response must follow these rules mentioned in the graphQL specification: {@link https://spec.graphql.org/draft/#sec-Errors}.
```
rbalicki2, line 24: Nit: The link is useful, but IMO the code is somewhat self-explanatory. Can we change the comment to "enforce that the network response follows the specification: ..."

rbalicki2 (changes requested):
Amazing! Thank you for doing this. Just a small nit

demos/vite-demo/src/App.tsx
adrtivv: Okay

adrtivv:
done

rbalicki2 (approved):
Shipittt

PR #323, opened 2025-01-21
refactor iteration in argument validation 

PatrykWalach:
Refactor follow up to #291


rbalicki2 (approved):
Yesss, so clean. I love it. Once it's mergeable, let's land it

PR #325, opened 2025-01-23
Generate artifacts for `@updatable` scalars

PatrykWalach:
Follow up to #314
Part of updatable fields #312


crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -436,37 +438,70 @@ pub(crate) fn generate_client_field_parameter_type(
...
+        .push_str(&format!("{}}}", "  ".repeat(indentation_level as usize)));
+
+    (
+        ClientFieldParameterType(client_field_parameter_type),
+        ClientFieldUpdatableDataType(client_field_updatable_data_type),
+    )
+}
 
-    ClientFieldParameterType(client_field_parameter_type)
+pub struct DualStringProxy<'a> {
```
rbalicki2, line 470: nice, I like this a lot. Let's add a comment to say why we're using this.

However, it's a bit unfortunate that we have to sometimes write through `DualStringProxy`, and sometimes let `DualStringProxy` drop and write directly to `readonly_data_type` and `updatable_data_type`.

-------

Maybe we should pass a single object around, `query_type_declaration: RegularAndUpdatableQueryTypeDeclaration`, which contains methods:

- write_scalar_field(name, value, is_updatable)
- write_linked_field(name, value, subfields, is_updatable)

or perhaps

- write_linked_field_start(name, value, is_updatable)
- close_linked_field

the object could keep track of the indentation

IMO this gets us closer to what we probably should be doing, which is creating TypeScript ASTs, instead of concatenating a bunch of strings.

-----

Let's make this change ☝️ as a separate PR that has no concept of updatable fields (i.e. just create a `QueryTypeDeclaration` object)

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -500,18 +535,40 @@ fn write_param_type_from_selection(
...
+                            ));
+                            readonly_data_type.push_str(&format!(
+                                "{}readonly {}: {},\n",
+                                "  ".repeat(indentation_level as usize),
+                                name_or_alias,
+                                print_javascript_type_declaration(&output_type)
+                            ));
+                        }
+                        ValidatedIsographSelectionVariant::Loadable(_) => {
+                            panic!("@loadble server fields are not supported")
```
rbalicki2, line 555: nit: typo

rbalicki2 (changes requested):
some nits, then I'll re-review

crates/graphql_artifact_generation/src/generate_artifacts.rs
```diff
@@ -597,41 +658,56 @@ fn write_param_type_from_selection(
...
+
+                    (
+                        associated_data
+                            .type_name
+                            .clone()
+                            .map(&mut |_| parameter_type.0.clone()),
+                        associated_data
+                            .type_name
+                            .clone()
+                            .map(&mut |_| updatable_data_type.0.clone()),
```
PatrykWalach, line 696: I also don't like recursively cloning two Strings here `parameter_type` and `updatable_data_type`

crates/graphql_artifact_generation/src/generate_artifacts.rs
rbalicki2: lets do this as a separate issue

crates/graphql_artifact_generation/src/generate_artifacts.rs
rbalicki2: yeah, I agree, it would be better to avoid this. I'll think about it (but still merge)

rbalicki2 (approved):

rbalicki2 (approved):
NIIIIICE shipitttt

rbalicki2:
once the merge conflicts are fixed, lets ship this

PR #326, opened 2025-01-23
use `cross-env` to call `isograph_cli` binaries

PatrykWalach:
Using `cross-env` makes it not required to replace paths on windows (`../../target/debug/isograph_cli` with `..\\..\\target\\debug\\isograph_cli`)

also makes env variables work `ISO_PRINT_ABSOLUTE_FILEPATH=1` on windows


rbalicki2:
was `pnpm build-demos` broken on windows?!

PatrykWalach:
Yes, because the paths are escaped differently in the terminal
running `../../target/debug/isograph_cli` in terminal works
but doing so through `pnpm run` does not 😖 

So I was changing these paths locally, and was sometimes commiting these by mistake

rbalicki2:
sweet, glad we caught it then

PR #330, opened 2025-01-24
runtime import `entrypoints` with `no_babel_transform: true`

### LINKED ISSUE #333: Runtime returns the actual iso content if no babel transform

PatrykWalach:
Randall requested this so I've done a quick prototype
Closes #333

### DIFF
--- a/crates/graphql_artifact_generation/src/iso_overload_file.rs
+++ b/crates/graphql_artifact_generation/src/iso_overload_file.rs
@@ -8,7 +8,7 @@ use isograph_schema::{
     ValidatedSchema,
 };
 
-use crate::generate_artifacts::ISO_TS;
+use crate::generate_artifacts::ISO_TS_FILE_NAME;
 
 fn build_iso_overload_for_entrypoint(
     validated_client_field: &ValidatedClientField,
@@ -158,7 +158,7 @@ type MatchesWhitespaceAndString<
 export function iso(_isographLiteralText: string):
   | IdentityWithParam<any>
   | IdentityWithParamComponent<any>
-  | IsographEntrypoint<any, any>
+  | IsographEntrypoint<any, any, any>
 {\n",
             );
             content.push_str("  throw new Error('iso: Unexpected invocation at runtime. Either the Babel transform ' +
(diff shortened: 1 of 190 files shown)


crates/graphql_artifact_generation/src/iso_overload_file.rs
```diff
@@ -158,39 +168,66 @@ type MatchesWhitespaceAndString<
...
         }
         true => {
-            "  return (clientFieldResolver: any) => clientFieldResolver;"
+            let switch_cases =
+                sorted_entrypoints(schema)
+                    .into_iter()
+                    .map(|(field, iso_literal_text)| {
+                        format!(
+                            "    case '{}':
+      return import('./{}/{}/entrypoint{}').then(module => module.default);\n",
```
rbalicki2, line 203: Let's import everything and return the imported entrypoint directly.

If you don't use babel, and your bundler doesn't prune this well, you can (eventually) have an entrypoint where the query text is an ID, the normalization AST is a loader and the reader AST is a loader

If Randall needs the better perf, let's explore how to add that

crates/graphql_artifact_generation/src/iso_overload_file.rs
```diff
@@ -158,39 +168,66 @@ type MatchesWhitespaceAndString<
...
+                "
+export function iso(isographLiteralText: string):
+  | IdentityWithParam<any>
+  | IdentityWithParamComponent<any>
+  | Promise<IsographEntrypoint<any, any>>
+{
+  switch (isographLiteralText) {\n",
+            );
+
+            for switch_case in switch_cases {
```
rbalicki2, line 221: once we have directives, this won't quite be right...
In particular, we can have an `entrypoint Query.HomePage @lazy` or something in one file, and a `entrypoint Query.HomePage` in another.

Anyway, no big deal, we can cross that bridge later (since that will mean we probably can't have a single `entrypoint.ts` file anymore

rbalicki2 (changes requested):
Should be good once we do the non-async import

rbalicki2 (approved):

Issue #338, opened 2025-01-26
useResult overloads to accept null and return null


rbalicki2:
- This is a typescript only change
- useResult (which may be renamed) accepts a non-nullable fragment reference, suspends, and returns the result.
- We should add an overload (at the type level) such that if it receives null, it returns null
- **I do not want an overload where if it receives a nullable fragment reference, it returns a nullable response**. Unless, instead, receive null -> return null. These should have identical behavior, but null -> null is conceptually cleaner. If you come across some reason while implementing this why it should be different, then go for it.
- In the implementation, the tricky thing is that we still need to call all the hooks, even if we are passed null :/ the rules of hooks are cruel. So, this may have some subtasks:
    - make useReadAndSubscribe accept a nullable fragment reference
    - make useRerenderOnChange accept a nullable fragment reference
    - etc
- The key thing is that this has to work if we pass null, then a fragment reference, then null, then a fragment reference, etc. and it should work whether or not there is an intervening commit.

adrtivv:
```ts
type OptionalValue<T> = T | undefined;
type NullableValue<T> = T | null;
type NullishValue<T> = T | null | undefined;
```
I think in conversations it's better to differentiate between the possible unions of `null` and `undefined` any any other javscript data type using the following terms(just how like [zod does it](https://github.com/colinhacks/zod?tab=readme-ov-file#optional)).

Anyway, is there really a need for handling both `null` and `undefined` states? Is the fragment reference from `useLazyReference` going to exist in both of those states in the future?

> - In the implementation, the tricky thing is that we still need to call all the hooks, even if > we are passed null :/ the rules of hooks are cruel. So, this may have some subtasks:
>   - make useReadAndSubscribe accept a nullable fragment reference
>   - make useRerenderOnChange accept a nullable fragment reference
>   - etc

I am assuming that you meant to say that since react hooks cannot be called conditionally, the `null` and `undefined` states have to be handled internally by the react hooks used within `useResult` hook.

PR #342, opened 2025-01-28
draft

adrtivv:
addresses #338

@rbalicki2 `useLazyReference` hook never returns a nullish fragment reference. And even if was nullish I cannot see how handling a nullish fragment reference in the `useResult` hook is useful in any way other than what I've implemented in this pull request.

>- We should add an overload (at the type level) such that if it receives null, it returns null

Unless you meant something internal to the fragment reference here.


rbalicki2:
Thanks @adrtivv for starting this! Per our discussion, it looks like we're blocked on https://github.com/isographlabs/isograph/issues/346 :/

PR #344, opened 2025-01-29
Get closer(?) to a fix for the TypeScript caching(?) + overloads issue

rbalicki2:
On this commit, I can change the entrypoint's normalization ast type to NormalizationAst <-> NormalizationAstLoader, and everything behaves correctly, i.e. it fails to typecheck then passes typechecking, instead of being stuck in a "fails to typecheck" state.

Unfortunately, one ts test does not behave as we would hope:
```
  // if the type is unknown there can be no ast so we should use the same rules
  // @ts-expect-error if the type is unknown, require `shouldFetch` to be specified
  useLazyReference(withAstOrLoader, {});
  useLazyReference(withAstOrLoader, {}, { shouldFetch: 'Yes' });
  // @ts-expect-error if the type is unknown, `shouldFetch` can't be `IfNecessary`
  useLazyReference(withAstOrLoader, {}, { shouldFetch: 'IfNecessary' });
```

Both `ts-expect-error`'s fail to be triggered.

This can be understood based on the `type Blah`, which has type `"They are equal?" | "Of course theyre not"`.

Which, I guess, is fair.

Anyway, still trying to figure this out. I've looked into this before, and I don't remember if I ever came up with a good solution for this.

### DIFF
--- a/demos/pet-demo/src/components/__isograph/Query/HomeRoute/entrypoint.ts
+++ b/demos/pet-demo/src/components/__isograph/Query/HomeRoute/entrypoint.ts
@@ -1,4 +1,4 @@
-import type {IsographEntrypoint, NormalizationAst, RefetchQueryNormalizationArtifactWrapper} from '@isograph/react';
+import type {IsographEntrypoint, NormalizationAst, NormalizationAstLoader, RefetchQueryNormalizationArtifactWrapper} from '@isograph/react';
 import {Query__HomeRoute__param} from './param_type';
 import {Query__HomeRoute__output_type} from './output_type';
 import readerResolver from './resolver_reader';
--- a/demos/pet-demo/src/components/__isograph/Query/HomeRoute/entrypoint.ts
+++ b/demos/pet-demo/src/components/__isograph/Query/HomeRoute/entrypoint.ts
@@ -1,7 +1,8 @@
-import type {IsographEntrypoint, NormalizationAst, NormalizationAstLoader, RefetchQueryNormalizationArtifactWrapper} from '@isograph/react';
+import type {FetchOptions, IsographEntrypoint, NormalizationAst, NormalizationAstLoader, RefetchQueryNormalizationArtifactWrapper} from '@isograph/react';
 import {Query__HomeRoute__param} from './param_type';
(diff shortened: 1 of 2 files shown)


rbalicki2:
```
export type Equal2<X, Y, IfTrue, IfFalse> =
  (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y ? 1 : 2
    ? IfTrue
    : IfFalse;
```

works in the file. But it causes correct `useLazyReference` calls to fail to typecheck

rbalicki2:
```
  ...[fetchOptions]: TNormalizationAst extends {
    kind: 'NormalizationAstLoader';
  }
    ? [fetchOptions: RequiredFetchOptions<TClientFieldValue>]
    : [fetchOptions?: FetchOptions<TClientFieldValue>]
```
seems to work at the callsites, but also fails to catch the `withAstOrLoader` errors

rbalicki2:
Okay, the latest:

- it correctly typechecks for entrypoints with type `NormalizationAst` or `NormalizationAstLoader`
- it also allows you to pass two args to `NormalizationAst | NormalizationAstLoader`, *but* in that case, the return type is `unknown`
- and you can't destructure `unknown` to get a `fragmentReference`

I think we can also add a runtime check that the value passed to `useLazyReference` is stable. Thoughts?

I wonder if this will be a problem when we add routing support. At that point, some API will need to be generic over many different entrypoints. But perhaps you could close over them somehow.

But regardless, I would be happy to land like this.

------

Other thoughts, which I haven't explored:

```
useLazyReference(entrypoint.combineWithFetchPolicy('Yes'), variables, otherOptions)
```
Now, the fetch policy stuff is nowhere to be found in the type signature of `useLazyReference`

The solution in this PR might get untenable, if we have more variations (lazy normalization ast? lazy reader AST? etc), in a way that the builder pattern does not.

rbalicki2:
Maybe we can also consider making `Yes` the default for entrypoints w/normalization ast loaders? (I think that won't help)

In addition, some entrypoints will have their ast already loaded. I wonder how we can (in the future) allow for `IfLoadedIfNecessaryOtherwiseYes` and `IfLoadedIfNecessaryOtherwiseNo` LOL

PR #351, opened 2025-01-30
Print combined schemas

rbalicki2:
## Summary

As part of generating artifacts, also generate a graphql schema that contains all of the objects + fields, and scalars, that exist in the schema object, and write it to `/combined-graphql-schema.graphql`.

I'm not in love with the name.

## Why is this so badly written??

I would love a set of eyes on this. I don't think this is my finest work, but I'm (unless convinced otherwise) okay with this being not super good, because (in the long run) there is work that is required to do this "right", and that work is not the highest priority at the moment.

## Ways in which this is necessarily bad:

- we don't keep track of whether objects are input objects or output objects, because that doesn't matter to Isograph
- we also don't keep track of whether something is an interface, or union, or what.
- so, everything is "type"

## I would like to (as a follow up):

- Include in client fields' descriptions a link to the file and line number, and perhaps the text of the isograph literal

## In the end, doing this right will entail:

- Adding a type parameter to IsographSchema, which varies based on whether we are in GraphQL-land, SQL-land, etc. This parameter will include info on whether the item was an Input, or an Interface, whether to skip printing the field (for typename) etc

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
       - name: Make artifact executable
         run: chmod +x ./artifacts/linux-x64/isograph_cli
       - name: 'Build project'
-        run: ./artifacts/linux-x64/isograph_cli --config ./demos/${{ matrix.target.folder }}/isograph.config.json
+        run: cd ./demos/${{ matrix.target.folder }} && ../../artifacts/linux-x64/isograph_cli --config ./isograph.config.json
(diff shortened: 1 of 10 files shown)


PatrykWalach:
We also should print the arguments

rbalicki2:
Oh yeah, I forgot about that, thanks

rbalicki2:
closing for the time being, will re-implement atop work related to #388

PR #356, opened 2025-01-31
parse client pointer `to` keyword

PatrykWalach:
Part of client pointers #53

### DIFF
--- a/crates/isograph_lang_parser/src/parse_iso_literal.rs
+++ b/crates/isograph_lang_parser/src/parse_iso_literal.rs
@@ -236,6 +236,20 @@ fn parse_client_pointer_declaration_inner(
             .parse_string_key_type(IsographLangTokenKind::Identifier)
             .map_err(|with_span| with_span.map(IsographLiteralParseError::from))?;
 
+        let target_type = match tokens.parse_source_of_kind(IsographLangTokenKind::Identifier) {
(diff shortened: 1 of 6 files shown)


crates/isograph_lang_parser/src/parse_iso_literal.rs
```diff
@@ -236,6 +236,20 @@ fn parse_client_pointer_declaration_inner(
             .parse_string_key_type(IsographLangTokenKind::Identifier)
             .map_err(|with_span| with_span.map(IsographLiteralParseError::from))?;
 
+        let target_type = match tokens.parse_source_of_kind(IsographLangTokenKind::Identifier) {
+            Ok(keyword) => {
```
rbalicki2, line 240: What's the use case for making `to Type` optional?

Could you add fixtures that exercise both cases (if we keep both cases) or modify the existing fixture if `to Type` is required?

rbalicki2 (changes requested):

crates/isograph_lang_parser/src/parse_iso_literal.rs
PatrykWalach: I'll make it required, it seems unlikely to point to the same type

rbalicki2 (approved):
slick

PR #358, opened 2025-01-31
Generate client pointer param type

PatrykWalach:
Part of client pointers #53


rbalicki2 (approved):
LGTM lets shipit and eventually do the interface thing

PR #359, opened 2025-01-31
validate client pointer is not selected as scalar

PatrykWalach:


crates/isograph_schema/src/validate_schema.rs
```diff
@@ -477,6 +480,20 @@ pub enum ValidateSchemaError {
         client_field_name: SelectableFieldName,
         field_parent_type_name: IsographObjectTypeName,
         field_name: SelectableFieldName,
+        client_type: String,
+    },
+
+    #[error(
+        "In the client {client_type} `{client_field_parent_type_name}.{client_field_name}`, the \
```
rbalicki2, line 487: nit:

```
        "In the client {client_type} `{client_field_parent_type_name}.{client_field_name}`, the \
        pointer `{field_parent_type_name}.{field_name}` is selected as a scalar. \
        However, client pointers can only be selected as linked fields."
```

It's not perfect... we should refine our vocabulary at some point

rbalicki2 (changes requested):
just a nit, otherwise looks great

PR #363, opened 2025-02-02
merge client pointer reader selection set

PatrykWalach:


rbalicki2 (approved):
shipitt!!! Sorry that it took so long to review

PR #364, opened 2025-02-02
refactor pico to make db immutable

ch1ffa:
### Motivation

The current `Storage` implementation does not allow passing arguments by reference to a memoized function because it leads to a situation where both mutable and immutable `Db` borrows occur. The correct way to fix this is to pass the `Db` argument as an immutable reference instead of a mutable one.

Unfortunately, this cannot be easily achieved using interior mutability or something like `DashMap` due to the creation of a temporary guard, which prevents returning a reference to a value from a map in functions like `get`.

### Solution

Here is the new `Storage` struct:

```rs
pub struct Storage<Db: Database + ?Sized> {
    pub current_epoch: Epoch,
    pub dependency_stack: DependencyStack<(Epoch, Dependency)>,
    pub params_map: DashMap<ParamId, Index>,
    pub derived_nodes: DerivedNodesStore,
    pub source_nodes: DashMap<Key, SourceNode>,
    pub capacity: usize,
    pub map: OnceMap<Epoch, Box<Generation<Db>>>,
}
```

Everything that implement `Copy`, do not need to return a reference, or can be put behind a guard due to the atomicity of changes, can be stored in something like `DashMap` or `RefCell`. 
- `current_epoch` remains unchanged. It can be incremented only upon a `Source` change, where we have a mutable reference to the `Database`. 
- `dependency_stack` is mutated during the `inner_fn` call and at the end of the `memo` call. For single-threaded cases, it is enough to put it behind a `RefCell`. 
- `source_nodes` is just a `DashMap` because it is modified between `memo` calls, where we have a mutable reference to the `Database` (if we introduce `get_ref` for sources, this must be changed)

Everything that needs to return a reference or cannot be `Copy` must be stored in an append-only structure. Accumulated data can be dropped altogether.

To store these values, we use `OnceMap`, which holds read-only "snapshots" (so-called `Generation`) for each epoch:

```rs
pub struct Generation<Db: Database + ?Sized> {
    pub params: Arena<Box<dyn Any>>,
    pub derived_nodes: Arena<DerivedNode<Db>>,
}
```

`Arena` is a new type based on `boxcar::Vec`. We can use any other structure for `Generation` and `Arena` as long as it satisfies the following:
- Allows appending values via `&self`.
- Returns `&T` without creating a guard.
- Implements `Debug` or has a non-consuming iterator.

Currently, we store parameters and derived nodes in the `map`. Source nodes may also be stored here if we find out that references are needed for it.

Inserting a value into `Arena` returns its index within the `Vec`. Together with the `Epoch` when it was inserted, it uniquely identifies the value (e.g. param or derived node). It is called `Index` and is stored in `params_map` for parameters and `derived_nodes.map` for derived nodes.

We must store `time_updated` and `time_verified` separately from `derived_node` data because these values can be updated without changing the actual `derived_node` structure during verification:

```rs
pub struct DerivedNodesStore {
    pub map: DashMap<DerivedNodeId, Index>,
    pub rev: DashMap<DerivedNodeId, DerivedNodeRevision>,
}
```

### What happens during a `memo` call?
- On the first run, every encountered node and parameter is calculated and stored in the `Generation` under `Epoch = 1`. Their indexes are also added to separate maps that store only the latest indexes (since they are `Copy`).
- On any source node change, the epoch is incremented, and a new `Generation` is created.
- If `memo` is called again, we can access a read-only version of the previous `Generation` and borrow data any times we want. All new data is stored in the current `Generation`.
- Without modifying a source node (i.e. incrementing the epoch), the current generation is unchanged (for pure functions)

Currently, the implementation keeps the last `N` generations, but we can use any other retention policy. If any parameter or derived node is missing, it is safe and they will be recalculated.


crates/pico_core/src/storage.rs
```diff
@@ -1,25 +1,160 @@
...
+    }
+}
+
+impl<Db: Database + ?Sized> Storage<Db> {
+    pub fn increment_epoch(&mut self) -> Epoch {
+        let current_epoch = self.current_epoch.increment();
+        self.map
+            .insert(current_epoch, |_| Box::new(Generation::new()));
+        // TODO: not the best way to do it, worth to create something like OnceOrderedMap
+        if self.map.read_only_view().len() >= self.capacity {
```
rbalicki2, line 49: IMO it's unexpected that `increment_epoch` would have such consequences...

crates/pico_core/src/storage.rs
```diff
@@ -1,25 +1,160 @@
...
+            let min_epoch = *self.map.read_only_view().keys().min().unwrap();
+            self.map.remove(&min_epoch);
+        }
+        current_epoch
+    }
+
+    pub fn contains_param(&self, param_id: ParamId) -> bool {
```
rbalicki2, line 56: perhaps we should use the entry API here? 

```
if !db.storage().contains_param(param_id) {
            db.storage().insert_param(param_id, param);
        }
```
might be more efficient, or at least more idiomatic, if we did so

crates/pico_core/src/dependency_stack.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#[derive(Debug, Default)]
+pub struct DependencyStack<T>(RefCell<Vec<Vec<T>>>);
+
+impl<T> DependencyStack<T> {
+    pub fn new() -> Self {
+        Self(RefCell::new(Vec::new()))
+    }
+
+    pub fn enter(&self) {
```
rbalicki2, line 11: nit: perhaps we return a guard that, when dropped, pops the item off the stack?

crates/pico_core/src/storage.rs
```diff
@@ -1,25 +1,160 @@
...
+        }
+    }
+}
+
+impl<Db: Database + ?Sized> Storage<Db> {
+    pub fn increment_epoch(&mut self) -> Epoch {
+        let current_epoch = self.current_epoch.increment();
+        self.map
+            .insert(current_epoch, |_| Box::new(Generation::new()));
+        // TODO: not the best way to do it, worth to create something like OnceOrderedMap
```
rbalicki2, line 48: `OnceOrderedMap` or perhaps a circular buffer? anyway let's reassess what we're doing for GC

crates/pico_core/src/node.rs
```diff
@@ -20,24 +20,26 @@ impl DerivedNodeId {
 }
 
 pub struct DerivedNode<Db: Database + ?Sized> {
-    pub time_verified: Epoch,
-    pub time_updated: Epoch,
     pub dependencies: Vec<Dependency>,
```
rbalicki2, line 23: Just idly thinking... I feel like a *lot* of functions will have one (or max a few) dependencies. So, maybe something like https://docs.rs/smallvec/latest/smallvec/ will help save a memory lookup in the common case.

In particular, I imagine we have many functions like `ast_without_locations(db, ast_id: SomeId) { db.lookup(ast_id).remove_locations() }`. If we have something like `hover_info`, which depends on `full_ast`, every (even irrelevant) change will cause `hover_info` to be recalculated. It's better to have that intermediate function.

Let's do that as a follow up, though

rbalicki2 (changes requested):
Looks amazing. A bunch of nits. I really really like this. I'm going to need to review a second time, since this is a lot 😂 

I think the only blocker (other than me re-reviewing) is — can we get rid of `Database`? It seems pretty useless right now. Everything else can be fixed on the `pico` branch.

> We must store time_updated and time_verified separately from derived_node data because these values can be updated without changing the actual derived_node structure during verification:

* I don't quite follow. Can you explain more? What prevents us from keeping them in the same struct?
* I wonder, do we have any risks related to GC? e.g. if a generation is GC'd, but there is still a param in params_map that refers to that epoch?

crates/pico/src/memo.rs
```diff
@@ -156,36 +153,40 @@ fn any_dependency_changed<Db: Database>(
 }
 
 fn source_node_changed_since<Db: Database>(db: &Db, key: Key, since: Epoch) -> bool {
-    match db.storage().source_nodes().get(&key) {
+    match db.storage().source_nodes.get(&key) {
         Some(source) => source.time_updated > since,
         None => panic!("Source node not found. This indicates a bug in Pico."),
```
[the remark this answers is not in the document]
ch1ffa: Yeah, on one hand, it's not a bug, but rather an incorrect source usage. I can't think of a situation when some `derived_node` could store a removed source key - it's not a pure function then.

On the other hand, here we check dependencies from the previous call, so we can return `true` instead of panic and force `inner_fn` to recalculate and probably fix dependencies.

crates/pico_core/src/storage.rs
```diff
@@ -1,25 +1,160 @@
...
-    fn current_epoch(&self) -> Epoch;
+#[derive(Debug)]
+pub struct Storage<Db: Database + ?Sized> {
+    pub current_epoch: Epoch,
+    pub dependency_stack: DependencyStack<(Epoch, Dependency)>,
+    pub params_map: DashMap<ParamId, Index>,
+    pub derived_nodes: DerivedNodesStore,
+    pub source_nodes: DashMap<Key, SourceNode>,
+    pub capacity: usize,
+    pub map: OnceMap<Epoch, Box<Generation<Db>>>,
```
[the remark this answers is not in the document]
ch1ffa: Maybe just `id_to_index`. It will be used as `self.derived_nodes.id_to_index.get(&derived_node_id)?`

crates/pico_core/src/database.rs
```diff
@@ -1,17 +1,16 @@
...
-    source::SourceId,
-    storage::{Storage, StorageMut},
+    source::{Source, SourceId},
+    storage::Storage,
 };
 
 pub trait Database {
-    fn storage(&self) -> &impl Storage<Self>;
-    fn storage_mut(&mut self) -> &mut impl StorageMut<Self>;
+    fn storage(&self) -> &Storage<Self>;
```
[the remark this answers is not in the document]
ch1ffa: I removed traits, because there is no LRU container anymore, so no need to abstract over it. 

As for `Database`, the main idea behind it is to hide the pico internals. For example, we could have a singe `LSPState` structure that has `storage: Storage<Self>` field and derives `Database`. LSP state can also store any information that is accessible during computation, but should not be a part of memoization.

Alternatively, we can just pass `lsp_state.storage` field to a memoized function first argument and avoid possible side-effects from LSPState data being used incorrectly (making it non-pure).

So I'm not sure what is better. I'll do what you decide here. We can always rollback if we find a case for it.

crates/pico_macros/src/db.rs
```diff
@@ -54,24 +49,22 @@ pub fn derive_db(input: TokenStream) -> TokenStream {
...
-                use ::pico_core::{storage::{Storage, StorageMut}, container::Container};
                 let id = ::pico_core::source::SourceId::new(&source);
-                let time_updated = if self.storage().source_nodes().contains_key(&id.key) {
+                let time_updated = if self.storage().source_nodes.contains_key(&id.key) {
                     self.increment_epoch()
                 } else {
                     self.current_epoch()
                 };
-                self.storage_mut().source_nodes_mut().insert(id.key, ::pico_core::node::SourceNode {
+                self.storage_mut().source_nodes.insert(id.key, ::pico_core::node::SourceNode {
```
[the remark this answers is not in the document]
ch1ffa: yeah, internally in `db.rs` we can use `storage` field. But in the `drop` test I call `state.storage_mut().epoch_to_generation_map.remove(&min_epoch)` to drop a generation.

crates/pico_core/src/storage.rs
ch1ffa: Agreed. As we discussed, it is better to have a GC task to preserve more information during user's typing. Need to think about a good API to expose.

crates/pico_core/src/storage.rs
```diff
@@ -1,25 +1,160 @@
...
+impl<Db: Database + ?Sized> Default for Storage<Db> {
+    fn default() -> Self {
+        Self::new(10000)
+    }
+}
+
+#[derive(Debug, Default)]
+pub struct DerivedNodesStore {
+    pub map: DashMap<DerivedNodeId, Index>,
+    pub rev: DashMap<DerivedNodeId, DerivedNodeRevision>,
```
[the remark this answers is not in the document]
ch1ffa: what about `id_to_index` and `id_to_revision`?

crates/pico_core/src/storage.rs
ch1ffa: Unfortunately, there are two maps involved and no single entry exists. Can you show what can we do here?

crates/pico_core/src/dependency_stack.rs
ch1ffa: Oh, that's neat! We can protect a case when we forgot to call `leave`

crates/pico_core/src/storage.rs
ch1ffa: Ah, you mean to use entry API for that particular insert. Yeah, makes sense.

ch1ffa: idk, I need to check that epoch is not dropped. It's more verbose
```rs
pub fn insert_param<T: Clone + 'static>(&self, param_id: ParamId, param: T) {
    match self.param_id_to_index.entry(param_id) {
        Entry::Occupied(mut entry) => {
            if !self.epoch_to_generation_map.contains_key(&entry.get().epoch) {
                let idx = self
                    .epoch_to_generation_map
                    .get(&self.current_epoch)
                    .unwrap()
                    .params
                    .push(Box::new(param));
                entry.insert(Index::new(self.current_epoch, idx));
            }
        },
        Entry::Vacant(entry) => {
            let idx = self
                .epoch_to_generation_map
                .get(&self.current_epoch)
                .unwrap()
                .params
                .push(Box::new(param));
            entry.insert(Index::new(self.current_epoch, idx));
        },
    }
}
```

crates/pico_core/src/dependency_stack.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    pub fn leave(&self) -> Vec<T> {
+        self.0
+            .borrow_mut()
+            .pop()
+            .expect("Dependency stack should not be empty. Leave must be called after enter. This indicates a bug in Pico.")
+    }
+
+    pub fn push_checked(&self, make_val: impl FnOnce() -> T) {
```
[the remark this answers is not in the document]
ch1ffa: I need to push a value only if stack is not empty. Maybe `push_if_not_empty`?

crates/pico_core/src/storage.rs
rbalicki2: okay, no worries then! It was just a thought. I didn't think too hard about how that would actually be implemented

crates/pico_core/src/params.rs
```diff
@@ -2,41 +2,23 @@ use std::collections::hash_map::DefaultHasher;
 use std::hash::Hash;
 use std::hash::Hasher;
 
-use crate::container::Container;
 use crate::database::Database;
-use crate::storage::StorageMut;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct ParamId(u64);
```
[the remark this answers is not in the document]
rbalicki2: Ah I see. Okay, just a thought, and totally nbd, and definitely not a blocker

crates/pico_core/src/dependency_stack.rs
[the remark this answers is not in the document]
rbalicki2: I like that name

crates/pico_core/src/storage.rs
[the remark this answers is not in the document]
rbalicki2: I like those names

crates/pico/src/memo.rs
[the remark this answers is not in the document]
ch1ffa: ok, I'll rephrase it

crates/pico_core/src/database.rs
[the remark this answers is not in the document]
rbalicki2: > Alternatively, we can just pass lsp_state.storage field to a memoized function first argument and avoid possible side-effects from LSPState data being used incorrectly (making it non-pure).

Yeah, I was thinking that the memoized function would receive an `&Storage` param (which no longer needs to be generic, right?)

If you're passing anything else around (e.g. `LSPState.foo`) then that should be another param, so that we can memoize it.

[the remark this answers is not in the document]
ch1ffa: Yeah, it is not generic anymore. That's cool. Maybe we should rename `Storage` to `Database` then?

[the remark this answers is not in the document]
rbalicki2: Good call, let's rename `Storage` to `Database`

crates/pico/tests/arg_reference.rs
```diff
@@ -0,0 +1,79 @@
...
+        value: "2 + 2 * 2".to_string(),
+    });
+
+    let value = evaluate_input(&db, input);
+    // assert that the value was not cloned
+    assert_eq!(VALUE.load(Ordering::SeqCst), 0);
+
+    let expected = Expected(6);
+    assert_result(&db, value, expected);
+    // assert that the argument of type `&Value` was cloned only once to params store
```
rbalicki2, line 27: Nit: `&Value` was cloned once, when it was added to the params store.

----

Does this sort of negate the utility of using references?

crates/pico/tests/arg_reference.rs
```diff
@@ -0,0 +1,79 @@
...
+    // and internally used by reference
+    assert_eq!(VALUE.load(Ordering::SeqCst), 1);
+    // compare with the argument of type `Expected` which is cloned twice
+    assert_eq!(EXPECTED.load(Ordering::SeqCst), 2);
```
rbalicki2, line 31: nit: once to store it in the params store, and once to call `inner_fn`.

------

This makes me wonder — can we avoid cloning when calling `inner_fn`, if the params were passed in? (as they were in this case)

crates/pico_core/src/index.rs
```diff
@@ -0,0 +1,21 @@
+use std::fmt;
+
+use crate::epoch::Epoch;
+
+#[derive(Clone, Copy)]
+pub struct Index {
```
rbalicki2, line 6: nit: can `Index` be generic over the type of some phantom data? Seeing `DashMap<ParamId, Index>` makes me nervous

crates/pico_core/src/database.rs
```diff
@@ -1,17 +1,221 @@
...
+    pub derived_nodes: DerivedNodesStore,
+    pub source_nodes: DashMap<Key, SourceNode>,
+    pub capacity: usize,
+    pub epoch_to_generation_map: OnceMap<Epoch, Box<Generation>>,
+}
+
+impl Database {
+    pub fn new(capacity: usize) -> Self {
+        let epoch_to_generation_map = OnceMap::new();
+        epoch_to_generation_map.insert(Epoch::new(), |_| Box::new(Generation::new()));
```
rbalicki2, line 32: nit: probably worth making `Epoch::new` a variable for clarity, even though it implements copy

crates/pico_core/src/dependency_stack.rs
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
+
+pub struct DependencyStackGuard<'a, T> {
+    stack: &'a DependencyStack<T>,
+    released: bool,
+}
+
+impl<T> DependencyStackGuard<'_, T> {
+    pub fn release(mut self) -> Vec<T> {
```
rbalicki2, line 42: Is there an actual case we would want to release early? I think not...

crates/u64_newtypes/src/lib.rs
```diff
@@ -0,0 +1,50 @@
+#[macro_export]
```
rbalicki2, line 1: copycat :-P

rbalicki2 (changes requested):
partial review, a few nits. Overall, I love this, and see nothing that is preventing a merge.

* I would like to review and really think about the Database struct, clearly this is gonna take a few days for me to understand 😭 
* I would like to tackle GC before we merge into master

crates/pico_core/src/dependency_stack.rs
```diff
@@ -0,0 +1,55 @@
+use std::cell::RefCell;
+
+#[derive(Debug, Default)]
+pub struct DependencyStack<T>(RefCell<Vec<Vec<T>>>);
```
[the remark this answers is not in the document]
rbalicki2: Anyway, `(Epoch, Dependency)` is confusing... so maybe what we actually have is a struct
```
struct Foo {
  dependencies: Vec<Dependency>,
  max_time_updated: Epoch
}
impl Foo {
  fn push(&mut self, dependency, time_updated) {
    self.max_time_updated = std::cmp::max(time_updated, self.max_time_updated)
  }
}
```
since this is what we're doing anyway. That might make it easier to understand.

But also, could be follow up

crates/pico_core/src/dependency_stack.rs
ch1ffa: Wdym? We have to call release in the end to pop the `Vec` anyway. `Drop` implemented if we forgot to do it.

crates/pico_core/src/database.rs
```diff
@@ -1,17 +1,221 @@
...
-    fn increment_epoch(&mut self) -> Epoch;
-    fn get<T: Clone + 'static>(&mut self, id: SourceId<T>) -> T;
-    fn set<T: Source + DynEq>(&mut self, source: T) -> SourceId<T>;
-    fn remove<T>(&mut self, id: SourceId<T>);
+#[derive(Debug)]
+pub struct Database {
+    pub current_epoch: Epoch,
+    pub dependency_stack: DependencyStack<(Epoch, Dependency)>,
+    pub param_id_to_index: DashMap<ParamId, Index>,
+    pub derived_nodes: DerivedNodesStore,
```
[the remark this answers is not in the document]
ch1ffa: Yep, I added it to explore if they are used together, but they are not.

crates/pico_core/src/dependency_stack.rs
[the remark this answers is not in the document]
ch1ffa: Maybe something like `TrackedDependencies`?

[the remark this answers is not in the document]
ch1ffa: nvm, it is just a `Dependency`

crates/pico/src/lib.rs
```diff
@@ -1,3 +1,53 @@
...
+mod source;
+mod u64_types;
+
+pub use database::*;
+pub use derived_node::*;
+pub use memo::*;
+pub use source::*;
+pub use u64_types::*;
+
+pub fn intern_param<T: Hash + Clone + 'static>(db: &Database, param: T) -> ParamId {
```
rbalicki2, line 26: not blocking, but I'll move this (and any other fns I find that are only used by the macro) into a sub module

rbalicki2 (approved):
shipittt

PR #369, opened 2025-02-06
make `startUpdate` stable

PatrykWalach:


libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -19,6 +20,17 @@ export type ComponentCache = {
   };
 };
 
+export type Resolver = TopLevelReaderArtifact<any, any, any>['resolver'];
+
+export type StartUpdateCache = WeakMap<
+  // this is equivalent to CaomponentCache, but we use reference to Resolver
+  // instead of component name, because only components have names
+  Resolver,
+  {
```
rbalicki2, line 29: I don't care whether we use a weak map or an object, but is it theoretically possible to use an object? (So to be clear, don't take this comment as a suggestion you modify the code.) Namely, does the fact that we're using a weak map mean that we're going to have trouble storing this in the store?

For components, we do id -> component name (equivalent to resolver name) -> args, can we do the same for startUpdate? I think that this would require including the name of the client field in the eager reader artifact, which I am okay with.

I am 90% sure that I added the name to the component artifact to make the component cache work, FWIW

libs/isograph-react/src/core/read.ts
```diff
@@ -263,11 +264,41 @@ function readData<TReadFromStore>(
               recordLink: data.recordLink,
             };
           }
+
+          const readerWithRefetchQueries = {
+            kind: 'ReaderWithRefetchQueries',
+            readerArtifact: field.condition,
+            // TODO this is wrong
+            // should map field.condition.usedRefetchQueries
+            //  but it doesnt' exist,
```
rbalicki2, line 273: spacing and `doesnt'`

rbalicki2 (changes requested):
wdyt?

libs/isograph-react/src/core/IsographEnvironment.ts
rbalicki2: We can also fix the `TODO_FRAGMENT_NAME` in `FragmentReference.ts`

PatrykWalach: To use object as a key, we have to use `WeakMap` or `Map`
If we generate name for normal resolvers, then we can use an object

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -19,6 +20,17 @@ export type ComponentCache = {
   };
 };
 
+export type Resolver = TopLevelReaderArtifact<any, any, any>['resolver'];
+
+export type StartUpdateCache = WeakMap<
+  // this is equivalent to CaomponentCache, but we use reference to Resolver
```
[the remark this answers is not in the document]
PatrykWalach: I think this a good idea plus I just realized I don't have to save `startUpdate` for the `Component`
```tsx
function startUpdate() {
  // ...
}

store[key] = function Component(){
  const arg = {
    startUpdate, // <-- this is stable
  }

  // ...
}
```

libs/isograph-react/src/core/IsographEnvironment.ts
rbalicki2: okay, yeah, IMO we should just generate the name for the resolver then

PatrykWalach: I agree

rbalicki2: nice. This can probably also make the component cache single layer

libs/isograph-react/src/core/componentCache.ts
```diff
@@ -7,51 +7,63 @@ import { IsographEnvironment } from './IsographEnvironment';
...
-            : undefined,
-        },
-        additionalRuntimeProps,
+  switch (cacheEntry.kind) {
+    case 'Component': {
+      return cacheEntry.component;
+    }
+    case 'EagerReader': {
+      throw new Error(
+        'Expected component. ' + 'This is indicative of a bug in Isograph.',
```
rbalicki2, line 65: nit combine strings

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,13 +1,48 @@
...
+
+  switch (cacheEntry.kind) {
+    case 'EagerReader': {
+      return (cacheEntry.startUpdate ??= createStartUpdate(
+        environment,
+        fragmentReference,
+      ));
+    }
+    case 'Component': {
+      throw new Error(
```
rbalicki2, line 33: hm. Since we're just throwing if we get the wrong one, maybe we should just have two separate caches. Sorry for going back and forth on this one.

Thoughts?

See #372

rbalicki2 (changes requested):
let's split out these caches then looks great to me

rbalicki2 (approved):

PR #370, opened 2025-02-06
fix collisions in component cache

PatrykWalach:
We use only `fragmentReference.root.__link` `id` instead of `id` + `__typename` in cache,
We don't see any errors, because second layer `componentName` includes `__typename` (e.g. `Pet.resolver`)

For resolvers defined on an abstract types we would encounter collisions

### DIFF
--- a/libs/isograph-react/src/core/componentCache.ts
+++ b/libs/isograph-react/src/core/componentCache.ts
@@ -15,12 +15,9 @@ export function getOrCreateCachedComponent(
   fragmentReference: FragmentReference<any, any>,
   networkRequestOptions: NetworkRequestReaderOptions,
 ): React.FC<any> {
-  // cachedComponentsById is a three layer cache: id, then component name, then
-  // stringified args. These three, together, uniquely identify a read at a given
-  // time.
-  const cachedComponentsById = environment.componentCache;
+  const cachedComponentsByStableFragmentReferenceId = environment.componentCache;
 
-  const componentsByName = (cachedComponentsById[
+  const componentsByName = (cachedComponentsByStableFragmentReferenceId[
     stableIdForFragmentReference(fragmentReference)
   ] ??= {});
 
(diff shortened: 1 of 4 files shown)


rbalicki2:
nice, great catch. Will take a look!

libs/isograph-react/src/core/componentCache.ts
```diff
@@ -18,16 +20,11 @@ export function getOrCreateCachedComponent(
   // time.
   const cachedComponentsById = environment.componentCache;
```
rbalicki2, line 21: nit: `cachedComponentsByStableFragmentReferenceId`

rbalicki2 (approved):
awesome work. Will rewrite variable names and a comment slightly and land

rbalicki2:
Hm... do we have collisions? Yeah, we have `componentName: "NewsfeedItem.NewsfeedAdOrBlog",`, but if I attempt to define a field `NewsfeedAdOrBlog` on `AdItem`, I get a compile error

`Reason: The Isograph compiler attempted to create a field named "NewsfeedAdOrBlog" on type "AdItem", but a field with that name already exists.`

And if I do that via `AnotherUnion.NewsfeedAdOrBlog`, same thing

rbalicki2 (changes requested):
☝️  previous comment

PatrykWalach:
given
a single field `Union.NewsfeedAdOrBlog`
and we do `getOrCreateCachedComponent({ __link: 1, __typename: "NewsfeedAd" })`
we will write key: `1/Union.NewsfeedAdOrBlog/variables/`

when
we do `getOrCreateCachedComponent({ __link: 1, __typename: "Blog" })`
we will read from key: `1/Union.NewsfeedAdOrBlog/variables/`


and receive the component reading from old fragment reference `{ __link: 1, __typename: "NewsfeedAd" }`

rbalicki2:
WOW

PR #373, opened 2025-02-08
Introduce pico into codebase

rbalicki2:
- Pico is our incremental calculation framework. Amazing work by @ch1ffa


rbalicki2:
doesn't need to be an open PR yet

PR #375, opened 2025-02-08
Saving work in progress

edmondop:
Initial pico commit, amazing work by @ch1ffa 

TLDR
- this is a simpler, smaller version of Salsa, which is the incremental compilation framework used by Rust Analyzer
- the workhorse is the memo function


rbalicki2:
closing, will use the branch as reference

PR #378, opened 2025-02-11
reuse interned value in fn args

ch1ffa:
### The Problem

The current implementation of Pico interns memoized function parameters as a tuple, which has the following disadvantages:
- Possible data duplication in the database if the same argument is used in different memoized functions—it's cloned as many times as it is used.
- Unable to reuse a memoized function result. Even though it is already present in the database, it gets cloned and put in the params storage anyway. `SourceId` is `copy`, but it is better to reuse it and not intern it as a param.

To fix this, we need to intern parameters independently and find a way to detect arguments of types `SourceId` and arguments that are results of another memoized function.

### The Solution

Instead of interning a tuple, we intern each param individually. First, we generate a `ParamId` by hashing an argument. Then, we check if it is already present in the database to avoid unnecessary cloning. If not, we clone the argument and insert it into the params storage in the database.

If an argument is of the `SourceId` type, we simply convert `SourceId` to `ParamId` (internally they are just new types over a `u64` hash) before the `memo` call and convert it back inside `innerFn`.

To support reusing a memoized function result, we introduce a new struct:
```rs
pub struct MemoRef<'db, T> {
    db: &'db Database,
    derived_node_id: DerivedNodeId,
    phantom: PhantomData<T>,
}
```

The `reference` parameter has been removed from the macro, and the memoized function now always returns `MemoRef<'_, T>`. It implements `Deref` to get `&T` from the database and includes a `to_owned` method to clone a value from the database.

The `memo` macro now has `#[memo(inner)]` and `#[memo(inner_ref)]` parameters to support the previous implementation, which returned `&T` and `T` directly. These parameters can be used for a root query or inside a memoized function that consumes the result without passing it further (but this may be redundant).

`DerivedNodeId` now has a different implementation because we need to store all `ParamId`s:
```rs
#[derive(Debug, PartialEq, Eq, Hash, Serialize, Deserialize, Default)]
pub struct DerivedNodeDescriptor {
    pub key: Key,
    pub params: ArrayVec<[ParamId; 8]>,
}

intern_struct! {
    pub struct DerivedNodeId = Intern<DerivedNodeDescriptor> {}
}
```

Conversion between `DerivedNodeId` and `ParamId` is done using the interned index:
```rs
impl From<&DerivedNodeId> for ParamId {
    fn from(value: &DerivedNodeId) -> Self {
        Self::from(value.index() as u64)
    }
}

impl From<ParamId> for DerivedNodeId {
    fn from(value: ParamId) -> Self {
        DerivedNodeId::from_index_checked(**value.inner() as u32).unwrap()
    }
}
```

This is safe because the conversion is handled by generated code that knows the argument position and can extract the argument type from the corresponding `DerivedNodeId`.

**Note:** Arguments are analyzed by their type names, so `SourceId` and `MemoRef` should not be aliased.

---

**To Discuss:**  
I removed the `param_id` check in the `derived_node_changed_since` function. I think we should return `Option<Box<dyn DynEq>>` or `Result` from `inner_fn` and force re-evaluation of the parent `inner_fn` upon receiving `None` during the dependency check step. This may occur only if some `ParamId` is garbage collected. We can safely `unwrap` it when calling `inner_fn` in the new node creation step because parameters are interned right before the `inner_fn` call and cannot yet be garbage collected.

---

### TODO:
- [ ] Add new tests to cover the new implementations.  
- [ ] Restore the `ParamId` check before calling `inner_fn`.

### DIFF
--- a/crates/pico/src/derived_node.rs
+++ b/crates/pico/src/derived_node.rs
@@ -1,29 +1,43 @@
-use std::{fmt, hash::Hash};
+use std::{fmt, hash::Hash, marker::PhantomData, ops::Deref};
+
+use intern::{intern_struct, InternId};
+use serde::{Deserialize, Serialize};
+use tinyvec::ArrayVec;
 
 use crate::{
     dependency::Dependency,
     dyn_eq::DynEq,
     epoch::Epoch,
-    u64_types::{Key, ParamId},
+    intern::{Key, ParamId},
     Database,
 };
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct DerivedNodeId {
+#[derive(Debug, PartialEq, Eq, Hash, Serialize, Deserialize, Default)]
+pub struct DerivedNodeDescriptor {
     pub key: Key,
-    pub param_id: ParamId,
+    pub params: ArrayVec<[ParamId; 8]>,
+}
+
+intern_struct! {
+    pub struct DerivedNodeId = Intern<DerivedNodeDescriptor> {}
 }
 
 impl DerivedNodeId {
-    pub fn new(key: Key, param_id: ParamId) -> Self {
-        Self { key, param_id }
+    pub fn new(key: Key, params: ArrayVec<[ParamId; 8]>) -> Self {
+        Self::intern(DerivedNodeDescriptor { key, params })
+    }
+}
+
+impl From<ParamId> for DerivedNodeId {
+    fn from(value: ParamId) -> Self {
+        DerivedNodeId::from_index_checked(**value.inner() as u32).unwrap()
     }
 }
 
 #[derive(Debug, Copy, Clone)]
-pub struct InnerFn(pub fn(&Database, ParamId) -> Box<dyn DynEq>);
+pub struct InnerFn(pub fn(&Database, DerivedNodeId) -> Box<dyn DynEq>);
 impl InnerFn {
-    pub fn new(inner_fn: fn(&Database, ParamId) -> Box<dyn DynEq>) -> Self {
+    pub fn new(inner_fn: fn(&Database, DerivedNodeId) -> Box<dyn DynEq>) -> Self {
         InnerFn(inner_fn)
     }
 }
@@ -47,4 +61,41 @@ impl fmt::Debug for DerivedNode {
 pub struct DerivedNodeRevision {
     pub time_updated: Epoch,
     pub time_verified: Epoch,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct MemoRef<'db, T> {
+    db: &'db Database,
+    derived_node_id: DerivedNodeId,
+    phantom: PhantomData<T>,
+}
+
+impl<'db, T> MemoRef<'db, T> {
+    pub fn new(db: &'db Database, derived_node_id: DerivedNodeId) -> Self {
+        Self {
+            db,
+            derived_node_id,
+            phantom: PhantomData,
+        }
+    }
+}
+
+impl<T: 'static + Clone> MemoRef<'_, T> {
+    pub fn to_owned(&self) -> T {
+        self.deref().clone()
+    }
+}
+
+impl<T: 'static> Deref for MemoRef<'_, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        self.db
+            .get_derived_node(self.derived_node_id)
+            .unwrap()
+            .value
+            .as_any()
+            .downcast_ref::<T>()
+            .unwrap()
+    }
 }
--- a/crates/pico/src/memo.rs
+++ b/crates/pico/src/memo.rs
@@ -4,7 +4,7 @@ use crate::{
     derived_node::{DerivedNode, DerivedNodeId},
     dyn_eq::DynEq,
     epoch::Epoch,
-    u64_types::{Key, ParamId},
+    intern::Key,
     InnerFn,
 };
 
@@ -71,8 +71,7 @@ fn create_derived_node(
     derived_node_id: DerivedNodeId,
     inner_fn: InnerFn,
 ) -> (Epoch, DidRecalculate) {
-    let (value, tracked_dependencies) =
-        with_dependency_tracking(db, derived_node_id.param_id, inner_fn);
+    let (value, tracked_dependencies) = with_dependency_tracking(db, derived_node_id, inner_fn);
     db.insert_derived_node(
         derived_node_id,
         DerivedNode {
@@ -98,8 +97,7 @@ fn update_derived_node(
     prev_value: &dyn DynEq,
     inner_fn: InnerFn,
 ) -> (Epoch, DidRecalculate) {
-    let (value, tracked_dependencies) =
-        with_dependency_tracking(db, derived_node_id.param_id, inner_fn);
+    let (value, tracked_dependencies) = with_dependency_tracking(db, derived_node_id, inner_fn);
     let did_recalculate = if *prev_value != *value {
         db.set_derive_node_time_updated(derived_node_id, tracked_dependencies.max_time_updated);
         DidRecalculate::Recalculated
@@ -140,9 +138,6 @@ fn source_node_changed_since(db: &Database, key: Key, since: Epoch) -> bool {
 }
 
 fn derived_node_changed_since(db: &Database, derived_node_id: DerivedNodeId, since: Epoch) -> bool {
-    if !db.contains_param(derived_node_id.param_id) {
-        return true;
-    }
     let inner_fn = if let Some(derived_node) = db.get_derived_node(derived_node_id) {
         if let Some(rev) = db.get_derived_node_revision(derived_node_id) {
             if rev.time_updated > since {
@@ -161,11 +156,11 @@ fn derived_node_changed_since(db: &Database, derived_node_id: DerivedNodeId, sin
 
 fn with_dependency_tracking(
     db: &Database,
-    param_id: ParamId,
+    derived_node_id: DerivedNodeId,
     inner_fn: InnerFn,
 ) -> (Box<dyn DynEq>, TrackedDependencies) {
     let guard = db.dependency_stack.enter();
-    let value = inner_fn.0(db, param_id);
+    let value = inner_fn.0(db, derived_node_id);
     let tracked_dependencies = guard.release();
     (value, tracked_dependencies)
 }
--- a/crates/pico_macros/src/memo.rs
+++ b/crates/pico_macros/src/memo.rs
@@ -14,7 +14,9 @@ use syn::{
 #[derive(Debug, FromMeta)]
 struct MemoArgs {
     #[darling(default)]
-    reference: bool,
+    inner: bool,
+    #[darling(default)]
+    inner_ref: bool,
 }
 
 pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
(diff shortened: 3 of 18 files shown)


crates/pico_macros/src/memo.rs
```diff
@@ -51,23 +53,27 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
...
-    let reconstructed_args = argument_types
+    let param_ids_blocks = other_args
         .clone()
-        .zip(unpacked_args.clone())
-        .map(|(ty, arg)| {
-            match **ty {
-                syn::Type::Reference(_) => quote!(#arg),
-                _ => quote!(#arg.clone()), // Clone
+        .zip(argument_types.clone())
+        .map(|(arg, ty)| match ArgType::parse(ty) {
```
rbalicki2, line 59: YES, so cool

crates/pico_macros/src/memo.rs
```diff
@@ -14,7 +14,9 @@ use syn::{
 #[derive(Debug, FromMeta)]
 struct MemoArgs {
     #[darling(default)]
-    reference: bool,
+    inner: bool,
+    #[darling(default)]
+    inner_ref: bool,
```
rbalicki2, line 19: should we add a test case that exercises `inner_ref`?

crates/pico_macros/src/memo.rs
```diff
@@ -97,7 +103,9 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
 
     let mut new_sig = sig.clone();
 
-    if args_.reference {
+    if args_.inner {
+        return_expr = quote! { #return_expr.clone() };
+    } else if args_.inner_ref {
         let lifetime = new_sig.generics.params.iter().find_map(|param| {
             if let GenericParam::Lifetime(lt) = param {
                 Some(&lt.lifetime)
```
rbalicki2, line 111: on line 119 (`ReturnType::Type(parse_quote!(->), Box::new(parse_quote!(&#lt #return_type)));`) do we also need to make sure that that lifetime is attached to the `db` param?

crates/pico_macros/src/memo.rs
```diff
@@ -124,27 +132,90 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                    .to_compile_error()
+                    .into();
+            }
+        }
+        new_sig.output = ReturnType::Type(
+            parse_quote!(->),
+            Box::new(parse_quote!(::pico::MemoRef<'db, #return_type>)),
+        );
+        return_expr = quote! {
+            ::pico::MemoRef::new(#db_arg, derived_node_id)
```
rbalicki2, line 150: so the memoized function now returns a `MemoRef<T>` which the user can deref into the actual value?

What's the advantage of doing that, instead of deref'ing for the user? Is it to encourage the use of `MemoRef<T>` as a parameter to memoized functions, so that we can take advantage of the trick above and below?

crates/pico_macros/src/memo.rs
```diff
@@ -124,27 +132,90 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
...
-                    .expect("param should exist. This is indicative of a bug in Pico.");
-                let (#(#unpacked_args,)*) = {
-                    let (#(#inner_args,)*) = param_ref
-                        .downcast_ref::<(#(#get_ref_types,)*)>()
-                        .expect("param type must to be correct. This is indicative of a bug in Pico.");
-                    (
-                        #(#reconstructed_args,)*
-                    )
-                };
+            let mut param_ids = ::pico::macro_fns::init_param_vec();
```
rbalicki2, line 210: nit: this can probably just be inlined here, right?

rbalicki2 (commented):
an initial review... this is gonna take a few passes to understand :)

crates/pico_macros/src/memo.rs
ch1ffa: yeah, working on it. I'm not 100% convinced that it is a very useful parameter, but will see if `MemoRef` is enough.

crates/pico_macros/src/memo.rs
ch1ffa: Exactly. We need a way to recognise that an argument is already in the database

crates/pico_macros/src/memo.rs
ch1ffa: If we ok to reexport `ArrayVec`

crates/pico_macros/src/memo.rs
ch1ffa: hmm, just in case if a function has a lifetime but not the database argument? Ok, let's check that it has the same lifetime as the function.

ch1ffa: btw, we need a test for it

crates/pico/src/derived_node.rs
```diff
@@ -1,29 +1,43 @@
-use std::{fmt, hash::Hash};
+use std::{fmt, hash::Hash, marker::PhantomData, ops::Deref};
+
+use intern::{intern_struct, InternId};
+use serde::{Deserialize, Serialize};
+use tinyvec::ArrayVec;
```
rbalicki2, line 5: should we use `TinyVec`? I mean, maybe we will never have >8 params, but it would be nice to be able to do something about that.

Or maybe the macro can intern params 8+ as a tuple, and you just accept some perf loss in that case

ch1ffa: Let's intern a tuple in that case. When we integrate it to the compiler, I think, it would be easier to measure the impact of having tinyvec here or something else.

rbalicki2: I don't consider that a blocker for merging to master though fwiw

crates/pico/src/derived_node.rs
```diff
@@ -48,3 +62,52 @@ pub struct DerivedNodeRevision {
...
 }
+
+#[derive(Debug, Clone, Copy)]
+pub struct MemoRef<'db, T> {
+    db: &'db Database,
+    derived_node_id: DerivedNodeId,
+    phantom: PhantomData<T>,
+}
+
+impl<'db, T: 'static + Clone> MemoRef<'db, T> {
```
rbalicki2, line 73: is everything going to be `T: 'static`?

For example, what if we have
```
#[memo]
fn get_client_field() -> ClientField

#[memo]
fn get_client_field_selections() -> &ClientFieldSelections
```

Returning a subset or view into another object will probably be a common pattern, because this is what will allow us to avoid unnecessary recalculation. e.g. if the AST changed, but not in a way that affected the selections, we would need something like this to avoid doing a bunch of extra work

If this is the case, are we forced to keep clones of the `selection_set` redundantly in the database?

crates/pico_macros/src/memo.rs
rbalicki2: Oh I see, then it makes sense to use `init_param_vec`.

crates/pico_macros/src/memo.rs
```diff
@@ -97,56 +90,97 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                        ty
+                    };
+                    let binding_expr = match **ty {
+                        syn::Type::Reference(_) => quote!(inner),
+                        _ => quote!(inner.clone()),
+                    };
+                    quote! {
+                        let #arg = {
+                            let param_ref = ::pico::macro_fns::get_param(#db_arg, derived_node_id.params[#i])?;
+                            let inner = param_ref.downcast_ref::<#target_type>()?;
```
rbalicki2, line 162: We should probably panic here, instead of returning `None`

crates/pico_macros/src/memo.rs
```diff
@@ -97,56 +90,97 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
...
     }
 
-    let unpacked_args = other_args.clone();
-    let inner_args = other_args.clone();
+    let extract_parameters = other_args.clone().zip(argument_types.clone())
+        .enumerate()
+        .map(|(i, (arg, ty))| {
+            match ArgType::parse(ty) {
+                ArgType::Source => {
+                    let maybe_ref = if matches!(**ty, syn::Type::Reference(_)) {
```
rbalicki2, line 124: tbh it doesn't make sense to ever pass a reference to a `Source` or `MemoRef`, since they implement `Copy`

This is more of a comment than a request to change anything

crates/pico_macros/src/memo.rs
```diff
@@ -97,56 +90,97 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                        &reference.elem
+                    } else {
+                        ty
+                    };
+                    let binding_expr = match **ty {
+                        syn::Type::Reference(_) => quote!(inner),
+                        _ => quote!(inner.clone()),
+                    };
+                    quote! {
+                        let #arg = {
```
rbalicki2, line 160: Can we have `let #arg: #ty` for consistency here?

crates/pico_macros/src/memo.rs
```diff
@@ -160,3 +194,76 @@ fn hash(input: &Signature) -> u64 {
...
+
+    if let syn::Type::Path(type_path) = inner {
+        if let Some(segment) = type_path.path.segments.last() {
+            return segment.ident == target;
+        }
+    }
+    false
+}
+
+fn get_fn_lifetime(sig: &mut Signature) -> Lifetime {
```
rbalicki2, line 231: nit: `get_or_create_fn_lifetime` or something?

crates/pico_macros/src/memo.rs
```diff
@@ -160,3 +194,76 @@ fn hash(input: &Signature) -> u64 {
...
+        sig.generics.params.push(parse_quote!('db));
+    }
+    lt.unwrap_or(parse_quote!('db))
+}
+
+fn check_db_lifetime(arg: &mut FnArg, lifetime: Lifetime) -> Result<(), syn::Error> {
+    if let FnArg::Typed(PatType { ty, .. }) = arg {
+        if let syn::Type::Reference(ref mut reference) = **ty {
+            if let Some(db_lifetime) = &reference.lifetime {
+                if db_lifetime != &lifetime {
```
rbalicki2, line 249: Should this fail to compile?

```
fn parse_ast<'foo>(db: &Database, id: SourceId<Input>, bar: &'foo ()) -> Result<Program> {
```

it does not

crates/pico/src/memo.rs
```diff
@@ -98,22 +98,30 @@ fn update_derived_node(
...
     };
-    db.insert_derived_node(
-        derived_node_id,
-        DerivedNode {
-            dependencies: tracked_dependencies.dependencies,
-            inner_fn,
-            value,
-        },
-    );
     (tracked_dependencies.max_time_updated, did_recalculate)
```
rbalicki2, line 125: If we receive `error`, `max_time_updated` is `Epoch::new()`. That seems like a sane default. But we should think about whether that's true.

crates/pico_macros/src/memo.rs
```diff
@@ -97,56 +90,97 @@ pub(crate) fn memo(args: TokenStream, item: TokenStream) -> TokenStream {
...
-                    (
-                        #(#reconstructed_args,)*
-                    )
-                };
+            let mut param_ids = ::pico::macro_fns::init_param_vec();
+            #(
+                #param_ids_blocks
+            )*
+            let derived_node_id = ::pico::DerivedNodeId::new(#fn_hash.into(), param_ids);
+            ::pico::memo(#db_arg, derived_node_id, ::pico::InnerFn::new(|#db_arg, derived_node_id| {
```
rbalicki2, line 178: perhaps we should add `debug_assert!(!matches!(memo_result, DidRecalculate::Error))`

crates/pico_macros/src/memo.rs
ch1ffa: Agreed, but what should we do? Accept references or throw a compilation error?

crates/pico_macros/src/memo.rs
ch1ffa: Macro assigns `'foo` to the Database in that case. Tbh I'm not sure what else we need to cover. My case is I need the same lifetime for function itself, db and MemoRef. I'll add `'db` or take the first lifetime from the function and assign it to the db (or check if this lifetime has been assigned by a user)

crates/pico/src/memo.rs
ch1ffa: It makes sense even if it does not change overall execution. If we return error and register with wrong time_updated, we will call parent `inner_fn` which will call this memo again and will re-register with correct time_updated.

crates/pico/src/memo.rs
```diff
@@ -156,16 +161,19 @@ fn derived_node_changed_since(db: &Database, derived_node_id: DerivedNodeId, sin
...
+    )
 }
 
 fn with_dependency_tracking(
     db: &Database,
-    param_id: ParamId,
+    derived_node_id: DerivedNodeId,
     inner_fn: InnerFn,
-) -> (Box<dyn DynEq>, TrackedDependencies) {
+) -> (Option<Box<dyn DynEq>>, TrackedDependencies) {
```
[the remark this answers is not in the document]
ch1ffa: Yeah, wrap everything to `Option` is better. Dependencies could be not empty if we got `None` from some other than the very first dependency.

crates/pico/src/derived_node.rs
ch1ffa: I'll do it as a follow up PR with additional test.

crates/pico_macros/src/memo.rs
rbalicki2: accept haha, just talking to myself. Maybe warn in debug mode?

crates/pico/src/memo.rs
```diff
@@ -156,16 +162,17 @@ fn derived_node_changed_since(db: &Database, derived_node_id: DerivedNodeId, sin
         return true;
     };
     let did_recalculate = memo(db, derived_node_id, inner_fn);
-    matches!(did_recalculate, DidRecalculate::Recalculated)
+    matches!(
+        did_recalculate,
+        DidRecalculate::Recalculated | DidRecalculate::Error
```
rbalicki2, line 167: I think what's happening is that if any nested memoized function fails with `DidCalculate::Error`, then that propagates all the way up to whatever top-level function we have.

So that's good

rbalicki2 (approved):
LGTM shipit

PR #379, opened 2025-02-11
create upadtable proxy and apply updates

PatrykWalach:
I've had to copy a bunch of code from `readData` to map the data back to the store.

- This doesn't support arrays, I have not done this part on rust part yet.
- I also need to generate `isUpdatable` for each node.

Is`Pet.PetTaglineCard` broken? The tagline doesn't refresh after mutation.

Part of updatable fields #312

### DIFF
--- a/demos/pet-demo/src/components/PetTaglineCard.tsx
+++ b/demos/pet-demo/src/components/PetTaglineCard.tsx
@@ -22,7 +22,7 @@ export const PetTaglineCard = iso(`
       <CardContent>
         <h2>Tagline</h2>
         <p>&quot;{pet.tagline}&quot;</p>
-        {mutationRef === null ? (
+        {mutationRef === null || pet.tagline !== 'SUPER DOG' ? (
           <Button
             onClick={() => {
               loadMutation({
--- a/demos/pet-demo/src/components/PetUpdater.tsx
+++ b/demos/pet-demo/src/components/PetUpdater.tsx
@@ -29,9 +29,6 @@ export const PetUpdater = iso(`
   // We should find a way to work around this.
 
   const updateTagline = () => {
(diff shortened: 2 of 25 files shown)


PatrykWalach:
for objects I will have to traverse normalization ast at the same time so we can call `normalizeData` on the set object, and then use the `Update` to set key to the `Link`
```tsx
startUpdate(data=>{
  data.friend = { id: 12, name: "Friend" }
})
```

PatrykWalach:
Right now I use an `Update[]` as a result of updater,
but I'm thinking of changing it to be the same type as `IsographStore`
then I will create a function that swaps between these two stores.

It should be a bit faster than an array of updates, because updates are batched per `Link`

e.g.
```tsx

const mainStore = {
  Query: {
    __ROOT: {
      pet: { __typename: 'Pet', id: 1 },
    },
  },
  Pet: {
    1: {
      id: 1,
      name: 'Original name',
      species: 'Species',
    },
  },
};

const updateStore = {
  Pet: {
    1: {
      name: 'New name',
    },
  },
};

// when
swap(mainStore, updateStore);

// then
expect(mainStore).toEqual({
  Query: {
    __ROOT: {
      pet: { __typename: 'Pet', id: 1 },
    },
  },
  Pet: {
    1: {
      id: 1,
      name: 'New name',
      species: 'Species',
    },
  },
});

expect(updateStore).toEqual({
  Pet: {
    1: {
      name: 'Original name',
    },
  },
});

```
After swapping, `updateStore` becomes a store that can be sued to restore previous values in case of optimisitcUpdates.

and whenever we `normalizeData` from network request (or when we `startUpdate` on an object) we would also create a store of previous values.

Let me know what you think about that

PR #386, opened 2025-02-12
generate undefined for optional input fields

### LINKED ISSUE #280: Generate `foo?: Foo | undefined | null` for optional fields, instead of `foo: Foo | null`

PatrykWalach:
Closes #280


rbalicki2 (approved):
lgtm shipit

PR #391, opened 2025-02-16
coerce literals when typechecking

### LINKED ISSUE #383: [bug] Accept (string? int?) scalar literals for ID!

PatrykWalach:
FIxes #383 and allows Int literals for Float arguments

### DIFF
--- a/crates/isograph_schema/src/validate_argument_types.rs
+++ b/crates/isograph_schema/src/validate_argument_types.rs
@@ -157,7 +157,25 @@ pub fn value_satisfies_type(
             field_argument_definition_type,
             schema_data,
             selection_supplied_argument_value.location,
-        ),
+        )
+        .or_else(|error| {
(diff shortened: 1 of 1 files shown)


crates/isograph_schema/src/validate_argument_types.rs
```diff
@@ -157,7 +157,25 @@ pub fn value_satisfies_type(
...
-        ),
+        )
+        .or_else(|error| {
+            scalar_literal_satisfies_type(
+                &schema_data.float_type_id,
+                field_argument_definition_type,
+                schema_data,
+                selection_supplied_argument_value.location,
+            )
+            .map_err(|_| error)
```
PatrykWalach, line 168: This keeps original error message "Expected Type found String" instead of "Expected Type found Float"

rbalicki2:
lgtm shipit

Issue #402, opened 2025-02-19
Pico garbage collection


rbalicki2:
I will sketch out my idea for pico garbage collection here.

## Why generational GC is a poor fit

- During the initial compilation, everything goes into generation one
- If there is at least one item that never changes, e.g. if you never modify a particular file with an iso literal, then the first generation will never be GC'ed
- But it's the largest generation :/
- Plus, everything is held in a Box, so the underlying array is going to be quite compact anyway. What we really care about is GC'ing what's held in the Box.

## Alternative: user-driven GC

- I think we have the ability to run GC periodically, whenever the user is inactive for e.g. 10 seconds
- In practice, we might want to GC in other situations, too, like when the number of retained items is too high, but anyway.

## Implementation

- We don't want to require &mut db to call a memoized function, and (to make the lifetimes work), we need to give out &'static references to the derived nodes (this is based on my recollection. So this bullet might be wrong on some details.)
- So, a single BoxcarVec is a good fit for the derived node storage
- We keep an LRU cache of the top-level fields that were called.
  - In practice, we could keep a vec of top-level calls, and only update the LRU when we GC, so that the hot path is fast.
  - We should also allow users to retain top level queries and manually prevent them from being GC'ed, but that isn't part of v1. The reason we want this is that we're going to have many top-level "hover on this character" requests, so we're very quickly going to kick out e.g. the validated schema.
- When we GC, we traverse the retained top-level queries, keeping track of the reachable derived nodes. Then, we create a new boxcar vec with the retained items. Then, we update `derived_node_id_to_index` to include the new index.

## Misc

- After we GC, every MemoRef is invalidated. So, the MemoRef's should contain a gcEpoch, and panic if dereferenced after we've GC'ed.
- So, in practice, you should never hold onto a MemoRef! If you want to store it elsewhere, deref and clone it.
- The MemoRef has a lifetime parameter of `'db` (i.e. the lifetime of the db), so in practice you can't hold onto it, and then later call `.run_garbage_collection()` (which takes `&mut self)`.

## Consider

- Can we GC sources? Probably not. But it would be nice

PR #403, opened 2025-02-19
Pico GC [1/n] — refactor storage

rbalicki2:
* Refactor Database to store the derived nodes and params each in a single `boxcar::Vec`, getting rid of generations and arenas.
* Like last time, whenever we create a derived node value, or encounter a new value, we push that node onto the `derived_nodes` vec

## Next steps

* Add a `gc_epoch` field on `Database` and a `valid_during_gc_epoch` field on `MemoRef`.
* Implement GC as described in #402 

## Additional refactor: use `entry` everywhere

* I think we can refactor `memo` to use `entry`. Right now, that deadlocks because `update_derived_node` attempts to read the same entry from the map. But, we can pass an `OccupiedEntry` to `update_derived_node` and handle that.
* We are currently assuming that the memoize functions form a DAG (i.e. have no loops). In the above world, if we had a loop, then calling `inner_fn` would deadlock. We can replace `.entry` with `try_entry`, and panic if we encounter a deadlock instead.
* I've gone ahead and removed a bunch of functions as part of this. My intention isn't to keep it this way, but to add back some helpful intermediate functions after we use the `try_entry` API everywhere.

## Testing

* Tests pass

Part of #402


ch1ffa (approved):
sgtm

PR #404, opened 2025-02-20
test linked fields updates

PatrykWalach:
I would like to go ahead with updatable fields but there's a bunch of stuff I'm not sure about.

Let me know if I'm on a right track or not

Part of #312

### DIFF
--- a/crates/generate_artifacts/src/eager_reader_artifact.rs
+++ b/crates/generate_artifacts/src/eager_reader_artifact.rs
@@ -12,10 +12,10 @@ use std::{borrow::Cow, collections::BTreeSet, path::PathBuf};
 
 use crate::{
     generate_artifacts::{
-        generate_client_field_parameter_type, generate_output_type, generate_parameters,
-        ClientFieldFunctionImportStatement, RESOLVER_OUTPUT_TYPE, RESOLVER_OUTPUT_TYPE_FILE_NAME,
-        RESOLVER_PARAMETERS_TYPE_FILE_NAME, RESOLVER_PARAM_TYPE, RESOLVER_PARAM_TYPE_FILE_NAME,
-        RESOLVER_READER_FILE_NAME,
+        generate_client_field_parameter_type, generate_client_field_updatable_data_type,
+        generate_output_type, generate_parameters, ClientFieldFunctionImportStatement,
+        RESOLVER_OUTPUT_TYPE, RESOLVER_OUTPUT_TYPE_FILE_NAME, RESOLVER_PARAMETERS_TYPE_FILE_NAME,
+        RESOLVER_PARAM_TYPE, RESOLVER_PARAM_TYPE_FILE_NAME, RESOLVER_READER_FILE_NAME,
     },
     import_statements::{
         param_type_imports_to_import_param_statement, param_type_imports_to_import_statement,
@@ -208,7 +208,16 @@ pub(crate) fn generate_eager_reader_param_type_artifact<TOutputFormat: OutputFor
     let mut loadable_fields = BTreeSet::new();
     let mut link_fields = false;
     let mut updatable_fields = false;
-    let (client_field_parameter_type, updatable_data_type) = generate_client_field_parameter_type(
+    let client_field_parameter_type = generate_client_field_parameter_type(
+        schema,
+        client_field.selection_set_for_parent_query(),
+        parent_type,
+        &mut param_type_imports,
+        &mut loadable_fields,
+        1,
+        &mut link_fields,
+    );
+    let updatable_data_type = generate_client_field_updatable_data_type(
         schema,
         client_field.selection_set_for_parent_query(),
         parent_type,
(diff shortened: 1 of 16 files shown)


PatrykWalach:
# More test cases

## 1. We should validate that `@updatable` linked field has `id` selected whenever possible

## 2. We need to decide how to handle abstract types:

It should be impossible to set two concrete types at the same time

```tsx
startUpdate((data) => {
  data.node = {
    asFoo: {
      id: 12,
    },
    asBar: {
      id: 14,
    },
  };
});
```

### 1. If we have an @updatable interface we have to:

- [] validate that `@updatable` linked field has `__typename`
- [] generated setter accepting a `discriminated union`, we will inline fields from each `asConcreteType` field

This is great because we can update common fields on the interface.

- when we want to set a single concrete type we can mark `asConcreteType` with `@updatable`
- when we don't know the type we can mark the parent abstract type as `@updatable`

PatrykWalach:
### 3. When we write linked/scalar field `client fields need to reevaluate`
```tsx
startUpdate((data) => {
  data.firstName = "Hello"
  data.secondName = "World"
  expect(data.fullName).toBe("Hello World")
})
```

rbalicki2:
A grab bag of responses

# What data to read
- if you read from updatableData, that needs to calculate the field value at the time of the read
- that's because you can assign `user.best_friend = { id: "4", __typename }` (or something), and then read `user.name`. That `name` will come from the the user with id `4`
- and of course, you can assign `userWithId4.name = "Lizard"` and in the above you would read that
- So, you cannot materialize more than what was actually read out. e.g. `{ best_friend: { name }, worst_enemy: { name }}`, if `best_friend` and `worst_enemy` were the same, if you assigned `worst_enemy.name = "You"` and read `best_friend.name`, you should get the same value `"You"`.

# Assignment and multiple stores and reading impl
- Let's say the original user had `id: 1`
- Assigning a field either mutates an object *or* adds the assignment to a list. I think they're the same, so let's go with "mutates an object"
- This object is separate from the store
- So in the above case, the store is unchanged, and the new object ("optimistic layer", for lack of a better term at the moment) is `{ User: { 1: { bestFriend: { __ref: { id: 4, __typename: "User" }} }, { 4: name: "Lizard" } }}`
- This is coincidentally identical to how optimistic updates can work
- Now, when you read, you first read through the optimistic layer. If the field is not found, try to read through the base layer.
  - This is only true for the underlying records. Resolvers and pointers are not assignable! But they, when read, are executed, and *also* read in the same way.
  - So, everything sees a consistent view of the world.

# Reading and assignment examples

## Example 1: raw data
```js
{
  best_friend @updatable { name @updatable }
}

const user1 = data.best_friend;
data.best_friend = someoneElse;
user1.name = 'Lizard'; // this should work, and it should effectively be
const {id, __typename} = user1.id;
store[typename][id].name = "Lizard"

// it should not be
const {id, __typename} = data;
const {id: id2, __typename: __typename2} = store[__typename][id].user
store[__typename2][id2] = 'Lizard'

// because then the above code would be broken
```

## Example 2: client field

```js
// If we have a client field on User called superName:
(data) => 'Super' + data.name

// then superName's value depends on when it was read: 
// SuperMark or SuperLizard
```

- This answers: `3. When we update linked/scalar fields client fields need to reevaluate`

# Aliases, arguments

- Aliases and arguments don't show up in the updatableData, so we have to go back to the reader AST whenever we read or write and figure out what field we read from
- It seems conceptually cleaner to isntead go from reader AST to fields, i.e. to assign a bunch of getters and setters using `Object.defineProperty`. But it could also be done with a proxy. But I would definitely prefer the getters/setters.

# What is and isn't assignable
- typename and id should never be assignable. There should be a `assignability` value on each server field.
- client pointers and client fields are not assignable (in their current state)
- instead of assigning `isFoo`, you should reassign the parent field `newsfeed.first_item = { typename: 'AdItem', id }` or whatever
- I think you should always have to assign one field at a time. (See "assigning an object vs assigning a link" below)

# A rant about synthetic IDs that you can completely ignore
- synthetic IDs are a mistake. If you have `event.date` which has `{ month, day }` but not `id`, then you shouldn't be able to do `event1.date = event2.date`. But if you do, it should automatically deep copy it. Or something.
- It's very unclear what should happen if you do `event1.date = event2.date; event2.date.month = 'Smarch'; console.log(event1.date.month)`. What if the object cannot be deep copied? (e.g. contains a DOM element or function)
- This is hard. We can revisit and make this work properly later.
- So yes, we should only allow `@updatable` linked fields if they have `id`s. That's a good idea for now.

# Overall workflow
- After the function call completes, we write the data to the store (i.e. merge the object with the parent one), and call all affected subscriptions, and everything works
- So, in that function, we don't expect any other fields' values to update

# Assigning an object vs assigning a link
- I think the setter should always just accept a link. I don't know if this is accurate syntax

```ts
type foo = {
  get best_friend(): UserBestFriend__output__type;
  set bar(v: Link | null);
}
```
- What I don't like about this is that you now have potentially missing fields, and thus infinite suspense. But I think that's okay, it's a footgun for now, but we can fix it later.
- I would like some way to enforce that the linked object has certain fields if they're required in the destination! But anyway, that's a giant issue and doesn't need to be handled now.

# We would need to generate more accurate data

- We should do that! I think with the above suggestions, we don't need that at the moment, though, since you can only assign the parent link.

PR #407, opened 2025-02-20
fix clippy issues

PatrykWalach:


rbalicki2:
nice, thanks for handling this

Issue #408, opened 2025-02-20
Pico - multithreadedness


rbalicki2:
I have an inkling that separating the dependency stack from the storage will make something better. e.g. if the API was

```
let storage = create_storage(); // this is Sync
let stack = storage.create_stack(); // you can call this many times. It contains a reference to storage
memoized_function(stack); // this takes &mut stack

// since its Send you can
tokio::spawn(|| {
  let stack = &storage.create_stack();
  memoized_function(stack)
})
```

I think the stack is the only thing we care about being in one thread.

If everything is a DAG, deadlocks are avoidable, e.g. in this case:

```
fn a(db_stack) {
  first(db_stack)
  second(db_stack)
}
fn b(db_stack) {
  second(db_stack)
  first(db_stack)
}
```

Both a and b could just wait for the other to complete `first` or `second`. And that's optimal behavior, anyway.

And you can't modify a source without `&mut`, so there can't be a dependency stack created in that case.

------

Anyway, even before we do multi-threadedness, this seems to clean things up conceptually.

PR #409, opened 2025-02-20
Split storage and dependency stack into separate structs

rbalicki2:
This does not affect the API, this is purely shuffling things arounds to be little cleaner and to prepare us for #408.

### DIFF
--- a/crates/pico/src/database.rs
+++ b/crates/pico/src/database.rs
@@ -16,6 +16,11 @@ use crate::derived_node::{DerivedNode, DerivedNodeId, DerivedNodeRevision};
 #[derive(Debug)]
 pub struct Database {
     pub(crate) dependency_stack: DependencyStack,
+    pub(crate) storage: DatabaseStorage,
+}
+
+#[derive(Debug)]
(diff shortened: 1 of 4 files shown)


crates/pico/src/database.rs
```diff
@@ -16,6 +16,11 @@ use crate::derived_node::{DerivedNode, DerivedNodeId, DerivedNodeRevision};
 #[derive(Debug)]
 pub struct Database {
     pub(crate) dependency_stack: DependencyStack,
```
ch1ffa, line 18: Yeah, I like it. My initial thought about multithreading is that we should initialise the dependency stack from a top-level query or after Database cloning (assuming every top-level query is going to be in a separate thread). So even if `DependencyStack` may be something different in the future it is good to keep it separately from the storage.

ch1ffa (approved):
lgtm!

crates/pico/src/database.rs
rbalicki2: Yes, I'm very much on board with this

PR #410, opened 2025-02-21
Pico gc [2/3?]: implement garbage collection

rbalicki2:
# Add garbage collection to pico

## When top-level memoized functions are called
- Keep track of the invocation (uniquely identified by its DerivedNodeId) and insert that into a boxcar::Vec (top_level_calls)

## If we run garbage collection:
- assert no memoized function is in the process of being called
- transfer the top_level_calls into an LRU cache
- at this point, for each top level node in the LRU cache, keep each derived node reachable from that top-level node (including the node itself), as well as all reachable params

## GC
- Recreate derived_nodes, params, param_id_to_index and derived_node_id_to_revision
- Basically, we have three goals:
    - Without cloning, push all params and derived node values onto new vectors
    - Update indexes (in param_id_to_index and derived_node_id_to_revision; revisions contain indexes)
    - Drop all non-reachable values
- The best way to achieve these goals is to create new data structures

## Testing
- I added some basic testing. But we can add more, especially related to whether params are cloned.

## Next steps
- Add a `MemoRef::retain()` method


ch1ffa (approved):
Awesome!

PR #411, opened 2025-02-21
Pico: assert dependency stack is empty when calling db.set or db.remove

rbalicki2:


crates/pico/src/database.rs
```diff
@@ -74,12 +74,21 @@ impl Database {
...
         self.storage.set_source(source)
     }
 
     pub fn remove<T>(&mut self, id: SourceId<T>) {
+        self.assert_empty_dependency_stack();
         self.storage.remove_source(id)
     }
+
+    fn assert_empty_dependency_stack(&self) {
+        assert!(
```
ch1ffa, line 87: Let's keep it like this for now. I don't have a big picture yet how exactly we are going to use pico in the multithreaded case: should it be a task queue or a cancelation of  in-flight requests, so maybe this `assert` is going to be something like `anyhow::assert` and we should return error here. Anyway, will see.

ch1ffa (approved):
lgtm

PR #412, opened 2025-02-21
Add "retained queries" to the database

rbalicki2:
# Retained queries

- You can pass a MemoRef to db.retain, and get a RetainedQuery
- While this retained query exists, the derived node related to the invocation of that memoized function + any of its dependencies + any reachable params will not be GC'ed
- The retained query will panic if dropped. Instead, it must be passed to db.clear_retain or you can call MemoRef::permanently_retain_query.

## Naming

- I'm not a fan of the naming, and want to take a pass at that later

## Testing

- Some basic testing. I think we can do more testing: assert that it panics, etc.
- We can also assert that a retained query can also be part of the LRU cache


crates/pico/src/retained_query.rs
```diff
@@ -0,0 +1,75 @@
...
+                (*occupied_entry.get_mut()) -= 1;
+                if occupied_entry.get() == &0 {
+                    occupied_entry.remove();
+                }
+            }
+            Entry::Vacant(_) => {
+                panic!("RetainedQuery not found in databse. This is indicative of a bug in Pico.")
+            }
+        }
+        retained_query.cleared = true;
```
rbalicki2, line 73: TBH we could make this more efficient by calling `std::mem::forget`, and then doing the "panic in a const context" trick in the Drop impl... LOL.

crates/pico/src/retained_query.rs
```diff
@@ -0,0 +1,75 @@
...
+        self.cleared = true;
+    }
+}
+
+impl std::ops::Drop for RetainedQuery {
+    fn drop(&mut self) {
+        if !self.cleared {
+            panic!(
+                "RetainedQuery dropped while still retained. If this is intentional, \
+                consume the RetainedQuery by calling retained_query.forget()."
```
ch1ffa, line 38: we have `clear_retain` method but not `forget`

rbalicki2: thanks

PR #419, opened 2025-02-22
split readData

PatrykWalach:
I need to split `readData` so I can reuse reading in `startUpdate`

### DIFF
--- a/libs/isograph-react/src/core/read.ts
+++ b/libs/isograph-react/src/core/read.ts
@@ -14,7 +14,6 @@ import {
 } from './entrypoint';
 import {
   ExtractData,
-  ExtractParameters,
   FragmentReference,
   Variables,
   type UnknownTReadFromStore,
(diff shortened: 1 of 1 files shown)


PatrykWalach:
I have to split scalar/linked to make minimal `startUpdate`, then I can split the rest

libs/isograph-react/src/core/read.ts
```diff
@@ -117,12 +122,13 @@ export function readButDoNotEvaluate<
   }
 }
 
-export type ReadDataResult<TReadFromStore> =
-  | {
-      readonly kind: 'Success';
-      readonly data: ExtractData<TReadFromStore>;
-      readonly encounteredRecords: EncounteredIds;
```
rbalicki2, line 124: good catch that this isn't needed

rbalicki2 (approved):
LGTM shipit, this probably should've been done a long time ago haha, read is way too large of a function

PR #423, opened 2025-02-23
create optimistic proxy

PatrykWalach:


libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,145 @@
...
+        return optimisticStoreRecord;
+      }
+
+      const storeRecord = recordsById[p];
+
+      if (storeRecord == null) {
+        return storeRecord;
+      }
+
+      optimisticStoreRecord = target[p] ??= {};
```
PatrykWalach, line 57: This is a part that's bugging me, if we have a `record` in original store we create a new record in `optimisticLayer` just in case we need to write to it.

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -134,8 +137,11 @@ export function createIsographEnvironment(
   logFunction?.({
     kind: 'EnvironmentCreated',
   });
+  let optimisticLayer: OptimisticLayer = {};
```
rbalicki2, line 140: Rust brain has gotten to you
Also this should be caught by eslint?? (Also below)

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,124 @@
...
+
+      if (patch == null || data == null) {
+        recordById[recordId] = patch;
+      } else {
+        Object.assign(data, patch);
+      }
+    }
+  }
+
+  resetOptimisticLayer(environment);
```
rbalicki2, line 117: can this just be `environment.optimisticLayer = {}` or something?

rbalicki2 (changes requested):

rbalicki2:
Okay, I thought about this more deeply. Take a look at #427

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: I thought it shouldn't because then I have to create new proxy, but maybe it should because otherwise the old proxies point to the records that's been deleted, thoughts?

PatrykWalach: `startUpdate` run inside `startUpdate` would surface this issue.
```tsx
startUpdate((data) => {
  startUpdate(()=>{}) // this is merging old changes and assigning new object to `optimisticLayer` 

  data // this still points to the old layer 
})

rbalicki2:
closing for now

PR #428, opened 2025-02-24
Start update

PatrykWalach:
Part of #312

### DIFF
--- a/libs/isograph-react/src/core/startUpdate.ts
+++ b/libs/isograph-react/src/core/startUpdate.ts
@@ -1,28 +1,264 @@
 import {
+  callSubscriptions,
+  getParentRecordKey,
+  insertIfNotExists,
+  type EncounteredIds,
+} from './cache';
+import type { RefetchQueryNormalizationArtifactWrapper } from './entrypoint';
+import {
   stableIdForFragmentReference,
+  type ExtractParameters,
   type ExtractStartUpdate,
+  type ExtractUpdatableData,
   type FragmentReference,
   type UnknownTReadFromStore,
 } from './FragmentReference';
-import type { IsographEnvironment } from './IsographEnvironment';
+import {
+  assertLink,
+  type IsographEnvironment,
+  type Link,
+} from './IsographEnvironment';
+import { readPromise, type PromiseWrapper } from './PromiseWrapper';
+import {
+  readLinkedFieldData,
+  readScalarFieldData,
+  type NetworkRequestReaderOptions,
+  type ReadDataResultSuccess,
+} from './read';
+import type { ReaderAst } from './reader';
 
 export function getOrCreateCachedStartUpdate<
   TReadFromStore extends UnknownTReadFromStore,
 >(
   environment: IsographEnvironment,
-  fragmentReference: FragmentReference<TReadFromStore, any>,
+  fragmentReference: FragmentReference<TReadFromStore, unknown>,
   eagerResolverName: string,
+  networkRequestOptions: NetworkRequestReaderOptions,
 ): ExtractStartUpdate<TReadFromStore> {
   return (environment.eagerReaderCache[
     stableIdForFragmentReference(fragmentReference, eagerResolverName)
-  ] ??= createStartUpdate(environment, fragmentReference));
+  ] ??= createStartUpdate(
+    environment,
+    fragmentReference,
+    networkRequestOptions,
+  ));
 }
 
 export function createStartUpdate<TReadFromStore extends UnknownTReadFromStore>(
-  _environment: IsographEnvironment,
-  _fragmentReference: FragmentReference<TReadFromStore, any>,
+  environment: IsographEnvironment,
+  fragmentReference: FragmentReference<TReadFromStore, unknown>,
+  networkRequestOptions: NetworkRequestReaderOptions,
 ): ExtractStartUpdate<TReadFromStore> {
-  return (_updater) => {
-    // TODO start update
+  return (updater) => {
+    let mutableUpdatedIds: EncounteredIds = new Map();
+
+    let data = createUpdatableProxy(
+      environment,
+      fragmentReference,
+      networkRequestOptions,
+      mutableUpdatedIds,
+    );
+
+    updater(data);
+
+    callSubscriptions(environment, mutableUpdatedIds);
+  };
+}
+
+export function createUpdatableProxy<
+  TReadFromStore extends UnknownTReadFromStore,
+>(
+  environment: IsographEnvironment,
+  fragmentReference: FragmentReference<TReadFromStore, unknown>,
+  networkRequestOptions: NetworkRequestReaderOptions,
+  mutableUpdatedIds: EncounteredIds,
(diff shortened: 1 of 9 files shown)


libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+) {
+  let value = get();
+  let lastInvalidated = Date.now();
+  Object.defineProperty(target, property, {
+    configurable: true,
+    enumerable: true,
+    get: () => {
+      if (lastInvalidated <= mutableState.lastInvalidated) {
+        value = get();
+        lastInvalidated = Date.now();
```
rbalicki2: can we use an incrementing counter, for consistency with pico?

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+
+function readUpdatableData<TReadFromStore extends UnknownTReadFromStore>(
+  environment: IsographEnvironment,
+  ast: ReaderAst<TReadFromStore>,
+  root: Link,
+  variables: ExtractParameters<TReadFromStore>,
+  nestedRefetchQueries: RefetchQueryNormalizationArtifactWrapper[],
+  networkRequest: PromiseWrapper<void, any>,
+  networkRequestOptions: NetworkRequestReaderOptions,
+  mutableState: {
```
rbalicki2: let's extract this into a type alias

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+    configurable: true,
+    enumerable: true,
+    get: () => {
+      if (lastInvalidated <= mutableState.lastInvalidated) {
+        value = get();
+        lastInvalidated = Date.now();
+      }
+      return value;
+    },
+    ...(set && {
```
rbalicki2, line 137: TIL we don't need `updatable: true`

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+      lastInvalidated: number;
+    };
+    get(): any;
+    set?(v: any): void;
+  },
+) {
+  let value = get();
+  let lastInvalidated = Date.now();
+  Object.defineProperty(target, property, {
+    configurable: true,
```
rbalicki2: `configurable: false`, I would imagine?

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+    mutableState,
+  }: {
+    mutableState: {
+      lastInvalidated: number;
+    };
+    get(): any;
+    set?(v: any): void;
+  },
+) {
+  let value = get();
```
rbalicki2: Can we make `value` have type `{ kind: 'NotSet' } | { kind: 'Set', value: T, validatedAt: number }`?

That way we don't have to fully materialize the tree until we actually access a value. (Besides being a nice-to-have, it might actually make a difference if we have a very large fragment, e.g. containing a list with many items)

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+  ).data;
+}
+
+function defineCachedProperty<T>(
+  target: T,
+  property: PropertyKey,
+  {
+    get,
+    set,
+    mutableState,
```
rbalicki2: Let's either make this function have a single parameter (an object with named fields) or five params. I don't like the half-way split here

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+              environment,
+              field,
+              root,
+              variables,
+              nestedRefetchQueries,
+              networkRequest,
+              networkRequestOptions,
+              new Map(),
+            );
+            if (data.kind === 'MissingData') {
```
rbalicki2: do we risk throwing if we assign something like `pet.bestFriend = { id: '0', __typename: 'Pet' }` (or whatever we would have to do) and fields on the new best friend are missing?

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -1,28 +1,323 @@
...
+  environment: IsographEnvironment,
+  fragmentReference: FragmentReference<TReadFromStore, unknown>,
+  networkRequestOptions: NetworkRequestReaderOptions,
 ): ExtractStartUpdate<TReadFromStore> {
-  return (_updater) => {
-    // TODO start update
+  return (updater) => {
+    let mutableUpdatedIds: EncounteredIds = new Map();
+
+    let data = createUpdatableProxy(
```
rbalicki2, line 58: `createUpdatableProxy` can throw, e.g. if there is missing data. And `updater` can throw, too, if due to a change there is missing data. Or because the user wrote some dumb code.

So, we should probably call `callSubscriptions` in a `finally` block for now, and re-throw the error (?). It's an error that probably won't be caught anywhere! But at least users can try to handle it themselves, if they want to e.g. show a toast.

In the long term, I would prefer:

- we write data to a new store layer (e.g. an optimistic layer), and read through that layer. That simplifies things a bit, e.g. no need for the `lastInvalidated` value.
- we merge that layer into the store when the update completes, and trigger subscriptions
- if we throw, we simply throw away that new optimistic layer.

This isn't blocking.

rbalicki2: This "new layer -> merge -> trigger subscriptions" is probably a general pattern, as well. We could, for example, write network responses to the store in that way. (Though not obviously clear what benefit that gives us).

rbalicki2 (commented):
I'm really, really excited about this. Just a few nits (for now), and let's fucking shipit

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: done

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: done

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: done

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: done five params

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: Yes, if you read `pet.bestFriend` after setting, it will throw

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: all `lastInvalidated` does is cache the getters, there should be no difference if we use extra layer or not

rbalicki2 (approved):
shipitttttt hell yeah!!!!

PR #432, opened 2025-02-24
add singleton macro

### LINKED ISSUE #414: Pico - make #[key] optional (or conditional)
Some type of sources are singletons, e.g. `CompilerConfig`. Adding a special field that represents a unique key is not necessary for such structs - it is enough to use the struct name as a key. 

Consider making `#[key]` attribute optional or add a parameter to the `Source` macro that does not require using the `key` attribute.

ch1ffa:
Closes #414 by introducing `Singleton` macro that uses `TypeId` of a struct as `Key`.


rbalicki2 (approved):
👏 👏 👏 👏 👏  shipit

PR #451, opened 2025-03-02
intern memoref

ch1ffa:
This PR explores how to use Pico with `Result` return type.

A memoized function puts its return value to the storage as `Box<dyn Any>` so we can't easily use error conversion by `?` or `map_err` which is a common pattern in the isograph compiler code. To achieve it we need to store `MemoRef` in the database using a pinch of `unsafe` code and explore a good pattern how to map errors properly.

### DIFF
--- a/crates/pico/tests/intern.rs
+++ b/crates/pico/tests/intern.rs
@@ -1,0 +1,56 @@
+use pico::{Database, MemoRef, SourceId};
+use pico_macros::{memo, Source};
+use thiserror::Error;
+
+#[test]
+fn intern() {
+    let mut db = Database::default();
+
+    let id = db.set(Input {
+        key: "key",
+        value: "asdf".to_string(),
+    });
+
+    assert_eq!(**process_input(&db, id).as_ref().unwrap(), 'a');
+}
+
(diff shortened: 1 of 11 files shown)


crates/pico/tests/intern.rs
```diff
@@ -0,0 +1,56 @@
...
+        .map(|v| db.intern(v))
+}
+
+#[memo]
+fn process_input(
+    db: &Database,
+    input_id: SourceId<Input>,
+) -> Result<MemoRef<char>, ProcessInputError> {
+    let result = first_letter(db, input_id).to_owned()?;
+    Ok(result)
```
ch1ffa, line 55: I think this is the right pattern atm. We enforce Ok path to be MemoRef and Err to implement Clone so it looks cleaner than the following:
```rs
first_letter(db, input_id)
  .as_ref()
  .map(|v| *v)
  .map_err(|e| e.clone().into())
```

rbalicki2: nice, yeah, this is clean

rbalicki2 (commented):
partial review... I love the API. All the tests look like how I would want to write pico. I still need to think about the implementation (e.g. think about the double interning problem)

PR #459, opened 2025-03-02
Es/client field directives

dialectician4:
Preliminary work for creating client field directives.


rbalicki2:
Can you fix the endings and re-run generate-fixture-tests and we can land this!

PR #462, opened 2025-03-07
add pico to the compiler [1/N]

ch1ffa:
This is the first iteration of applying `Pico` to the Isograph compiler.

This implementation recreates what we are currently doing in the `main` branch by memoizing only parsed sources. It has slightly worse performance due to pico overhead.

What has been done:
- extracted creating `source_files` step from the `compile` function 
- removed intermediate `SourceFiles` structure (now it is a struct of `SourceId`)
- added 3 memoized functions that parse source texts
- all parse functions inlined into the `create_unvalidated_schema` step awaiting further refactoring
- added simple GC call that runs no sooner than 60 secs after the last change

### DIFF
--- a/crates/isograph_compiler/src/create_unvalidated_schema.rs
+++ b/crates/isograph_compiler/src/create_unvalidated_schema.rs
@@ -1,0 +1,167 @@
+use std::{
+    collections::HashMap,
+    error::Error,
+    ops::{Deref, DerefMut},
+};
+
+use common_lang_types::{CurrentWorkingDirectory, RelativePathToSourceFile, TextSource};
+use isograph_config::CompilerConfig;
+use isograph_lang_parser::IsoLiteralExtractionResult;
+use isograph_lang_types::{IsoLiteralsSource, SchemaSource};
+use isograph_schema::{OutputFormat, UnvalidatedSchema};
+use pico::{Database, SourceId};
+
+use crate::{
+    batch_compile::BatchCompileError,
+    isograph_literals::{parse_iso_literal_in_source, process_iso_literals},
+    refetch_fields::add_refetch_fields_to_objects,
+    source_files::SourceFiles,
+};
+
+pub fn create_unvalidated_schema<TOutputFormat: OutputFormat>(
+    db: &Database,
+    source_files: &SourceFiles,
+    config: &CompilerConfig,
+) -> Result<(UnvalidatedSchema<TOutputFormat>, ContainsIsoStats), Box<dyn Error>> {
+    let mut unvalidated_isograph_schema = UnvalidatedSchema::<TOutputFormat>::new();
+    let schema = TOutputFormat::parse_type_system_document(db, source_files.schema)?.to_owned();
+    let outcome = TOutputFormat::process_type_system_document(
+        &mut unvalidated_isograph_schema,
+        schema,
+        &config.options,
+    )?;
+    let schema_extensions =
+        parse_schema_extensions::<TOutputFormat>(db, &source_files.schema_extensions)?;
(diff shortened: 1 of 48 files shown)


crates/isograph_compiler/src/create_unvalidated_schema.rs
```diff
@@ -0,0 +1,167 @@
...
+fn parse_iso_literals(
+    db: &Database,
+    iso_literals_sources: &HashMap<RelativePathToSourceFile, SourceId<IsoLiteralsSource>>,
+    current_working_directory: CurrentWorkingDirectory,
+) -> Result<ContainsIso, BatchCompileError> {
+    let mut contains_iso = ContainsIso::default();
+    let mut iso_literal_parse_errors = vec![];
+    for (relative_path, iso_literals_source_id) in iso_literals_sources.iter() {
+        match parse_iso_literal_in_source(db, *iso_literals_source_id, current_working_directory)
+            .to_owned()
```
rbalicki2, line 85: nit: we can call `deref` here and `.to_owned` only inside the `Ok` branc

crates/isograph_compiler/src/isograph_literals.rs
```diff
@@ -47,14 +42,14 @@ pub(crate) fn read_files_in_folder(
...
 }
 
 pub fn read_file(
     path: PathBuf,
-    canonicalized_root_path: &Path,
-) -> Result<(PathBuf, String), BatchCompileError> {
+    current_working_directory: CurrentWorkingDirectory,
+) -> Result<(RelativePathToSourceFile, String), BatchCompileError> {
     // This isn't ideal. We can avoid a clone if we changed .map_err to match
     let path_2 = path.clone();
```
rbalicki2, line 54: nit: we can remove this clone, and just clone in the the `map_err` fn. I can't make that change in main because it causes conflicts :joy:

crates/isograph_compiler/src/isograph_literals.rs
```diff
@@ -110,45 +102,51 @@ fn visit_dirs_skipping_isograph(dir: &Path, cb: &mut dyn FnMut(&DirEntry)) -> io
...
+    iso_literals_source_id: SourceId<IsoLiteralsSource>,
+    current_working_directory: CurrentWorkingDirectory,
+) -> Result<
+    Vec<(IsoLiteralExtractionResult, TextSource)>,
+    Vec<WithLocation<IsographLiteralParseError>>,
+> {
+    let IsoLiteralsSource {
+        relative_path,
+        content,
+    } = db.get(iso_literals_source_id).clone();
```
rbalicki2, line 146: nit: no need to clone here, we can just pass content as an &str, and relative impls copy

crates/isograph_compiler/src/watch.rs
```diff
@@ -44,18 +52,50 @@ pub async fn handle_watch_command<TOutputFormat: OutputFormat>(
...
+                                    &state.config,
+                                );
+                                state.source_files = Some(source_files);
+                                result
+                            }
+                        })
                     } else {
                         info!(
                             "{}",
                             "Too many changes. Starting a full compilation.".cyan()
```
rbalicki2, line 87: Perhaps this should go... re-reading everything means that we call `db.set` on everything and throw all intermediate results away, but the whole point of pico is that it can handle whatever we throw at it and gracefully recompute only what's needed. Beforehand, I would've believed that we had some poorly tested logic, but with pico, I think that this is unnecessary caution.

Anyway, could totally be separate PR

rbalicki2 (approved):
shipittt, the nits are there, but we can just fix them afterward. I don't want to wait

PR #472, opened 2025-03-14
add persisted documents

ch1ffa:
This PR introduces **Persisted Documents** support.

A [persisted document](https://spec.graphql.org/draft/#ExecutableDocument) is a GraphQL document that has been persisted such that the server may retrieve it based on an identifier indicated in the HTTP request.

Since server and client implementations may vary, this PR focuses on minimal work to enable configuration through the `IsographEnvironment`. It also unlocks future work targeting SQL

### How it works

The first parameter of `networkFunction`, previously `queryText`, is now replaced with the `IsographOperation` type:
```ts
export type IsographOperation = {
  readonly kind: 'Operation';
  readonly documentId: string | null;
  readonly operationName: string | null;
  readonly operationKind: 'Query' | 'Mutation' | 'Subscription';
  readonly text: string | null;
};
```
Either `text` or `documentId` must be non-null. Users can choose which to use in the `networkFunction` implementation, depending on whether persisted documents are enabled. Here's an example using a `graphql-yoga` server that follows the Apollo APQ Specification:
```ts
fetch('http://localhost:4000/graphql', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
  },
  body: JSON.stringify({
    variables,
    extensions: {
      persistedQuery: {
        version: 1,
        sha256Hash: operation.documentId,
      },
    },
  }),
})
```
According to the draft specification, it can also be sent as a URL-encoded GET request:
```
https://example.com/graphql?documentId=sha256:71f7dc5758652baac68e4a10c50be732b741c892ade2883a99358f52b555286b&variables=%7B%22id%22%3A%22QVBJcy5ndXJ1%22%7D
```
that is why we need to provide `operationKind` as well, as mutations must still be sent via `POST` requests in such cases. 
The specification also describes an optional `operationName` parameter, which is included in the `IsographOperation` argument.

To enable persisted documents, users must add an `options.persisted_documents` section to the `isograph.config.js` file:
```json
"persisted_documents": {
  "directory": null,
  "filename": "persisted_documents.json",
  "algorithm": "sha256"
}
```
where:
- `directory` - the relative path to the directory where the compiler should create the persisted documents file. Defaults to the artifacts directory
- `filename` - the file name used by the compiler to save persisted document information. Defaults to `persisted_documents.json`
- `algorithm` - the hashing algorithm used to compute document hashes. Acceptable values are `sha256` or `md5`


crates/isograph_config/src/compilation_options.rs
```diff
@@ -98,6 +99,20 @@ pub enum JavascriptModule {
     EsModule,
 }
 
+#[derive(Debug, Clone)]
+pub struct PersistedDocumentsOptions {
+    pub directory: Option<AbsolutePathAndRelativePath>,
+    pub filename: ArtifactFileName,
```
rbalicki2, line 105: `artifact_directory` uses a pathbuf, e.g. `pub artifact_directory: Option<PathBuf>,`, can we do the same here?

rbalicki2: I don't think any of these need to be required, e.g. perhaps it should default to `__isograph/persistedDocuments.json` if `filename` is missing?

ch1ffa: There is another struct ConfigFilePersistedDocumentsOptions where directory is an Option and filename is initialized with default value.

rbalicki2: Sorry, to clarify, I'm suggesting we make it a single `Option<PathBuf>` in both `PersistedDocumentsOptions` and `ConfigFilePersistedDocumentsOptions`, and that pathbuf is relative to the `__isograph` directory

crates/common_lang_types/src/string_types.rs
```diff
@@ -9,6 +11,6 @@ macro_rules! derive_display {
     };
 }
 
-#[derive(Debug)]
+#[derive(Debug, Serialize)]
```
rbalicki2, line 14: heh, we should probably not have `QueryText` and instead just have `OperationText`: #473 (no need to fix it here, unless you want to)

crates/isograph_config/src/compilation_options.rs
ch1ffa: Do you mean we should have a single option instead of both directory and filename? I thought we might need one day a different serialization strategy, e.g. hot chocolate checks the specified folder for files in the following format: {Hash}.graphql.

crates/generate_artifacts/src/entrypoint_artifact.rs
```diff
@@ -176,15 +189,36 @@ pub(crate) fn generate_entrypoint_artifacts_with_client_field_traversal_result<
...
-    let mut paths_and_contents = EntrypointArtifactInfo {
-        query_text,
-        query_name,
-        parent_type: parent_object,
-        normalization_ast_text,
-        refetch_query_artifact_import,
-        concrete_type: concrete_type.name,
+    let mut paths_and_contents = vec![];
+    let persist = persisted_documents_options.is_some();
+    if !persist {
```
rbalicki2, line 194: I'm a bit torn here.

On the one hand, I think it's worth generating the query text, even if it isn't used, because it is nice to see that what effect a change to an isograph literal had, and query texts have `format::pretty`, so its easy to see what changed.

On the other hand, that might be unexpected behavior.

Anyway, I think it's best to always generate the query. Perhaps it makes sense to generate a `.sql` or `.graphql` file with the query always, instead of something that exports a string, but anyway, for now, lets always generate the `query_text.ts` file

crates/isograph_config/src/compilation_options.rs
rbalicki2: That's a good idea to support that eventually. Once we do that, it would probably be best to add a discriminator, i.e. have the `persisted_operations` object have the following type (in JS)
```
{ kind: 'OneJson', file, algorith } | { kind: 'FolderContainingHashedFiles', folder, algorith }
```

I'm okay with backwards-incompatible changes here. (And really, in general. I don't think users can expect stability until at least we have all the primitives figured out.)

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -85,4 +132,10 @@ pub enum GenerateArtifactsError {
         \nReason: {message:?}"
     )]
     UnableToDeleteDirectory { path: PathBuf, message: String },
+
+    #[error(
+        "Unable to write to the persistend documents file. \
```
rbalicki2, line 137: nit: typo, `persistend` should be `persisted`

crates/generate_artifacts/src/entrypoint_artifact.rs
```diff
@@ -197,14 +231,18 @@ pub(crate) fn generate_entrypoint_artifacts_with_client_field_traversal_result<
             &reachable_variables,
             index,
         );
-
-        paths_and_contents.push(get_artifact_for_imperatively_loaded_field(
-            schema,
-            artifact_info,
-        ))
+        let (artifacts, imperatively_loaded_persistend_documents) =
```
rbalicki2, line 234: nit: typo, `persistend` should be `persisted`

crates/generate_artifacts/src/operation_text.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use crate::generate_artifacts::OperationText;
+
+pub(crate) fn generate_operation_text(
+    query_name: QueryOperationName,
+    operation_name: IsographObjectTypeName,
+    query_text: QueryText,
+    persisted_documents_options: Option<&PersistedDocumentsOptions>,
+    indentation_level: u8,
+) -> (OperationText, Vec<(DocumentId, QueryText)>) {
```
rbalicki2, line 15: nit: can you use the alias `PersistedDocumentEntry` here?

crates/isograph_config/src/compilation_options.rs
rbalicki2: yeah

crates/generate_artifacts/src/operation_text.rs
```diff
@@ -0,0 +1,54 @@
...
+    let indent = "  ".repeat((indentation_level + 1) as usize);
+    let mut persisted_documents = vec![];
```
rbalicki2, line 17: In the non-persisted case, we create an empty vec and pass that around, and then drop it. That's no big deal.

But it does open us up to bugs where the developer might mess up that logic and generate a persisted queries json file without all of the expected persisted queries present.

So let's restructure this to accept an `Option<(&PersistedDocumentsOptions, &mut Vec<PersistedDocumentEntry>)>`, so that it's harder to make that mistake

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -1,17 +1,21 @@
...
 };
 
-use common_lang_types::ArtifactPathAndContent;
+use common_lang_types::{ArtifactPathAndContent, DocumentId, QueryText};
 use intern::string_key::Lookup;
+use isograph_config::CompilerConfig;
 use thiserror::Error;
 
 pub(crate) fn write_artifacts_to_disk(
     paths_and_contents: impl IntoIterator<Item = ArtifactPathAndContent>,
```
rbalicki2, line 14: Can we keep this file unchanged, and make `path_and_contents` contain the persisted documents JSON if necessary?

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -59,6 +63,49 @@ pub(crate) fn write_artifacts_to_disk(
                 message: e.to_string(),
             })?;
     }
+
+    if let Some(pd_config) = config.options.persisted_documents.as_ref() {
```
rbalicki2, line 67: Also just commenting. If the user changes the config, we might end up in a situation where the old persisted json was written at `/1.json`, and now is written at `/2.json`, and both end up created (i.e. #1 is not deleted). That's fine!

We should eventually keep track of all the generated artifacts that we know about from the previous build and diff that with the new set of generated artifacts. So this problem will solve itself.

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -85,4 +132,10 @@ pub enum GenerateArtifactsError {
         \nReason: {message:?}"
     )]
     UnableToDeleteDirectory { path: PathBuf, message: String },
+
+    #[error(
+        "Unable to write to the persistend documents file. \
+        \nReason: {message:?}"
+    )]
+    UnableToSerializePersistentDocuments { message: String },
```
rbalicki2, line 140: nit: `Persisted`, not `Persistent`, but this error message probably won't exist once the persisted json object is part of the `path_and_contents`

crates/string_key_newtype/src/lib.rs
```diff
@@ -36,6 +36,16 @@ macro_rules! string_key_newtype {
             }
         }
 
+        impl serde::Serialize for $named {
```
rbalicki2, line 39: nit: there's an outdated comment above `// TODO serialize, deserialize`

demos/pet-demo-persist/backend/src/index.js
```diff
@@ -0,0 +1,27 @@
+import { createServer } from 'node:http';
+import { usePersistedOperations } from '@graphql-yoga/plugin-persisted-operations';
+import { createYoga } from 'graphql-yoga';
+import persistedDocumentsDictionary from '../../src/components/__isograph/persisted_documents.json' with { type: 'json' };
```
rbalicki2, line 4: TIL about `with` and imports

rbalicki2 (commented):
Overall looks amazing!!!

If it's trivial to split this into smaller parts, that would be dope, but nbd — once you fix the comments, this should be good to land.

Smaller parts:

- generate query text for imperatively loaded fields
- new demo
- these changes

But again, nbd, I'm happy to land it as a giant commit

ch1ffa:
Yeah! Let's split it

crates/isograph_compiler/src/write_artifacts.rs
ch1ffa: Yes, but only if we agreed on generating json inside artifact_directory. I think we can do it and refactor it later if necessary

PR #475, opened 2025-03-17
generate client pointers `output_type.ts`

PatrykWalach:
This generates `output_type.ts` with `type Pet__pointer__output_type = Link`,
not sure if this should be `type Pet__pointer__output_type = ReturnType<typeof resolver>`

Part of client pointers #53

### DIFF
--- a/crates/generate_artifacts/src/eager_reader_artifact.rs
+++ b/crates/generate_artifacts/src/eager_reader_artifact.rs
@@ -3,9 +3,10 @@ use intern::Lookup;
 
 use isograph_config::{CompilerConfig, GenerateFileExtensionsOption};
 
+use isograph_lang_types::SelectionType;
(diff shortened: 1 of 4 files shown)


rbalicki2:
Hm. Will we ever access this type? e.g. if we have

```
field Query.Foo { 
pet_pointer {
  name
}
}
```

then the `Query.Foo` output type will be

```
readonly pet_pointer: {
  name: string
}
```

i.e. we don't ever need to reference `Pet__pointer__output_type`

Wdyt?

PatrykWalach:
It's not accessed outside `reader_resolver.ts`.
I thought it's more consistent to generate this than to inline the type

rbalicki2 (approved):
LGTM, thanks for explaining why it makes sense to have the generated artifact

PR #476, opened 2025-03-18
persisted documents [2/3]: add persisted documents

ch1ffa:
This PR introduces **Persisted Documents** support.

A [persisted document](https://spec.graphql.org/draft/#ExecutableDocument) is a GraphQL document that has been persisted such that the server may retrieve it based on an identifier indicated in the HTTP request.

Since server and client implementations may vary, this PR focuses on minimal work to enable configuration through the `IsographEnvironment`. It also unlocks future work targeting SQL

### How it works

The first parameter of `networkFunction`, previously `queryText`, is now replaced with the `IsographOperation` or `IsographPersistedOperation` types:
```ts
export type IsographOperation = {
  readonly kind: 'Operation';
  readonly text: string;
};

export type IsographPersistedOperation = {
  readonly kind: 'PersistedOperation';
  readonly operationId: string;
  readonly extraInfo?: IsographPersistedOperationExtraInfo | null;
};

export type IsographPersistedOperationExtraInfo = {
  readonly kind: 'PersistedOperationExtraInfo';
  readonly operationName: string | null;
  readonly operationKind: 'Query' | 'Mutation' | 'Subscription';
};
```

It's up to users to provide a correct operation type for `networkFunction` implementation and how to use it to fetch data. Here's an example using a `graphql-yoga` server that follows the Apollo APQ Specification:
```ts
fetch('http://localhost:4000/graphql', {
  method: 'POST',
  headers: {
    'Content-Type': 'application/json',
  },
  body: JSON.stringify({
    variables,
    extensions: {
      persistedQuery: {
        version: 1,
        sha256Hash: operation.operationId,
      },
    },
  }),
})
```

According to the draft specification, it can also be sent as a URL-encoded GET request:
```
https://example.com/graphql?documentId=sha256:71f7dc5758652baac68e4a10c50be732b741c892ade2883a99358f52b555286b&variables=%7B%22id%22%3A%22QVBJcy5ndXJ1%22%7D
```
that is why we may need to provide `operationKind` as mutations must be sent via `POST` requests in such cases. The specification also describes an optional `operationName` parameter.  These optional parameters can be included as `extraInfo` using configuration flag `includeExtraInfo`. 

To enable persisted documents, users must add an `options.persisted_documents` section to the `isograph.config.js` file:
```json
"persisted_documents": {
  "file": "persisted_documents.json",
  "algorithm": "sha256",
  "includeExtraInfo": false
}
```
where:
- `file` - the file name used by the compiler to save persisted document information. Defaults to `persisted_documents.json`
- `algorithm` - the hashing algorithm used to compute document hashes. Acceptable values are `sha256` or `md5`
- `includeExtraInfo` - include extra info to the operation text. Defaults to `false`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3367,7 +3367,7 @@ version = "0.19.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b5bb770da30e5cbfde35a2d7b9b8a2c4b8ef89548a7a6aeab5c9a576e3e7421"
 dependencies = [
- "indexmap 2.7.1",
+ "indexmap 2.10.0",
  "toml_datetime",
  "winnow",
 ]
--- a/crates/isograph_schema/src/network_protocol.rs
+++ b/crates/isograph_schema/src/network_protocol.rs
@@ -1,10 +1,8 @@
 use std::{error::Error, fmt::Debug, hash::Hash};
 
 use common_lang_types::{
-    DescriptionValue, Location, QueryExtraInfo, QueryOperationName, QueryText,
-    ServerObjectEntityName, ServerSelectableName, UnvalidatedTypeName, WithLocation, WithSpan,
-    Location, QueryOperationName, QueryText, ServerObjectEntityName, ServerSelectableName,
-    UnvalidatedTypeName, WithLocation, WithSpan,
+    Location, QueryExtraInfo, QueryOperationName, QueryText, ServerObjectEntityName,
+    ServerSelectableName, UnvalidatedTypeName, WithLocation, WithSpan,
 };
 use graphql_lang_types::{GraphQLInputValueDefinition, GraphQLTypeAnnotation, RootOperationKind};
 use isograph_lang_types::Description;


ch1ffa:
@rbalicki2 I've rebased to the latest changes. Take a look please, when you have time.

rbalicki2:
sounds good will look

rbalicki2:
Yes yes yes, love it. Just a few nits about what we generate, but overall looks solid.

- love that queryText -> operationText
- This allows for the case where both the text and id are missing. Maybe we should instead have

```
{ kind: 'OperationText', operationText: string, documentId: string | null }
| { kind: 'OperationId', operationId: string }
```

- is documentId or operationId correct? [Benjie](https://benjie.dev/graphql/trusted-documents) refers to it as `documentId`, [apollo](https://www.apollographql.com/docs/kotlin/advanced/persisted-queries) uses `operationId`. I tend to lean toward `operationId`, because it's consistent with the rest of the struct.
- what do we need `operationName` and `operationType` for? These are both GraphQL-isms, especially `operationType`. There is always a root entrypoint we can use as an operation name. If these are needed (or useful in dev, e.g. for debugging), perhaps we can put them all under a separate key, like `{ kind: 'OperationText', extraInfo: { operationName, operationType } }`, so that:
    - these can be null in prod, or otherwise their existence can be determined via config or directive on an entrypoint
    - these more easily vary by network protocol
- I'm much more inclined to keep `operationName` than `operationType`, though, since that's derivable from the name.
- lets turn this on for at least one demo, and set it to "both". So I guess the config might be `null | { file, algorithm, alsoIncludeOperationText? }`?

rbalicki2:
One last thing: can we sort the entries in the persisted_documents object alphabetically by query name, so that when you make a small change to a client field, the differences are easily grokked?

rbalicki2:
And as a side note, I love that we don't accumulate persisted documents forever, which Relay does, and is a source of a lot of frustration. I have considered fixing that upstream, but it's probably easier for us to just avoid checking generated files in 😂

ch1ffa:
> - what do we need `operationName` and `operationType` for?

Some implementations might use GET for queries (to cache) and POST for mutations, so one can need this info inside network function.

rbalicki2:
Hm. For queries/GET, shouldn't we only be going to network if we either explicitly need the data (i.e. something is missing) or we're saying `shouldFetch: "Yes"`? It doesn't seem like the network function is a great place to do caching. Maybe I'm missing something?

In particular, imagine this scenario:
- I fetch Q1 `currentUser { bestFriend { name } }`
- I fetch Q2 `currentUser { bestFriend { age } }` and discover that the `bestFriend` has changed.
- So, I refetch Q1 (perhaps Isograph automates this). If we reuse the cached values, that's wrong! At that point, we might automatically refetch Q2. If we use cached values both times, we'll just spin the CPU lol

ch1ffa:
I mean cache on the network level, not our internal cache. It's not about isograph, we indeed don't need it, we can decide when to fetch exactly. But I'm talking about compatibility with some existing backend that uses GET for queries and POST for mutations. Anyway, we can remove operation type for now, tbh I've just tried to stick to the graphql documentation which is not always necessary for us.

rbalicki2:
Ah, I thought you were supposed to POST even queries, heh

ch1ffa:
Well, according to the draft specification you can use GET `https://example.com/graphql?documentId=sha256:71f7dc5758652baac68e4a10c50be732b741c892ade2883a99358f52b555286b&variables=%7B%22id%22%3A%22QVBJcy5ndXJ1%22%7D`

rbalicki2:
TIL!

crates/generate_artifacts/src/generate_artifacts.rs
```diff
@@ -397,6 +411,12 @@ fn get_artifact_path_and_content_impl<TNetworkProtocol: NetworkProtocol>(
         config.options.no_babel_transform,
     ));
 
+    if let Some(pd) = persisted_documents {
```
rbalicki2: nit: I think we should generate an empty file, if there are no persisted docs

crates/generate_artifacts/src/operation_text.rs
```diff
@@ -0,0 +1,87 @@
...
+                    {indent}  kind: \"PersistedOperation\",\n\
+                    {indent}  operationId: \"{operation_id}\",\n\
+                    {indent}  extraInfo: {query_extra_info},\n\
+                    {indent}}}"
+                ))
+            } else {
+                OperationText(format!(
+                    "{{\n\
+                    {indent}  kind: \"PersistedOperation\",\n\
+                    {indent}  operationId: \"{operation_id}\",\n\
```
rbalicki2: nit: let's make `extraInfo: null` here

crates/graphql_network_protocol/src/query_text.rs
```diff
@@ -71,50 +75,53 @@ fn write_selections_for_query_text<'a>(
     query_text: &mut String,
     items: impl Iterator<Item = &'a MergedServerSelection> + 'a,
     indentation_level: u8,
+    format: Format,
 ) {
+    let (new_line, indent) = match format {
+        Format::Pretty => ("\\\n", &"  ".repeat(indentation_level as usize).to_string()),
+        Format::Compact => (" ", &"".to_string()),
```
rbalicki2, line 88: nice

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -125,15 +128,32 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+    fn generate_query_extra_info(
+        query_name: QueryOperationName,
+        operation_name: ServerObjectEntityName,
+        indentation_level: u8,
+    ) -> QueryExtraInfo {
+        let indent = "  ".repeat((indentation_level + 1) as usize);
+        QueryExtraInfo(format!(
+            "{{\n\
+            {indent}  kind: \"PersistedOperationExtraInfo\",\n\
+            {indent}  operationName: \"{query_name}\",\n\
```
rbalicki2, line 151: Nit: So I think it would be nice to include or to have an option to include the query text here, i.e. to allow you to incrementally adopt persisted queries. But we can just leave that as a TODO, it doesn't seem urgent

crates/isograph_config/src/compilation_options.rs
```diff
@@ -39,6 +39,7 @@ pub struct CompilerConfigOptions {
     pub include_file_extensions_in_import_statements: GenerateFileExtensionsOption,
     pub module: JavascriptModule,
     pub generated_file_header: Option<GeneratedFileHeader>,
+    pub persisted_documents: Option<PersistedDocumentsOptions>,
```
rbalicki2, line 42: love that `persisted_documents: {}` just works

libs/isograph-react/src/core/cache.ts
```diff
@@ -93,9 +93,17 @@ export function getOrCreateCacheForArtifact<
   variables: ExtractParameters<TReadFromStore>,
   fetchOptions?: FetchOptions<TClientFieldValue>,
 ): ParentCache<FragmentReference<TReadFromStore, TClientFieldValue>> {
-  const cacheKey =
-    entrypoint.networkRequestInfo.queryText +
-    JSON.stringify(stableCopy(variables));
+  let cacheKey = '';
+  switch (entrypoint.networkRequestInfo.operation.kind) {
+    case 'Operation':
+      cacheKey = entrypoint.networkRequestInfo.operation.text;
```
rbalicki2: I think this should include `JSON.stringify(stableCopy(variables))` as well

libs/isograph-react/src/core/entrypoint.ts
```diff
@@ -19,9 +19,27 @@ export type ReaderWithRefetchQueries<
...
+
+export type IsographOperation = {
+  readonly kind: 'Operation';
+  readonly text: string;
+};
+
+export type IsographPersistedOperation = {
+  readonly kind: 'PersistedOperation';
+  readonly operationId: string;
+  readonly extraInfo?: IsographPersistedOperationExtraInfo | null;
```
rbalicki2: let's make it so that this field cannot be missing, i.e. instead make it null

rbalicki2 (changes requested):
Looks amazing!! Just a few minor tweaks, but overall, love it. I fixed the clippy stuff on master, and it rebases cleanly

rbalicki2:
> ... Apollo Client automatically sends the short hashed queries as GET requests enabling a CDN to serve those requests. For full-length queries and for all mutations, Apollo Client will continue to use POST requests.

Looks like sending everything via POST is valid! Anyway, not a blocker

rbalicki2:
just FYI fixed up some merge conflicts, incorrectly (since its hard to merge imports manually lol), just to unblock merging into master. Cool! Looks like I can push to this branch, so I just pushed to this branch to clean up my shit merge job 😂  okay now I know I can do it from the CLI

PR #480, opened 2025-03-29
Get rid of the validation step by processing selection sets after fields are created

rbalicki2:
# Why

## Removing single points of failure

- This is part of work that will make schema creation more incremental by removing steps that are "single points of failure". Namely, by removing the validation step.
- If the validation step depends on each field/type/selection set/etc, then a change to any selection set invalidates everything validation onward.
- But if instead, when we validate (say) `pet { name }`, we only need to ensure that the `pet` field exists, get its type (Pet), ensure that the Pet type has a field `name`, etc. This doesn't need to involve validating (say) `NewsfeedItem`

## Why did we have the validation step

- The validation step exists because originally, we created objects and their fields. But a field might refer to an object that hasn't been created. So, we needed to go back and validate. Likewise, we created client fields, along with their selection sets, but those selection sets might reference fields that hadn't been created.
- We previously restructured things so that fields are created after objects.
- Now, we do the same for selection sets: process them after objects.

## Bigger picture, next steps

- What we do now is to modify objects and add their fields as a follow up, and modify client fields to add their selection sets
- Instead, these should be separate objects
- At this point, we probably don't need a "schema", which is a large global object and a "single point of failure"
- get rid of entrypoint validation, then we don't need a function call validate_and_construct


rbalicki2 (commented):
LGTM shipit 10/10 nonotes

PR #484, opened 2025-04-03
Split selectables into two structs: Object selectable and scalar selectable

rbalicki2:
# What

- There previously was a struct `ServerScalarSelectable` which was incorrectly named, and which was actually `ServerSelectable`, i.e. it could represent both a scalar selectable (i.e. scalar field) and an object selectable (linked field)
- Now these are two separate fields, importantly, with `ServerScalarSelectable` having a field `target_scalar_entity: TypeAnnotation<ServerScalarId>,` and `ServerObjectSelectable` having ` target_object_entity: TypeAnnotation<ServerObjectId>,`.
- This allows us to avoid redundant checks and avoid panics in our codebase! (This PR gets rid of 4 expects and 5 panics, but this isn't an exact count.)
- Namely, once we are in a fully validated state, we can go from scalar selection -> scalar selectable -> scalar entity without any panics (in "user" code, at least). Namely, there we don't have to do `.as_scalar().expect(...)` or the like. NICE

# Big picture: what's next

- Relax? Drink a beer? Celebrate? NO.
- The next big goal is to do two things
    - Dissemble the schema (and store the parts in pico)
    - Make "validation" pull-based, (e.g. if we are validating a single selection, we just need to validate its selectable, and thus its scalar entity, not anything else). It's not really validation, though, it's more like "dissembled schema creation". 

## What does that mean

- If we open up VSCode and the isograph language server is running, and the initial request is to provide hover information for "name" in:
```
field User.UserName {
  pet {
    name
  }
  other_fields
}
```
- We can do the following:
    - (if possible) split the GraphQL schema into chunks, creating a map of each entity type/field name (e.g. `User`, `User.pet`, `Pet`, `Pet.name`, `String`) to a vec of unparsed chunks (with only one item)
    - do the bare minimum of parsing of iso literals, i.e. just parse their declarations, and thus add `User.UserName` -> unprocessed chunk to that map
    - validate that `User` exists by looking up `User` in the chunks DB. Validate that it is an object
    - validate that `User.UserName` is unique
    - break apart the selection set into chunks and identify the chunk under the cursor
    - validate that `User.name` exists and is an object, look up where it points (`Pet`)
    - validate that `Pet` exists and is an object
    - break apart the nested selection set `{ name }` into chunks (i.e. a single chunk) and pick the chunk under the cursor (the only chunk)
    - validate that `Pet.name` exists and is a scalar, look up where it points (`String`)
    - validate that `String` exists and is a scalar
    - Show information about the `String` scalar, `name` field, `Pet` entity.

- What we didn't do
    - validate the selection `other_selection` is valid
    - validate any other type or selectable

- We can consider:
    - validating that `UserName` is unique asynchronously, but presumably we need to at least create a map of `selectable -> chunk` in advance, since otherwise how would we know where to find items in the selection set we are parsing
    
Step 3: profit

### DIFF
--- a/crates/generate_artifacts/src/eager_reader_artifact.rs
+++ b/crates/generate_artifacts/src/eager_reader_artifact.rs
@@ -5,11 +5,11 @@ use isograph_config::{CompilerConfig, GenerateFileExtensionsOption};
 
 use isograph_schema::{
     initial_variable_context, ClientField, ClientFieldOrPointer, NetworkProtocol, Schema,
-    ValidatedSelectionType,
+    ServerObjectSelectable, ValidatedSelectionType,
 };
 use isograph_schema::{
-    RefetchedPathsMap, ServerFieldTypeAssociatedDataInlineFragment, ServerScalarSelectable,
-    UserWrittenClientTypeInfo, UserWrittenComponentVariant,
+    RefetchedPathsMap, ServerFieldTypeAssociatedDataInlineFragment, UserWrittenClientTypeInfo,
+    UserWrittenComponentVariant,
 };
 
 use std::{borrow::Cow, collections::BTreeSet, path::PathBuf};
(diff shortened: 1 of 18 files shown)


rbalicki2 (commented):
LGTM

PR #489, opened 2025-04-06
Combine two NetworkProtocol trait methods into one

rbalicki2:
- Combine NetworkProtocol's parse_type_system_document and parse_type_system_extension_document into parse_type_system_documents
- This is an initial step; the goal is to also combine process_type_system_document and process_type_system_extension_document into a single method that returns an object and doesn't mutate schema

@ch1ffa can you review? In particular, am I doing the right thing with

```
    Result<
        MemoRef<(
            Self::TypeSystemDocument,
            BTreeMap<RelativePathToSourceFile, Self::TypeSystemExtensionDocument>,
        )>,
        E
    >
```

My heuristic is that the correct (and performant) thing to do is to wrap `T` in `Result<T, E>` in `MemoRef`, so that we don't clone the underlying `T` excessively.

We still need to clone the underlying object in `create_unvalidated_schema`, which might not be necessary when we unify all of the methods, or I can try to refactor out in the future.

### DIFF
--- a/crates/graphql_network_protocol/src/graphql_network_protocol.rs
+++ b/crates/graphql_network_protocol/src/graphql_network_protocol.rs
@@ -1,6 +1,6 @@
-use std::error::Error;
+use std::{collections::BTreeMap, error::Error};
 
-use common_lang_types::{QueryOperationName, QueryText};
+use common_lang_types::{QueryOperationName, QueryText, RelativePathToSourceFile};
 use graphql_lang_types::{GraphQLTypeSystemDocument, GraphQLTypeSystemExtensionDocument};
 use isograph_lang_types::SchemaSource;
 use isograph_schema::{
@@ -10,7 +10,7 @@ use isograph_schema::{
 use pico::{Database, MemoRef, SourceId};
 
 use crate::{
-    parse_graphql_schema, parse_schema_extensions_file,
+    parse_graphql_schema,
     process_type_system_definition::{
         process_graphql_type_extension_document, process_graphql_type_system_document,
     },
@@ -27,18 +27,18 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
 
     type SchemaObjectAssociatedData = GraphQLSchemaObjectAssociatedData;
 
-    fn parse_type_system_document(
+    fn parse_type_system_documents(
         db: &Database,
         schema_source_id: SourceId<SchemaSource>,
-    ) -> Result<MemoRef<Self::TypeSystemDocument>, Box<dyn Error>> {
-        Ok(parse_graphql_schema(db, schema_source_id).to_owned()?)
-    }
-
-    fn parse_type_system_extension_document(
-        db: &Database,
-        schema_extension_source_id: SourceId<SchemaSource>,
-    ) -> Result<MemoRef<Self::TypeSystemExtensionDocument>, Box<dyn Error>> {
-        Ok(parse_schema_extensions_file(db, schema_extension_source_id).to_owned()?)
+        schema_extension_sources: &BTreeMap<RelativePathToSourceFile, SourceId<SchemaSource>>,
+    ) -> Result<
+        (
+            MemoRef<Self::TypeSystemDocument>,
+            BTreeMap<RelativePathToSourceFile, MemoRef<Self::TypeSystemExtensionDocument>>,
+        ),
+        Box<dyn Error>,
+    > {
+        Ok(parse_graphql_schema(db, schema_source_id, schema_extension_sources).to_owned()?)
     }
 
     fn process_type_system_document(
--- a/crates/graphql_network_protocol/src/read_schema.rs
+++ b/crates/graphql_network_protocol/src/read_schema.rs
@@ -1,6 +1,6 @@
-use std::{path::PathBuf, str::Utf8Error};
+use std::{collections::BTreeMap, path::PathBuf, str::Utf8Error};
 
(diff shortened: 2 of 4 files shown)


crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -27,18 +27,20 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+    ) -> Result<
+        MemoRef<(
+            Self::TypeSystemDocument,
+            BTreeMap<RelativePathToSourceFile, Self::TypeSystemExtensionDocument>,
+        )>,
+        Box<dyn Error>,
+    > {
+        parse_graphql_schema(db, schema_source_id, schema_extension_sources)
+            .to_owned()
+            .map_err(|e| Box::new(e) as Box<dyn std::error::Error>)
```
ch1ffa, line 43: just curious, why do you prefer `map_err` over `Ok(parse()?)`

crates/graphql_network_protocol/src/read_schema.rs
```diff
@@ -12,16 +12,31 @@ use thiserror::Error;
...
+
     let schema = parse_schema(content, *text_source)
-        .map(|document| db.intern(document))
         .map_err(|with_span| with_span.to_with_location(*text_source))?;
-    Ok(schema)
+
+    let mut schema_extensions = BTreeMap::new();
+    for (relative_path, schema_extension_source_id) in schema_extension_sources.iter() {
+        let extensions_document =
+            parse_schema_extensions_file(db, *schema_extension_source_id).to_owned()?;
```
ch1ffa, line 35: `to_owned` here is not cheap. In the previous implementation this was a boundary between memoized part of the compilation and the rest. Now that we have a parent `memo` function it make sense to return `BTreeMap<RelativePathToSourceFile, MemoRef<GraphQLTypeSystemExtensionDocument>>`, but it can be done later, when we memoize `process_type_system_extension_document` to pass `MemoRef<GraphQLTypeSystemExtensionDocument>` there

ch1ffa (approved):
Everything is correct on this stage of incrementalisation.

I remember your note, that you wanted to remove "graphql-ism" of having schema and extensions separately. But when I was thinking about sql target, it seems like it also could have some kind of extensions, e.g. we can have separate `sql` files that describe mutations (i.e. sql insert/update statements), that can be used to generate corresponding server side code (this idea is heavily used in [sqlc](https://sqlc.dev)). So yeah, I like that we still keep `TypeSystemExtensionDocument`

ch1ffa:
> My heuristic is that the correct (and performant) thing to do is to wrap T in Result<T, E> in MemoRef, so that we don't clone the underlying T excessively.

Yeah, intuition is that we wrap `T` in `MemoRef` if we are going to use it later as an argument of another memoized function.

crates/graphql_network_protocol/src/graphql_network_protocol.rs
rbalicki2: Oh, nice, didn't think of that! `Ok(parse()?)` is definitely cleaner

crates/graphql_network_protocol/src/read_schema.rs
rbalicki2: Okay, so I think option 3 is the best way forward, and it's what's in the latest commit

### Option 1: return `Result<MemoRef<(..., BTreeMap<..., MemoRef<...>>)>, E>`
- We can't call `db.intern(something_that_contains_a_memo_ref)`, so we can't return `MemoRef<(..., BTreeMap<..., MemoRef<...>>)>`, because `MemoRef` currently doesn't implement `Hash`, which `db.intern` requires. I don't see a reason `MemoRef` can't.
- It sounds like a good heuristic to only intern the "inner, expensive" part (though a BTreeMap might be expensive as well, so maybe it is desirable to double-intern.)
- Anyway, wdyt about double interning?
- Note that a `#[memo]` function can return `MemoRef<Result<MemoRef<...>, E>>`, so there's some workaround here

### Option 2: return `Result<(MemoRef<...>, BTreeMap<Key, MemoRef<...>>, E>`
- If `parse_schema_extension_file` returns `MemoRef<Result<GraphQLTypeSystemExtensionDocument, BatchCompileError>>` (as it does on main), the only way I can make a `MemoRef<GraphQLTypeSystemExtensionDocument>` to insert is
```
        let extensions_document =
            db.intern(parse_schema_extensions_file(db, *schema_extension_source_id).to_owned()?);
```
but that seems wrong
- basically, the issue is that `MemoRef<Result<T, E>>` cannot be unwrapped with `?`, but we can have `parse_schema_extension_file` return `MemoRef<Result<MemoRef<...>, E>>`

### Option 3: converrt `parse_schema_extension_file` to return `MemoRef<Result<MemoRef<GraphQLTypeSystemExtensionDocument>, BatchCompileError>>`
- This works. This is what's in this commit.
- This requires us to call `type_system_document.to_owned(),` and `extension_document.to_owned()` in `create_unvalidated_graphql_schema`, which is not ideal (?)
- As long as the functions `process_type_system_document` etc take ownership of `type_system_document`, it seems like cloning is inevitable. We may be able to clone smaller chunks of the struct, e.g. just the directives (👎) and fields, instead of the whole struct
- Perhaps the cloning can happen at level 1 only if the value actually changes, and at level 2, we return a reference, and level 3 (all the language server code) we only use references. I'm not sure.

Wdyt?

PR #497, opened 2025-04-12
generate client pointers reader ast

PatrykWalach:
Part of #53

### DIFF
--- a/crates/generate_artifacts/src/reader_ast.rs
+++ b/crates/generate_artifacts/src/reader_ast.rs
@@ -54,44 +54,70 @@ fn generate_reader_ast_node<TNetworkProtocol: NetworkProtocol>(
             }
         }
         SelectionTypeContainingSelections::Object(linked_field_selection) => {
-            path.push(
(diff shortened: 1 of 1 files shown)


crates/generate_artifacts/src/reader_ast.rs
```diff
@@ -652,33 +686,39 @@ fn refetched_paths_with_path<TNetworkProtocol: NetworkProtocol>(
...
-                                .arguments
-                                .iter()
-                                .map(|x| x.item.into_key_and_value()),
-                            // TODO this clearly does something, but why are we able to pass
-                            // the initial variable context here??
-                            initial_variable_context,
-                        ),
+                match linked_field_selection.associated_data.field_id {
+                    DefinitionLocation::Client(_) => {
+                        // Do not recurse into selections of client pointers
```
rbalicki2, line 691: should we recurse into the refetch selection set?

I'll merge regardless, doesn't seem like a blocker (since the refetch selection sets for now only have ids)

PR #506, opened 2025-04-27
[Client Pointers] generate entrypoints

PatrykWalach:
Some tests used:
- nested refetch:
```graphql
self {
  age
  __refetch
  self {
    tagline
  }
}
```        
```graphql
pointer Pet.self to Pet {
  debug: alt_tagline
  link
}
```

- nested args:
```graphql
  field Pet.PetStatsCard($petId: ID!) @component {
    id
    nickname
    age
    self {
      stats {
        weight
        intelligence
        cuteness
        hunger
        sociability
        energy
        refetch_pet_stats(id: $petId)
      }
    }
  }
```

Part of #53

### DIFF
--- a/crates/generate_artifacts/src/entrypoint_artifact.rs
+++ b/crates/generate_artifacts/src/entrypoint_artifact.rs
@@ -1,5 +1,3 @@
-use std::collections::BTreeSet;
-
 use common_lang_types::{
     ArtifactPathAndContent, ClientScalarSelectableName, ParentObjectEntityNameAndSelectableName,
     QueryOperationName, QueryText, ServerObjectEntityName, VariableName,
@@ -9,21 +7,26 @@ use isograph_lang_types::{
     DefinitionLocation, EmptyDirectiveSet, EntrypointDirectiveSet, ScalarSelectionDirectiveSet,
     SelectionType,
 };
+
+use isograph_schema::ClientScalarSelectable;
 use isograph_schema::{
     create_merged_selection_map_for_field_and_insert_into_global_map,
     current_target_merged_selections, get_imperatively_loaded_artifact_info,
     get_reachable_variables, initial_variable_context, ClientScalarOrObjectSelectable,
-    ClientScalarSelectable, EntrypointDeclarationInfo, FieldToCompletedMergeTraversalStateMap,
-    FieldTraversalResult, Format, MergedSelectionMap, NetworkProtocol, RootOperationName,
-    RootRefetchedPath, ScalarClientFieldTraversalState, Schema, ServerObjectEntity,
-    ValidatedVariableDefinition, WrappedSelectionMapSelection,
+    EntrypointDeclarationInfo, FieldToCompletedMergeTraversalStateMap, FieldTraversalResult,
+    Format, MergedSelectionMap, NetworkProtocol, RootOperationName, RootRefetchedPath,
+    ScalarClientFieldTraversalState, Schema, ServerObjectEntity, ValidatedVariableDefinition,
+    WrappedSelectionMapSelection,
 };
+use std::collections::BTreeSet;
 
+use crate::generate_artifacts::ENTRYPOINT_FILE_NAME;
+use crate::generate_artifacts::NORMALIZATION_AST_FILE_NAME;
+use crate::generate_artifacts::QUERY_TEXT_FILE_NAME;
 use crate::{
     generate_artifacts::{
-        NormalizationAstText, RefetchQueryArtifactImport, ENTRYPOINT_FILE_NAME, NORMALIZATION_AST,
-        NORMALIZATION_AST_FILE_NAME, QUERY_TEXT, QUERY_TEXT_FILE_NAME, RESOLVER_OUTPUT_TYPE,
-        RESOLVER_PARAM_TYPE, RESOLVER_READER,
+        NormalizationAstText, RefetchQueryArtifactImport, NORMALIZATION_AST, QUERY_TEXT,
+        RESOLVER_OUTPUT_TYPE, RESOLVER_PARAM_TYPE, RESOLVER_READER,
     },
     imperatively_loaded_fields::get_artifact_for_imperatively_loaded_field,
     normalization_ast_text::generate_normalization_ast_text,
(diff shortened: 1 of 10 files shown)


crates/isograph_schema/src/isograph_schema.rs
```diff
@@ -624,6 +624,13 @@ impl<TNetworkProtocol: NetworkProtocol> ServerEntityData<TNetworkProtocol> {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct PathToClientPointer {
+    pub linked_fields: Vec<NormalizationKey>,
+    pub field_name: ClientObjectSelectableName,
+    pub parent_selectable_key: ClientSelectableKey,
```
rbalicki2, line 631: What is this struct going to be used for? I'm not sure I follow, in particular, what the `parent_selectable_key` is going to be used for

rbalicki2 (commented):
Mind if we hop on a call so that I can understand what's going on? TBH I'm a bit lost, but it could be cause I'm really tired 😅

crates/isograph_schema/src/isograph_schema.rs
PatrykWalach: Sure, we can hop on call

PatrykWalach:
1. Right now this seems to be 90% working, `refetchQuery` for unnested pointer requests too many fields 😭 (but I have a feeling it'd run at runtime)
2. Refetch queries don't have `readerAst` so we use `selections` for that, this means, we don't need `usedRefetchQueries` 😄

PatrykWalach:
This is fully working! 😄

rbalicki2:
Omg I didn't notice your comment!!! Yes!!!! I should be able to finally get through the backlog tomorrow or next week

Once this lands and we add some uses of it in our demos, let's cut a release 😎

crates/isograph_schema/src/create_merged_selection_set.rs
```diff
@@ -64,6 +66,7 @@ pub struct RootRefetchedPath {
 pub enum MergedServerSelection {
     ScalarField(MergedScalarFieldSelection),
     LinkedField(MergedLinkedFieldSelection),
+    ClientPointer(MergedLinkedFieldSelection),
```
rbalicki2, line 69: did we discuss this? I'm a bit worried that this is the wrong move here, but TBH I feel like I would have noticed this and commented on it in the past.

Okay, so what's happening here (lol I'm literally prompting myself like I'm an LLM):

- scalar server fields, object server fields and client pointers exist in the same namespace, so there's no chance of a conflict here, that's fine
- this flows into the reachable variables. These flow into `get_used_variable_definitions`, so I think that's fine. Because if a variable is used in a client pointer, but not elsewhere, it should show up as a required parameter of the client pointer's loadable field type. (Does it??)

Okay, so as far as I can tell, it's not obviously broken. But does it belong? There are a few places we ignore this variant. Maybe that's indicative of it not belonging.

Okay, let's chat about this sometime.

PatrykWalach: I did it this way so it's easy to traverse in `current_target_merged_selections` when client pointer is in another pointer

PatrykWalach:
I thought we have a bug, but it's expected behavior:
- calling `set_best_friend` in `PetUpdater` will not refetch `best_friend_relationship`

because  `best_friend_relationship` is not included in the response after using `pointer` 

In the future, when client pointers support inlining selection set, the `best_friend_relationship` should be refetched, because isograph will understand the relationship between the selection set 

```tsx

export const self = iso(`
  pointer Pet.self to Pet {
    link
  }
`)(function self({ data }) {
  return data.link;
});

export const PetBestFriendCard = iso(`
  field Pet.PetBestFriendCard @component {
    id
    PetUpdater
    self {
      best_friend_relationship {
        picture_together
        best_friend {
          id
          name
          picture
        }
      }
    }
  }
`)(function PetBestFriendCardComponent({ data }) {

```

crates/generate_artifacts/src/entrypoint_artifact.rs
```diff
@@ -147,28 +145,36 @@ pub(crate) fn generate_entrypoint_artifacts_with_client_field_traversal_result<
         .refetch_paths
         .iter()
         .map(|((path, selection_variant), root_refetch_path)| {
-            let current_target_merged_selections = match selection_variant {
-                ScalarSelectionDirectiveSet::Updatable(_)
-                | ScalarSelectionDirectiveSet::None(_) => {
-                    current_target_merged_selections(&path.linked_fields, merged_selection_map)
+            let current_target_merged_selections = match &path.field_name {
+                SelectionType::Object(name_and_arguments) => {
+                    let mut linked_fields = path.linked_fields.clone();
```
rbalicki2, line 157: what's happening here? We receive an immutable `ClientTraversalState`, which IMO implies that we should have been done creating the merge traversal state entirely by now. Why do we need to continue mutating it here?

PatrykWalach: I think this `current_target_merged_selections` is actually the main reason,  ClientPointer is a NormalizationKey.

With addition of client pointers, path to imperative fields contains linked fields and client pointers and has to traverse through them

PatrykWalach: Ok, I think the `MergedServerSelection::ClientPointer` definitely doesn't belong and can be removed easily, but `NormalizationKey::ClientPointer` should stay

rbalicki2:
Okay, nice, just lmk when to re-review then

crates/generate_artifacts/src/entrypoint_artifact.rs
PatrykWalach: I don't see a different way to do `PathToRefetchField` without this

crates/isograph_schema/src/create_merged_selection_set.rs
```diff
@@ -72,7 +74,8 @@ impl MergedServerSelection {
     pub fn reachable_variables(&self) -> Vec<VariableName> {
         match self {
             MergedServerSelection::ScalarField(field) => get_variables(&field.arguments).collect(),
-            MergedServerSelection::LinkedField(field) => get_variables(&field.arguments)
+            MergedServerSelection::ClientPointer(field)
```
PatrykWalach, line 77: @rbalicki2 I've tried removing `MergedSelection::ClientPointer` but I couldn't  compute reachable variables here

PR #509, opened 2025-05-04
Checking for lazy & eager loading clash on entrypoints

dialectician4:


crates/isograph_schema/src/validate_entrypoint.rs
```diff
@@ -1,35 +1,69 @@
...
+impl LoadTypedIsoLiteralText {
+    pub fn new(entrypoint_declaration: WithSpan<EntrypointDeclaration>) -> Self {
+        let lazy_status = match entrypoint_declaration.item.entrypoint_directive_set {
+            EntrypointDirectiveSet::LazyLoad(_) => true,
+            _ => false};
+        LoadTypedIsoLiteralText { iso_literal_text: entrypoint_declaration.item.iso_literal_text, is_lazy_loaded: lazy_status }
+    }
+}
+
+impl fmt::Display for LoadTypedIsoLiteralText {
```
rbalicki2, line 36: probably not needed

crates/isograph_schema/src/validate_entrypoint.rs
```diff
@@ -1,34 +1,76 @@
...
         match validate_entrypoint_type_and_field(schema, text_source, entrypoint_declaration) {
             Ok(client_field_id) => {
-                entrypoints.insert(
-                    client_field_id,
-                    entrypoint_declaration.item.iso_literal_text,
-                );
+                let new_entrypoint = LoadTypedIsoLiteralText::new(entrypoint_declaration);
+                match entrypoints.entry(client_field_id) {
+                    Entry::Occupied(occupied_entry)
+                        if (occupied_entry.get().is_lazy_loaded
```
rbalicki2, line 58: let's do `Entry::Occupied(occupied_entry) { if (...) { ... }  }` and let's not do `_` but do `Entry::Vacant(vacant) => vacant.insert(...)`

rbalicki2 (changes requested):

crates/isograph_schema/src/validate_entrypoint.rs
dialectician4: Screenshot_20250504_165204

Unfortunately this happens without explicitly implementing the trait.

crates/isograph_schema/src/validate_entrypoint.rs
```diff
@@ -1,34 +1,76 @@
...
 use crate::{NetworkProtocol, Schema};
 
+#[derive(Debug)]
+pub struct LoadTypedIsoLiteralText {
+    pub iso_literal_text: IsoLiteralText,
+    pub is_lazy_loaded: bool,
+}
+
+impl LoadTypedIsoLiteralText {
+    pub fn new(entrypoint_declaration: WithSpan<EntrypointDeclaration>) -> Self {
```
rbalicki2, line 24: I think we can just inline this, there's no need to have a separate function here IMO

PR #512, opened 2025-05-08
wrap `Link` output type in `type_annotation`

PatrykWalach:
Part of #53


rbalicki2 (approved):
looks correct to me shipittt

PR #515, opened 2025-05-08
add client pointers runtime

PatrykWalach:
I will add support for lists as a follow up
Part of #53

### DIFF
--- a/libs/isograph-react/src/core/read.ts
+++ b/libs/isograph-react/src/core/read.ts
@@ -1,4 +1,4 @@
-import { CleanupFn } from '@isograph/disposable-types';
+import { CleanupFn, type ItemCleanupPair } from '@isograph/disposable-types';
 import {
   getParentRecordKey,
   insertIfNotExists,
@@ -196,6 +196,7 @@ function readData<TReadFromStore>(
           storeRecord,
           root,
           variables,
+          nestedRefetchQueries,
           networkRequest,
           (ast, root) =>
             readData(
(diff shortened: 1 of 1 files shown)


libs/isograph-react/src/core/read.ts
```diff
@@ -797,6 +798,104 @@ export function readLinkedFieldData(
     };
   }
   const targetId = link;
+  if (typeof field.refetchQuery === 'number') {
+    root = link;
```
rbalicki2, line 802: nit: this is a bit confusing. Maybe it will be a bit clearer to define a new variable, `clientPointerTargetId` or somthing?

libs/isograph-react/src/core/read.ts
```diff
@@ -797,6 +798,104 @@ export function readLinkedFieldData(
     };
   }
   const targetId = link;
+  if (typeof field.refetchQuery === 'number') {
```
rbalicki2, line 801: nit: `refetchQuery != null` I think is more idiomatic

can you add a comment saying `// if refetchQuery is not null, then the field is a client pointer`. Can you also add that to `ReaderLinkedField`?

rbalicki2 (approved):
LGTM! Just a few nits, but I can fix them on main

libs/isograph-react/src/core/read.ts
rbalicki2: well, we have that with `targetId` already!

PR #522, opened 2025-06-27
fix clippy

PatrykWalach:


rbalicki2:
Thanks! I hope this was an auto fix

PR #525, opened 2025-07-25
Singleton/Value traits + getters/setters on db

rbalicki2:
- **wip**
- **Add methods to db:**


crates/pico/src/source.rs
```diff
@@ -9,6 +9,10 @@ pub trait Source {
     fn get_key(&self) -> Key;
 }
 
+/// A marker trait that is implemented by the #[derive(Singleton)]
+/// proc-macro.
+pub trait Singleton {}
```
ch1ffa: What if we do:
```rs
pub trait Singleton: Source {
    fn get_key() -> Key;
}
```
and use basically the same macro as for `get_key(&self)` in `Source`?
Then we can:
- use `db.set` for both source and singleton
- use singleton type to get keys like `<T as Singleton>::get_key()` for generics of `SchemaFile::get_key()` for implementations.

ch1ffa (commented):
Looks awesome! Just a nit we discussed on a call

PR #526, opened 2025-07-26
Intern CurrentWorkingDirectory as a singleton

rbalicki2:
- Remove current_working_directory from the config
- Intern it as a singleton, and read it when necessary
- In functions that need it, but don't receive a db parameter, pass it along
- What's next? Interning the config itself. But it turned out that in many places, we didn't actually need the config. Instead, we needed just the current working directory.

- Break some code in the language server with todo!()'s. YOLO, that will be fixed when we finally have the incremental rewrite finished.


ch1ffa (approved):
YES! It was annoying to pass config everywhere. Now if we have `db` why not just query for it and have a dependency by doing it? Nice!

PR #530, opened 2025-07-28
derive db macro example

ch1ffa:
Here is an example how we can implement `#[derive(Db)]` on some struct and use it as a `Database`. This allows us to add new methods and fields, which can be used inside memoized functions.

### DIFF
--- a/crates/pico/src/memo_ref.rs
+++ b/crates/pico/src/memo_ref.rs
@@ -2,19 +2,36 @@ use std::{marker::PhantomData, ops::Deref};
 
 use intern::InternId;
 
-use crate::{Database, DerivedNodeId, ParamId};
+use crate::{DatabaseDyn, DerivedNodeId, ParamId};
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug)]
 pub struct MemoRef<T> {
-    pub(crate) db: *const Database,
+    pub(crate) db: *const dyn DatabaseDyn,
     pub(crate) derived_node_id: DerivedNodeId,
     phantom: PhantomData<T>,
 }
 
+impl<T> Copy for MemoRef<T> {}
+
+impl<T> Clone for MemoRef<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T> PartialEq for MemoRef<T> {
+    fn eq(&self, other: &Self) -> bool {
+        std::ptr::eq(self.db, other.db) && self.derived_node_id == other.derived_node_id
+    }
+}
+
+impl<T> Eq for MemoRef<T> {}
+
+#[allow(clippy::unnecessary_cast)]
 impl<T: 'static + Clone> MemoRef<T> {
-    pub fn new(db: &Database, derived_node_id: DerivedNodeId) -> Self {
+    pub fn new(db: &dyn DatabaseDyn, derived_node_id: DerivedNodeId) -> Self {
         Self {
-            db,
+            db: db as *const _ as *const dyn DatabaseDyn,
             derived_node_id,
             phantom: PhantomData,
         }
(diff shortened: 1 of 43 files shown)


rbalicki2:
Love it! Let's ship it. I'm a fan of it as it is. I haven't reviewed the code in detail, but I played around with the branch, and really really like that we can do `isograph_db.get_current_working_directory()` etc

That makes me think we don't need singletons! Instead, we can just have a field `isograph_db.config_source_id` etc. But anyway, amazing, this is awesome, let's ship it

rbalicki2:
One question I have is — is there any way to prevent us from relying on other fields in the DB? For example, `#[memo] fn foo(db: &IsographDb) { return db.foo }` would not be recalculated if the value of `db.foo` changed between calls.

If not, then we should just add some comments somewhere to indicate that that will break things

Obviously, you *can* rely on some stuff, like `db.config_source_id`. That's safe to read from!

rbalicki2:
Actually I wonder if we can replace the pub items with getters/setters so they can be tracked. I'm just spitballing absolutely no reason to do that in v1

rbalicki2:
Are we just converging on salsa lol

ch1ffa:
That is a good question. I think your example is valid. We can use this data if it's immutable or we don't care if it can change. It must be clear that after changing the data outside pico storage one may need to invalidate the storage.

It's totally ok to store:
- singleton ids (or use new functions to access it by type)
- source ids if keys are stable
- something append only, that we read only between compilations

I don't think getters/setters would help in every situation. Imagine someone is using RefCell and writing new value using getter in the middle of calculation - it's totally cursed.

rbalicki2:
Good point re: interior mutability. But you can already store a refcell in the DB and do that 😭

ch1ffa:
Oh yeah, interior mutability for memo should be forbidden completely. For user fields it could be legit if we don't read it during compilation.

Actually, getter/setter in combination could give us info that a user field has been read after write in the same epoch.

ch1ffa:
I mean, do we need to track user fields as a dependency? Maybe just have a check how it is accessed.

rbalicki2:
Not sure I follow. I would imagine that if there is `IsographDb { storage: Storage, is_cool: bool }` -> and we call `#[memo] memo_func(db) -> bool { db.get_is_cool() }`, we would want to track that `is_cool` was accessed (i.e. as a dependency), then calling `db.set_is_cool` is essentially a `db.set` call (i.e. an update to the source)?

If we allow you to mutate `is_cool` without going through a setter, we have to treat `is_cool` as always invalidated.

Anyway, perhaps we should keep all the sources on the DB?

rbalicki2:
```
struct IsographDb { storage, iso_literals: HashMap<FileName, String> }

#[memo]
fn get_parsed_iso_literal(db, file_name) -> AST {
  parse(get_iso_literal_at_filename(file_name))
}

#[memo]
fn get_iso_literal_at_filename(db, file_name) -> String {
  db.get_iso_literals().get(file_name).unwrap()
}

#[memo]
fn parse(db, ast) {
  // ...
}
```

it just works, right!??

An unrelated isograph literal changes and we call `get_parsed_iso_literal` again. We have to recalculate `get_iso_literal_at_filename`, but that returns the same value, so we can short circuit and not recalculate `get_parsed_iso_literal`

ch1ffa:
Yeah, no way if we don't increment epoch via db.set, sources should be in the db. I'm just arguing that a field outside storage should not be a source. Even if we introduce getter/setter, a field is still there and one can access it directly (if we use derive macro, we can't rewrite a structure, only add something). Salsa generates getter/setter for interned struct because that's how they do tracking. We rewrite fn output to MemoRef, they require to do it manually by returning interned value from fn and optionally short circuit over fn params.

rbalicki2:
okay. Maybe in the future I'll play around with changing the derive to something like
```
make_db!(struct Thermometer { temp: usize })
```
so we can rewrite the struct with getters/setters

But yeah, understood. Okay, so we should just not store anything besides sources on the IsographDb. Sgtm

ch1ffa:
Ah, you want to define sources on db struct rather than something detached. Yeah, that seems convenient. I'm also thinking we need somehow track hashmap items, maybe have our own tracked hashmap or whatnot

rbalicki2:
yeah it's a bit unfortunate to have to go through another memoized function just to use a hashmap

crates/pico/src/memo_ref.rs
```diff
@@ -2,19 +2,36 @@ use std::{marker::PhantomData, ops::Deref};
...
+
+impl<T> Eq for MemoRef<T> {}
+
+#[allow(clippy::unnecessary_cast)]
 impl<T: 'static + Clone> MemoRef<T> {
-    pub fn new(db: &Database, derived_node_id: DerivedNodeId) -> Self {
+    pub fn new(db: &dyn DatabaseDyn, derived_node_id: DerivedNodeId) -> Self {
         Self {
-            db,
+            db: db as *const _ as *const dyn DatabaseDyn,
```
ch1ffa, line 34: This is suspicious. Clippy says that it's unnecessary cast, but without it we got lifetime issues

rbalicki2: sgtm

PR #538, opened 2025-08-04
allow memo on methods

ch1ffa:
A new version of `memo` macro that allows us to apply it on methods as follows:
- if argument `db` is provided, e.g. `#[memo(db = my_db)` then we try to use this argument as a database. Unfortunately we don't know a concrete `Database` implementation so we can find it only by name
- if there is no `db` macro argument and the first argument is `&self` we try to use it as database
- if memoized function is a free function, then we use its first argument as database

one can use `#[memo(db)]` if name of the argument is db, `#[memo(db = test_db)]` or `#[memo(db = "test_db")]`

### DIFF
--- a/crates/pico/tests/memo_on_db_method.rs
+++ b/crates/pico/tests/memo_on_db_method.rs
@@ -1,0 +1,48 @@
+use std::sync::atomic::{AtomicUsize, Ordering};
+
+use pico::{Database, SourceId, Storage};
+use pico_macros::{memo, Db, Source};
+
+static FIRST_LETTER_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+#[derive(Db, Default)]
+struct TestDatabase {
+    pub storage: Storage<Self>,
+}
+
+impl TestDatabase {
+    #[memo]
+    fn first_letter(&self, input_id: SourceId<Input>) -> char {
+        FIRST_LETTER_COUNTER.fetch_add(1, Ordering::SeqCst);
+        let input = self.get(input_id);
+        input.value.chars().next().unwrap()
+    }
+}
+
+#[test]
+fn memo_on_db_method() {
+    let mut db = TestDatabase::default();
+
+    let input_id = db.set(Input {
+        key: "key",
+        value: "asdf".to_string(),
+    });
+
+    assert_eq!(*db.first_letter(input_id), 'a');
+    assert_eq!(FIRST_LETTER_COUNTER.load(Ordering::SeqCst), 1);
+
+    db.set(Input {
+        key: "key",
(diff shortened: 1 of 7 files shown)


crates/pico/tests/memo_on_db_method.rs
```diff
@@ -0,0 +1,48 @@
...
+
+static FIRST_LETTER_COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+#[derive(Db, Default)]
+struct TestDatabase {
+    pub storage: Storage<Self>,
+}
+
+impl TestDatabase {
+    #[memo]
```
rbalicki2, line 14: YES

crates/pico/tests/memo_on_struct_method.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Db, Default)]
+struct TestDatabase {
+    pub storage: Storage<Self>,
+}
+
+struct TestStruct;
+
+impl TestStruct {
+    #[memo(db = test_db)]
+    fn first_letter(&self, test_db: &TestDatabase, input_id: SourceId<Input>) -> char {
```
rbalicki2, line 18: nifty

rbalicki2: what if we access `self` here? That would break memoization, right?

I suppose that's true of methods on structs that implements Database, but maybe then it's clear that you are only supposed to store sources on that struct...

crates/pico_macros/src/memo_macro.rs
```diff
@@ -1,42 +1,144 @@
...
+            .position(|arg| {
+                matches!(arg, FnArg::Receiver(rcv) if rcv.reference.is_some())
+                    || matches!(arg, FnArg::Typed(_))
+            })
+            .unwrap_or(0)
+    };
+
+    let (db_arg, closure_db_arg) = match &sig.inputs[db_pos] {
+        FnArg::Receiver(rcv) => {
+            if rcv.reference.is_none() {
```
rbalicki2, line 89: should we disallow `&mut self` in the macro? It looks like we rewrite it to be `&self` regardless, which makes for confusing errors:

```
    #[memo]
    fn first_letter_2(&mut self, input_id: SourceId<Input>) -> char {
        self.storage.set(Input {
            key: todo!(),
            value: todo!(),
        });
       
```

```
      #[memo]
|     ^^^^^^^ `__db` is a `&` reference, so the data it refers to cannot be borrowed as mutable
```

rbalicki2 (changes requested):
Overall looks great! Just some minor nits.

PR #558, opened 2025-08-14
replace usages of removed `bacon -p` with `watch`

XiNiHa:
This PR updates the `bacon` scripts to stop using the [removed](https://github.com/Canop/bacon/pull/277) `-p` option.

### DIFF
--- a/bacon.toml
+++ b/bacon.toml
@@ -13,14 +13,17 @@ command = [
 	"cargo", "build",
 	"--color", "always",
 ]
+watch = ["crates"]
 need_stdout = false
 
 [jobs.watch-libs]
 command = ["pnpm", "compile-libs"]
+watch = ["libs"]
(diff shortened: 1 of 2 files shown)


bacon.toml
```diff
@@ -122,7 +125,8 @@ b = "job:build"
 
 [jobs.build-swc]
 command = [
-	"cargo", "build", "-p", "swc_isograph_plugin", "--release", "--target", "wasm32-wasip1", 
+	"cargo", "build", "-p", "swc_isograph_plugin", "--release", "--target", "wasm32-wasip1",
 	"--color", "always",
 ]
+watch = ["crates"]
```
rbalicki2, line 131: good catch... this should probably be `-p crates` instead of just watching the swc_isograph_plugin folder

PR #565, opened 2025-08-15
chore: replace rm with rimraf

### LINKED ISSUE #398: Watch mode seems to not work in windows (without a Linux terminal?)

PatrykWalach:
I've had enough of switching into git terminal

Closes #398


rbalicki2:
nice, thanks for doing this!

PR #567, opened 2025-08-16
add tree-sitter grammar

XiNiHa:


tree-sitter/src/grammar.json
```diff
@@ -0,0 +1,764 @@
...
+    "entrypoint_declaration": {
+      "type": "SEQ",
+      "members": [
+        {
+          "type": "STRING",
+          "value": "entrypoint"
+        },
+        {
+          "type": "SYMBOL",
+          "name": "field_name"
```
rbalicki2: `parent_object_entity_name_and_selectable_name` is a better name than `field_name`. field is the selectable part, and I am trying to use `selectable` everywhere instead of `field` (since `field` is a scalar selectable, and pointer is an object selectable, so using field to refer to both is unclear)

tree-sitter/src/grammar.json
```diff
@@ -0,0 +1,764 @@
...
+          "name": "selection_set"
+        }
+      ]
+    },
+    "field_name": {
+      "type": "SEQ",
+      "members": [
+        {
+          "type": "FIELD",
+          "name": "parent_type",
```
rbalicki2: `parent_object_entity_name`

tree-sitter/src/grammar.json
```diff
@@ -0,0 +1,764 @@
...
+            "name": "identifier"
+          }
+        },
+        {
+          "type": "STRING",
+          "value": "."
+        },
+        {
+          "type": "FIELD",
+          "name": "name",
```
rbalicki2: `selectable_name`

tree-sitter/src/grammar.json
```diff
@@ -0,0 +1,764 @@
...
+              "type": "BLANK"
+            }
+          ]
+        },
+        {
+          "type": "CHOICE",
+          "members": [
+            {
+              "type": "STRING",
+              "value": ","
```
rbalicki2: `,` are no longer allowed by the grammar, you have to have a line break here

tree-sitter/src/grammar.json
```diff
@@ -0,0 +1,764 @@
...
+                  "type": "SYMBOL",
+                  "name": "argument"
+                },
+                {
+                  "type": "REPEAT",
+                  "content": {
+                    "type": "SEQ",
+                    "members": [
+                      {
+                        "type": "STRING",
```
rbalicki2: also, commas not allowed, you must have a line break here

tree-sitter/src/grammar.json
```diff
@@ -0,0 +1,764 @@
...
+                  "name": "object_field"
+                },
+                {
+                  "type": "REPEAT",
+                  "content": {
+                    "type": "SEQ",
+                    "members": [
+                      {
+                        "type": "STRING",
+                        "value": ","
```
rbalicki2: commas not allowed, need line break

(I need to consider what I want long term but that is the current state of the parser)

XiNiHa:
> I think many of these files are generated, yeah?
> 
> i.e. I shouldn't try to review `node-types.json` and `parser.c`, right? What about the other files?

The linguist config is set up so you only see the diffs for non-generated files if you review on the GitHub Web UI, so basically it's just a `grammar.js` and some tests

rbalicki2:
Ah okay, well then I'm glad I reviewed something that's pretty 1:1 with the grammar js file :-D

XiNiHa:
Explicitly handling newlines in specific points while skipping in most cases requires writing a custom scanner for handling this; it'll take some time to get it working I guess 😅

PR #568, opened 2025-08-16
add Zed extension (grammar only)

XiNiHa:
Stacked on https://github.com/isographlabs/isograph/pull/567

LSP support should also be investigated soon, but the current implementation is already pretty helpful!

스크린샷 2025-08-17 오전 12 28 52


rbalicki2:
Overall, LGTM. Can you either make this part of the same workspace (i.e. move everything into `./crates`, or modify the root `Cargo.toml` to include `./zed-extensions`)? Or alternatively, if we want to keep it separate (but why would we? Lots of code is obviously shareable) can we ensure that the zed code builds, formats, etc during CI?

XiNiHa:
> Can you either make this part of the same workspace

I was kinda worried about Zed not supporting that use case since I've seen none of the extensions published in a workspace before 😅 Just did a test and can confirm it also works!

PR #571, opened 2025-08-18
refactor: nextjs installtion without turbopack - issue #541

jcesarprog:
add `--no-turbopack` flag so NextJS wont ask for turbopack during its installation


rbalicki2:
LGTM shipit!

PR #573, opened 2025-08-18
feat(build): update turbo tasks, scripts forf better dev experience

### LINKED ISSUE #569: Watch the isograph binary and restart the compiler/lsp if it changes?
This would make LSP/watch mode development a lot smoother

PatrykWalach:
I've noticed running pet demo is very complicated, I've needed like 5 terminals,
- backend
- frontend
- watch build cargo
- watch iso
- watch libs

plus ideally swc plugin needs to be watched as well.

After this PR:
- `pnpm -w watch-pet-demo` - runs: watch build cargo and watch iso. 
- `pnpm -w dev-pet-demo` - runs: watch build cargo, watch iso, backend and frontend. 

Closes #569


PatrykWalach:
I think we don't need `iso-watch` in `package.jsons` anymore, what do you think?

rbalicki2:
yeah, as long as `pnpm watch-pet-demo` etc do the right thing, we can remove `iso-watch` from the demos

rbalicki2:
awesome, I'll check it out locally, but I'm on board with this. Can you also update development-workflow.md if needed?

PatrykWalach:
I've tested it works as expected.

Alos right now all the logs are mixed, we can split it if necessary with extra args `--ui=tui`
image

rbalicki2:
soooo good. I'll ship it and land `--ui=tui`

PR #580, opened 2025-08-22
Fix NOT_SET checks

XiNiHa:
This PR makes `NOT_SET` a unique symbol instead of a `Symbol`, allowing for narrowing the types using conditionals.


rbalicki2:
Shipittt

PR #583, opened 2025-08-23
fix LSP on windows

PatrykWalach:
`pathdiff::diff_paths` doesn't support windows path starting with `\\?\`

### DIFF
--- a/crates/common_lang_types/src/location.rs
+++ b/crates/common_lang_types/src/location.rs
@@ -1,6 +1,9 @@
-use std::{error::Error, fmt, path::PathBuf};
-
 use intern::string_key::{Intern, Lookup};
+use std::{
+    error::Error,
+    fmt,
+    path::{Path, PathBuf},
+};
 
 use crate::{
     text_with_carats::text_with_carats, CurrentWorkingDirectory, RelativePathToSourceFile, Span,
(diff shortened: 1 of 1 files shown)


crates/common_lang_types/src/location.rs
```diff
@@ -225,11 +228,32 @@ impl<T> From<WithEmbeddedLocation<T>> for WithLocation<T> {
...
+    path.to_path_buf()
+}
+
+pub fn diff_paths_with_prefix<P, B>(path: P, base: B) -> Option<PathBuf>
+where
+    P: AsRef<Path>,
+    B: AsRef<Path>,
+{
+    let path_ref = path.as_ref();
+    let clean_path = strip_windows_long_path_prefix(path_ref);
```
rbalicki2: can we only incur the cost of cloning on windows by:

```
if cfg!(target_os="windows") {
  let clean_path = strip_prefix(...)
  return pathdiff::diff_paths(&clean_path, base);
}
```

?

But lgtm otherwise

crates/common_lang_types/src/location.rs
```diff
@@ -225,11 +228,31 @@ impl<T> From<WithEmbeddedLocation<T>> for WithLocation<T> {
     }
 }
 
+pub fn strip_windows_long_path_prefix(path: &Path) -> &Path {
```
PatrykWalach, line 231: I need this pub for goto definition

crates/common_lang_types/src/location.rs
PatrykWalach: There's `strip_prefix` on `&Path` so I think it's perfect now

Issue #556, opened 2025-08-12
LSP GotoDefinition


rbalicki2:
- first, go to definition on entrypoint client field name, and parent type names (client field, client pointer, entrypoint)
- second, for selection sets

PR #586, opened 2025-08-23
[LSP] goto definition

PatrykWalach:
#556

### DIFF
--- a/crates/common_lang_types/src/location.rs
+++ b/crates/common_lang_types/src/location.rs
@@ -1,9 +1,5 @@
 use intern::string_key::{Intern, Lookup};
-use std::{
-    error::Error,
-    fmt,
-    path::{Path, PathBuf},
-};
+use std::{error::Error, fmt, path::PathBuf};
 
 use crate::{
     text_with_carats::text_with_carats, CurrentWorkingDirectory, RelativePathToSourceFile, Span,
@@ -228,31 +224,11 @@ impl<T> From<WithEmbeddedLocation<T>> for WithLocation<T> {
     }
 }
 
-pub fn strip_windows_long_path_prefix(path: &Path) -> &Path {
-    #[cfg(target_os = "windows")]
-    {
-        if let Ok(stripped) = path.strip_prefix(r"\\?\") {
-            return stripped;
-        }
-    }
-    path
-}
-
-pub fn diff_paths_with_prefix<P, B>(path: P, base: B) -> Option<PathBuf>
-where
-    P: AsRef<Path>,
-    B: AsRef<Path>,
-{
-    let clean_path = strip_windows_long_path_prefix(path.as_ref());
-
-    pathdiff::diff_paths(clean_path, base)
-}
-
 pub fn relative_path_from_absolute_and_working_directory(
     current_working_directory: CurrentWorkingDirectory,
     absolute_path: &PathBuf,
 ) -> RelativePathToSourceFile {
-    diff_paths_with_prefix(
+    pathdiff::diff_paths(
         absolute_path,
         PathBuf::from(current_working_directory.lookup()),
     )
--- a/crates/isograph_lsp/src/goto_definition.rs
+++ b/crates/isograph_lsp/src/goto_definition.rs
@@ -1,0 +1,170 @@
+use crate::format::char_index_to_position;
+use crate::hover::get_iso_literal_extraction_from_text_position_params;
+use crate::{
+    lsp_runtime_error::LSPRuntimeResult, semantic_tokens::delta_line_delta_start,
+    uri_file_path_ext::UriFilePathExt,
+};
+use common_lang_types::{
+    diff_paths_with_prefix, relative_path_from_absolute_and_working_directory,
+    strip_windows_long_path_prefix, SelectableName, ServerObjectEntityName, Span,
+};
+use isograph_compiler::CompilerState;
+use isograph_compiler::{
+    extract_iso_literals_from_file_content, get_validated_schema, process_iso_literal_extraction,
+    read_iso_literals_source_from_relative_path, IsoLiteralExtraction,
+};
+use isograph_lang_types::{
+    DefinitionLocation, Description, IsographResolvedNode, VariableDefinition,
+};
+use isograph_schema::{
+    get_parent_and_selectable_for_object_path, get_parent_and_selectable_for_scalar_path,
+    SchemaSource, SelectableTrait, ServerEntityName,
+};
+use isograph_schema::{IsographDatabase, ServerScalarSelectable};
+use isograph_schema::{NetworkProtocol, StandardSources};
+use lsp_types::request::GotoDefinition;
+use lsp_types::request::Request;
+use lsp_types::GotoDefinitionResponse;
+use lsp_types::Location;
+use lsp_types::LocationLink;
+use lsp_types::Range;
+use lsp_types::{request::HoverRequest, Hover, HoverContents, MarkupContent, MarkupKind};
+use lsp_types::{Position, TextDocumentIdentifier, TextDocumentPositionParams, Uri};
+use pico_macros::memo;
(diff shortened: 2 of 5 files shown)


crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -0,0 +1,153 @@
...
+    Ok(goto_location)
+}
+
+#[memo]
+fn server_definition_location<TNetworkProtocol: NetworkProtocol + 'static>(
+    db: &IsographDatabase<TNetworkProtocol>,
+    span: &Option<Span>,
+) -> Option<Location> {
+    let binding =
+        strip_windows_long_path_prefix(&db.get_isograph_config().schema.absolute_path.clone());
```
rbalicki2: you probably need to call `db.get_isograph_config()` on a prior line

crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -0,0 +1,153 @@
...
+    params: <GotoDefinition as Request>::Params,
+) -> LSPRuntimeResult<<GotoDefinition as Request>::Result> {
+    let db = &compiler_state.db;
+    Ok(on_goto_definition_impl(
+        db,
+        params.text_document_position_params.text_document.uri,
+        params.text_document_position_params.position,
+    )
+    .to_owned()?
+    .map(GotoDefinitionResponse::Scalar))
```
PatrykWalach, line 39: `GotoDefinitionResponse` doesn't implement `Eq`

crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    let StandardSources {
+        schema_source_id, ..
+    } = db.get_standard_sources();
+
+    let SchemaSource { content, .. } = db.get(*schema_source_id);
+
+    Some(Location {
+        uri: uri.ok()?,
+        range: span
```
rbalicki2: can we use `selectable.name.location`?

crates/isograph_lsp/src/goto_definition.rs
rbalicki2: LOL laame

rbalicki2: lsp_types is kinda wack

crates/isograph_lsp/src/goto_definition.rs
PatrykWalach: I don't need clone anymore

crates/isograph_lsp/src/goto_definition.rs
PatrykWalach: I'm using `selectable.name.location.span()`, do you mean something else?

rbalicki2: Yes, I mean something like 
```
// generated locations have no location, we can just bail early
let EmbeddedLocation { text_source, span: location_span } = location.and_then(|loc| loc.as_embedded())?;
```
use `text_source` and `location_span`, instead of the schema path. So, to get the text you care about:

- from text source, we get the relative path. We can read the content of that file from the DB.
- we then use `text_source.span` to get the span within the document that we care about. `text_source.span` is `None` for graphql docs, we care about the whole doc. But they're `Some` for iso literals.
- Then apply the `location_span` to get the text you care about in the text source content.

But don't want to read the content for the name, we actually want to get the line/char to go to, so we have to modify ☝️  accordingly.

crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -0,0 +1,181 @@
...
+                Cow::Borrowed(path)
+            };
+
+            let uri = Uri::from_str(&format!("file:///{normalized_path}")).ok()?;
+
+            let (_, content) = TextSource {
+                span: None,
+                ..location.text_source
+            }
+            .read_to_string();
```
PatrykWalach: My guess is this won't get invalidated like it's supposed to

crates/common_lang_types/src/location.rs
```diff
@@ -228,31 +224,11 @@ impl<T> From<WithEmbeddedLocation<T>> for WithLocation<T> {
     }
 }
 
-pub fn strip_windows_long_path_prefix(path: &Path) -> &Path {
-    #[cfg(target_os = "windows")]
-    {
-        if let Ok(stripped) = path.strip_prefix(r"\\?\") {
```
PatrykWalach, line 234: Turns out this is not required after I did the `canonize`🤔

crates/isograph_lsp/src/goto_definition.rs
PatrykWalach: Somehow it's working

crates/common_lang_types/src/location.rs
rbalicki2: NICE

crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -0,0 +1,181 @@
+use crate::format::char_index_to_position;
+use crate::hover::get_iso_literal_extraction_from_text_position_params;
+use crate::{lsp_runtime_error::LSPRuntimeResult, uri_file_path_ext::UriFilePathExt};
+use common_lang_types::{
+    relative_path_from_absolute_and_working_directory, Location, Span, TextSource,
+};
+use intern::string_key::Lookup;
+use isograph_compiler::CompilerState;
+use isograph_compiler::{get_validated_schema, process_iso_literal_extraction};
+use isograph_lang_types::{DefinitionLocation, IsographResolvedNode};
```
rbalicki2, line 10: I'm not sure why, but your rust analyzer doesn't merge imports :/ anyway nbd I'll merge them haha

crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -0,0 +1,181 @@
...
+                if let Ok((_, selectable)) =
+                    get_parent_and_selectable_for_scalar_path(&scalar_path, validated_schema)
+                {
+                    match selectable {
+                        DefinitionLocation::Server(server_selectable) => {
+                            server_definition_location(db, &server_selectable.name.location)
+                                .to_owned()
+                        }
+                        DefinitionLocation::Client(client_selectable) => {
+                            server_definition_location(db, &client_selectable.name.location)
```
rbalicki2: lets rename that fn, since it's not just for server definitions, and put it somewhere else. It's not specific to goto def (e.g. find all references will use it, maybe we can incorporate this in hover.)

crates/isograph_lsp/src/goto_definition.rs
rbalicki2: We shouldn't ever read from the file system in a memoized function. e.g. this is probably broken if you, without saving, move a client field from one file to another (or from one location in a file to another)

rbalicki2 (commented):
still reviewing, but overall looks 🔥

crates/isograph_lsp/src/goto_definition.rs
PatrykWalach: What should I use? `get_open_file_map`?

rbalicki2: That shouldn't be necessary to use directly. `read_iso_literals_source_from_relative_path` is what you want, which goes through the open file map

PatrykWalach: Yeah, this doesn't work for server stuff in schema

rbalicki2: oh I see, yeah. We should fix that (i.e. have an open file map for graphql schema files).

We should use `read_iso_literals_source_from_relative_path` is its a client field, otherwise, we can read from the standard sources content for server fields.

(For GraphQL server fields, we don't have an "open file map", but we should fix that. So, unsaved edits to the GraphQL schema aren't reflected in the LSP)

PR #587, opened 2025-08-24
Emit NormalizationAstLoader for `@lazyLoad` entrypoints

### LINKED ISSUE #43: [epic] Deferred normalization ASTs

XiNiHa:
Closes #43

### DIFF
--- a/crates/generate_artifacts/src/entrypoint_artifact.rs
+++ b/crates/generate_artifacts/src/entrypoint_artifact.rs
@@ -5,15 +5,18 @@ use common_lang_types::{
     QueryText, ServerObjectEntityName, VariableName,
 };
 use isograph_config::GenerateFileExtensionsOption;
-use isograph_lang_types::{DefinitionLocation, ScalarSelectionDirectiveSet, SelectionType};
+use isograph_lang_types::{
+    DefinitionLocation, EmptyDirectiveSet, EntrypointDirectiveSet, LazyLoadDirectiveParameters,
+    LazyLoadDirectiveSet, ScalarSelectionDirectiveSet, SelectionType,
+};
 use isograph_schema::{
     create_merged_selection_map_for_field_and_insert_into_global_map,
     current_target_merged_selections, get_imperatively_loaded_artifact_info,
     get_reachable_variables, initial_variable_context, ClientScalarOrObjectSelectable,
-    ClientScalarSelectable, FieldToCompletedMergeTraversalStateMap, FieldTraversalResult, Format,
-    MergedSelectionMap, NetworkProtocol, RootOperationName, RootRefetchedPath,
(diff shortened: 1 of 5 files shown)


crates/generate_artifacts/src/entrypoint_artifact.rs
```diff
@@ -37,12 +39,19 @@ struct EntrypointArtifactInfo<'schema, TNetworkProtocol: NetworkProtocol> {
     normalization_ast_text: NormalizationAstText,
     refetch_query_artifact_import: RefetchQueryArtifactImport,
     concrete_type: ServerObjectEntityName,
+    lazy_load_config: Option<EntrypointLazyLoadConfig>,
```
rbalicki2: let's store the `EntrypointDirectiveSet` directly here, and default to `EntrypointDirectiveSet::None` instead of passing `None` below

demos/pet-demo/src/components/HomeRoute.tsx
```diff
@@ -27,8 +27,9 @@ export const HomeRoute = iso(`
 
 export function HomeRouteLoader() {
   const { fragmentReference } = useLazyReference(
-    iso(`entrypoint Query.HomeRoute`),
+    iso(`entrypoint Query.HomeRoute @lazyLoad`),
```
rbalicki2: instead of adding it here, let's add it to the entrypoint in `FavoritePhrase.tsx`

rbalicki2 (approved):
shipitttt!!!! Lets goooo!!!

Lazy normalization ASTs are here!!! We officially support something that Relay doesn't 🎉

PR #589, opened 2025-08-24
verify deps before run

PatrykWalach:
This is super cool flag, whenever you run any command, it will first check if node_modules are up to date with lockfile and update if necessary.
It's all super fast and seemless.

You never again have to run `pnpm i`.

This is especially cool if you have a bot to update the dependecies, you can just `git pull` `git checkout` and run whatever you want without worry your deps are out of date.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -2,3 +2,4 @@ link-workspace-packages = deep
 prefer-workspace-packages = true
 enable-pre-post-scripts = true
 save-workspace-protocol = false
+verify-deps-before-run = install
--- a/package.json
+++ b/package.json
@@ -55,5 +55,5 @@
   "engines": {
     "node": "22.9.0"
   },
-  "packageManager": "pnpm@9.11.0"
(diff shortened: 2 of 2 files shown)


rbalicki2:
Niiiice.

If I run `pnpm dev-pet-demo`, should I see `pnpm i` get run?

```
# I tried this both at the root and in demos/pet-demo
# on verify-deps
git checkout $newBranch
pnpm add svelte
git add . && git commit -m 'new dep'
git checkout verify-deps
rm -rf node_modules
pnpm i
git checkout $newBranch
pnpm dev-pet-demo
```

Result: no svelte

PatrykWalach:
Oh I see, it doesn't work in pnpm@9, only in pnpm@10

PatrykWalach:
I've updated pnpm and tested, changed typescript 5.6.3 -> 5.6.2 in package.json and it run install before running `pnpm format`

PR #594, opened 2025-08-24
replace WithSpan with WithLocation for client names

PatrykWalach:


crates/isograph_lang_types/src/client_field_declaration.rs
```diff
@@ -44,7 +44,7 @@ pub struct ClientPointerDeclaration {
     // impl<T: ResolvePosition> ResolvePosition for GraphQLTypeAnnotation<T>?
     pub target_type: GraphQLTypeAnnotation<UnvalidatedTypeName>,
     #[resolve_field]
-    pub client_pointer_name: WithSpan<ClientObjectSelectableNameWrapper>,
+    pub client_pointer_name: WithLocation<ClientObjectSelectableNameWrapper>,
```
PatrykWalach, line 49: Is `#[resolve_field]` still working here?

rbalicki2: oops, no :)

rbalicki2: we can handle this adding handling for `WithLocation` in crates/resolve_position_macros/src/resolve_position_macro.rs, lines 209 and 185. (Note that this was written by claude, so it's a bit verbose)

For the purposes of resolve position: Spans are correct. Locations aren't. Basically, if we have only spans, this implies that everything comes from the same document. But locations imply that they might come from different documents.

So let's add handling in another PR, and just include a note about how this is bad.

rbalicki2 (approved):
lgtm if it passes CI lets shipit!

PR #598, opened 2025-08-26
feat(macros): refactor field parsing and codegen for resolve macro

PatrykWalach:
Now it's recursive and as result parses Option<Vec<...>> Vec<Option<...>> Vec<Vec<>> etc.

This makes adding support for WithLocation easier

### DIFF
--- a/crates/resolve_position_macros/src/resolve_position_macro.rs
+++ b/crates/resolve_position_macros/src/resolve_position_macro.rs
@@ -159,7 +159,7 @@ enum ResolveFieldInfoType {
 
 enum ResolveFieldInfoTypeWrapper {
     None(Box<ResolveFieldInfoType>),
-    Vec(Box<ResolveFieldInfoTypeWrapper>),
+    IteratorWrapper(Box<ResolveFieldInfoTypeWrapper>),
 }
 
 struct ResolveFieldInfo {
@@ -210,7 +210,9 @@ fn parse_resolve_field_type(
                 // Recursively parse the inner type
                 let inner_wrapper = parse_resolve_field_type(inner_path, generics_map)?;
 
-                return Ok(ResolveFieldInfoTypeWrapper::Vec(Box::new(inner_wrapper)));
+                return Ok(ResolveFieldInfoTypeWrapper::IteratorWrapper(Box::new(
+                    inner_wrapper,
+                )));
             }
         }
     }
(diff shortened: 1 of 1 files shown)


crates/resolve_position_macros/src/resolve_position_macro.rs
```diff
@@ -191,60 +249,63 @@ fn get_resolve_field_info(
...
+                    return #field_expr.item.resolve(new_parent, position);
+                }
+            },
+        },
+
+        ResolveFieldInfoTypeWrapper::Option(inner) => {
+            let inner_code = generate_resolve_code_recursive(inner, struct_name, quote!(item));
+
+            quote! {
+                if let Some(item) = &#field_expr {
```
PatrykWalach: Option can use `.iter()` so it doesn't need this code

crates/resolve_position_macros/src/resolve_position_macro.rs
```diff
@@ -56,31 +56,18 @@ fn handle_data_struct(
...
-                })
-                .collect::<Vec<_>>()
-        }
+        Ok(field_infos) => field_infos
+            .into_iter()
+            .flatten()
+            .map(
+                |ResolveFieldInfo {
+                     field_name,
+                     field_type,
```
PatrykWalach, line 65: how does it work without the `inner_type`?

PatrykWalach: ok, field_type has the inner_type

crates/resolve_position_macros/src/resolve_position_macro.rs
```diff
@@ -166,10 +153,73 @@ struct ResolvePositionArgs {
...
+        // Container types: Vec<T> or Option<T>
+        if last_segment.ident == "Vec" || last_segment.ident == "Option" {
+            if let Some(syn::Type::Path(syn::TypePath {
+                path: inner_path, ..
+            })) = extract_single_generic_type(last_segment)
+            {
+                // Recursively parse the inner type
+                let inner_wrapper = parse_resolve_field_type(inner_path, generics_map)?;
+
+                return Ok(ResolveFieldInfoTypeWrapper::Vec(Box::new(inner_wrapper)));
```
rbalicki2: `::Vec` is poorly named, because this also matches `Option`. Perhaps `IteratorWrapper`?

rbalicki2 (approved):
LGTM just renamed the variant

rbalicki2:
Thanks for doing this. Love that we're improving this code!!! Looks great

PR #601, opened 2025-08-27
fix: LSP on Windows

PatrykWalach:
As it turns out, some functions return paths in lowercase: "e:\isograph\..." (beacuse paths on windows are not case sensitive). Calling `canonicalize` corrects the casing.


rbalicki2 (approved):
lgtm, will fix a nit and shipit

PR #603, opened 2025-08-28
Support lazy loading entrypoint readers

### LINKED ISSUE #579: Make readerArtifact a lazy import for entrypoints
- [ ] Add fields to `LazyLoadDirectiveSet`, `reader: bool, normalization: bool`, if missing default to true
- [ ] If `reader: true` -> replace hard require with `() => import(...)` in entrypoint.ts file, and `readerArtifact: { kind: 'ReaderLoader', loader: () => import('./resolver_reader').then(x => x.default) }`
- [ ] Add support to `makeNetworkRequest` for fetching the reader ast, similar to how normalization ast's are loaded there. We also need to suspend on the missing reader AST when reading the entrypoint: in `getOrCreateCacheForArtifact`, we access the `readerArtifact`.  TBD

XiNiHa:
Closes #579

### DIFF
--- a/crates/generate_artifacts/src/entrypoint_artifact.rs
+++ b/crates/generate_artifacts/src/entrypoint_artifact.rs
@@ -423,12 +423,43 @@ fn entrypoint_file_content<TNetworkProtocol: NetworkProtocol>(
         }
     };
 
+    let (reader_import, reader_code) = {
+        let reader_resolver_file_path =
+            format!("'./{resolver_reader_file_name}{ts_file_extension}'");
+        match directive_set {
+            EntrypointDirectiveSet::LazyLoad(directive_set) if directive_set.lazy_load.reader => (
+                "".to_string(),
+                format!(
+                    "{indent}readerWithRefetchQueries: {{\n\
+                     {indent}  kind: \"ReaderWithRefetchQueriesLoader\",\n\
+                     {indent}  loader: () => import({reader_resolver_file_path})\n\
+                     {indent}    .then(module => ({{\n\
+                     {indent}      kind: \"ReaderWithRefetchQueries\",\n\
+                     {indent}      nestedRefetchQueries,\n\
+                     {indent}      readerArtifact: module.default,\n\
+                     {indent}    }}))\n\
+                     {indent}}}"
+                ),
+            ),
+            _ => (
+                format!("import readerResolver from {reader_resolver_file_path};\n"),
+                format!(
+                    "{indent}readerWithRefetchQueries: {{\n\
+                     {indent}  kind: \"ReaderWithRefetchQueries\",\n\
+                     {indent}  nestedRefetchQueries,\n\
+                     {indent}  readerArtifact: readerResolver,\n\
+                     {indent}}},"
+                ),
+            ),
+        }
+    };
+
     format!(
         "import type {{IsographEntrypoint, \
         {normalization_ast_type_name}, RefetchQueryNormalizationArtifactWrapper}} from '@isograph/react';\n\
         import {{{entrypoint_params_typename}}} from './{param_type_file_name}{ts_file_extension}';\n\
         import {{{entrypoint_output_type_name}}} from './{output_type_file_name}{ts_file_extension}';\n\
-        import readerResolver from './{resolver_reader_file_name}{ts_file_extension}';\n\
+        {reader_import}\
         import queryText from './{query_text_file_name}{ts_file_extension}';\n\
         {normalization_ast_import}\
         {refetch_query_artifact_import}\n\n\
@@ -444,11 +475,7 @@ fn entrypoint_file_content<TNetworkProtocol: NetworkProtocol>(
         {normalization_ast_code}\n\
         {indent}}},\n\
         {indent}concreteType: \"{concrete_type}\",\n\
-        {indent}readerWithRefetchQueries: {{\n\
-        {indent}  kind: \"ReaderWithRefetchQueries\",\n\
-        {indent}  nestedRefetchQueries,\n\
-        {indent}  readerArtifact: readerResolver,\n\
-        {indent}}},\n\
+        {reader_code}\n\
         }};\n\n\
         export default artifact;\n",
     )
--- a/crates/generate_artifacts/src/entrypoint_artifact.rs
+++ b/crates/generate_artifacts/src/entrypoint_artifact.rs
@@ -6,8 +6,8 @@ use common_lang_types::{
 };
 use isograph_config::GenerateFileExtensionsOption;
 use isograph_lang_types::{
-    DefinitionLocation, EmptyDirectiveSet, EntrypointDirectiveSet, LazyLoadDirectiveParameters,
-    LazyLoadDirectiveSet, ScalarSelectionDirectiveSet, SelectionType,
+    DefinitionLocation, EmptyDirectiveSet, EntrypointDirectiveSet, ScalarSelectionDirectiveSet,
+    SelectionType,
 };
 use isograph_schema::{
     create_merged_selection_map_for_field_and_insert_into_global_map,
@@ -405,7 +405,7 @@ fn entrypoint_file_content<TNetworkProtocol: NetworkProtocol>(
     let (normalization_ast_type_name, normalization_ast_import, normalization_ast_code) = {
         let file_path = format!("'./{normalization_text_file_name}{ts_file_extension}'");
         match directive_set {
-            EntrypointDirectiveSet::LazyLoad(LazyLoadDirectiveSet { lazy_load: LazyLoadDirectiveParameters {} }) => (
+            EntrypointDirectiveSet::LazyLoad(directive_set) if directive_set.lazy_load.normalization => (
                 "NormalizationAstLoader",
                 "".to_string(),
                 format!(
@@ -415,7 +415,7 @@ fn entrypoint_file_content<TNetworkProtocol: NetworkProtocol>(
                      {indent}  }},"
                 ),
             ),
-            EntrypointDirectiveSet::None(_) => (
+            _ => (
                 "NormalizationAst",
                 format!("import normalizationAst from {file_path};\n"),
                 format!("{indent}  normalizationAst,"),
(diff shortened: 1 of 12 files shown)


crates/isograph_lang_types/src/entrypoint_directive_set.rs
```diff
@@ -15,6 +15,44 @@ pub struct LazyLoadDirectiveSet {
...
+
+impl From<LazyLoadDirectiveParametersInput> for LazyLoadDirectiveParameters {
+    fn from(value: LazyLoadDirectiveParametersInput) -> Self {
+        match value {
+            LazyLoadDirectiveParametersInput {
+                reader: None,
+                normalization: None,
+            } => Self::default(),
+            _ => Self {
+                reader: value.reader.unwrap_or(false),
```
rbalicki2: can you leave a comment about what the behavior is in various circumstances? What defaults to true, what to false, etc?

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -60,13 +81,13 @@ export function readPromise<T, E>(p: PromiseWrapper<T, E>): T {
     }
   }
 
-  throw p.promise;
+  throw p.promise();
```
rbalicki2: I will investigate more, but I suspect that this may be problematic in terms of suspense, because now the promise is not guaranteed to be stable (though you did say `idempotent` above.) If anything, I would feel more comfortable with something like

```
{ promise: UNSET_SYMBOL | Promise, readonly promiseLoader: () => Promise, result: Result<...> }
```

That way we can enforce that the promise loader is called only once.

rbalicki2 (commented):
partial review, just a few nits so far

crates/generate_artifacts/src/entrypoint_artifact.rs
```diff
@@ -414,20 +414,53 @@ fn entrypoint_file_content<TNetworkProtocol: NetworkProtocol>(
...
                 format!("{indent}  normalizationAst,"),
             ),
         }
     };
 
+    let (reader_isograph_import, reader_import, reader_code) = {
+        let reader_resolver_file_path =
+            format!("'./{resolver_reader_file_name}{ts_file_extension}'");
+        match directive_set {
+            EntrypointDirectiveSet::LazyLoad(directive_set) if directive_set.lazy_load.reader => (
```
rbalicki2, line 430: Amazing, love the changes here

libs/isograph-react/src/core/PromiseWrapper.ts
XiNiHa: The `wrapPromise` and `wrapLoader` ensures the promise to be cached and therefore stable. If we want to be extra safe we could add a fake type-level symbol so the PromiseWrappers can be only constructed using the `wrap` helpers

PatrykWalach: `wrapResolvedValue` is not stable

PatrykWalach: but I guess it doesn't matter, because you have value already

libs/isograph-react/src/react/useImperativeReference.ts
```diff
@@ -70,12 +69,7 @@ export function useImperativeReference<
       setState([
         {
           kind: 'FragmentReference',
-          readerWithRefetchQueries: wrapResolvedValue({
-            kind: 'ReaderWithRefetchQueries',
-            readerArtifact: entrypoint.readerWithRefetchQueries.readerArtifact,
-            nestedRefetchQueries:
-              entrypoint.readerWithRefetchQueries.nestedRefetchQueries,
-          }),
+          readerWithRefetchQueries: entrypoint.readerWithRefetchQueries,
```
PatrykWalach: PromiseWrapper was supposed to be a promise you can throw or read in react component, it was clear it's already pending - the network request happened somewhere above. Now the request can happen during render if you forget to call `.promise()` leading to waterfalls.

Although `maybeMakeNetworkRequest` is called above and you are settings `readerWithRefetchQueries` to `PromiseWrapper` there's a chance the Reader is not feching. The reason is: if `maybeMakeNetworkRequest` has required data in cache yow will not call `.promise()`. You will fetch the reader during render which can be a lot slower.


@rbalicki2 I think the changes to `PromiseWrapper` should be reversed, and `maybeMakeNetworkRequest` should also return `PromiseWrapper` of Reader

XiNiHa: Makes sense, let me do it that way!

XiNiHa: It feels like loading bundles for readers is a separate process from making network requests, considering that readers are always needed even when network requests are not fired. So now it instead starts loading outside of `maybeMakeNetworkRequest` and injects the promise into it

PatrykWalach: This looks very good now!

One extra thought regarding the new argument you added:

Generated `__refetch` fields (for example: `demos\github-demo\src\isograph-components\__isograph\User\__refetch\refetch_reader.ts`) import `makeNetworkRequest`, but this new argument isn’t being passed there.

I’m not sure if that’s actually a problem, because right now `readerWithRefetchQueries` in `makeNetworkRequest` is only passed to:
- `normalizeData` – I checked, and this argument isn’t used. I’ll open another issue to remove it.
- `onComplete` – ~~I’m not sure if `onComplete` can be triggered for `__refetch`, it probably can~~ it can be triggered for entrypoints only.

XiNiHa: Isn't it a `RefetchQueryNormalizationArtifact`? It didn't contain `readerWithRefetchQueries` anyway so it just calls the function with the optional parameter omitted

PatrykWalach: Yeah, you don't need to do anything there

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -130,8 +150,9 @@ export function makeNetworkRequest<
       variables,
     ),
     loadNormalizationAst(artifact.networkRequestInfo.normalizationAst),
+    readerWithRefetchQueries?.promise,
```
PatrykWalach, line 153: `Promise.all` between network response and normalization ast makes sense, because we need both to normalize to store, now if we fail to fetch reader, we won't normalize the response @rbalicki2 is that correct thing to do? Here reader is used only for onComplete so should we use it just for this? 
```tsx
readerWithRefetchQueries?.promise.then(() =>{
  readDataForOnComplete(...);
})
```

XiNiHa: We can also `.catch(() => {})` if the error needs to be ignored

demos/pet-demo/src/components/FavoritePhrase.tsx
```diff
@@ -1,51 +1,7 @@
...
-        <ErrorBoundary>
-          <React.Suspense fallback="Loading favorite phrase...">
-            <FragmentReader fragmentReference={fragmentReference} />
-          </React.Suspense>
-        </ErrorBoundary>
-      )}
-    </>
-  );
-});
+console.log('Query.PetFavoritePhrase bundle being loaded!');
```
rbalicki2, line 4: this is probably accidentally left over from testing

libs/isograph-react/src/core/read.ts
```diff
@@ -872,6 +877,7 @@ export function readLinkedFieldData(
                 environment,
                 refetchQueryArtifact,
                 variables,
+                undefined,
```
rbalicki2, line 880: Since we don't pass a PromiseWrapper here (correctly, since the reader AST is part of the parent reader AST), I think this might be broken if the pointer is loaded with an `onComplete`. If we have an `onComplete`, we try to call `readDataForOnComplete` with an undefined `readerWithRefetchQueries`, which should throw on e.g. `const readerArtifact = resolvedReaderWithRefetchQueries.readerArtifact;`.

Not a blocker, because we don't have pointers shipped yet, but @PatrykWalach just something to be aware of.

That being said, I think it might be an easy fix to make the `readerWithRefetchQueries` parameter to `wrapResolvedValue` not optional, and to pass the `wrapResolvedValue(...)` call below.

libs/isograph-react/src/core/makeNetworkRequest.ts
rbalicki2: I am really torn about this. I don't know what the correct behavior is.
- I think that if the user passes `onComplete`, we should certainly await the reader before normalizing. Those should occur at the same time. And if the reader fails, we should also fail to normalize. As in, we should both normalize and call onComplete, or not at all.
- I don't think folks will always use the parameter passed to `onComplete` (and in particular, it improves perf to not read that data), so it's conceivable that we could also support an `onComplete` that doesn't read that data. But this seems fairly niche.
- I think if both the reader and the normalization AST are lazy loaded, we could probably have a single lazy import `() => import('./normalization_and_reader')` and use that. That probably will improve perf as well, and cut down on the number of distinct states we have to handle.
- But for now, I'm happy to leave it as is!

demos/pet-demo/src/components/FavoritePhrase.tsx
XiNiHa: I thought it'd be good to have this here for the demonstration purpose since it's a demo 😂 (it's a bit hard to see the resolver bundle being lazyloaded only with the network devtools since Next bundles the chunks)

libs/isograph-react/src/core/makeNetworkRequest.ts
XiNiHa: What's the primary usage of `onComplete` though? It feels like a field resolution is separate from making a network request, so IMO `onComplete` shouldn't be called with a resolver data if it's about the network request, and `onComplete` shouldn't be called here if it's about the entrypoint resolver. Personally it feels unnatural to couple the network request with resolver read. What if the resolver errors? Shouldn't the cache be reverted if we see those as coupled? Maybe it shows that it doesn't make sense to couple those two? 😅

libs/isograph-react/src/core/read.ts
XiNiHa: At least the types are fine for now. I believe the types will break as the feature comes in, so we can fix it at that point 😋

PR #606, opened 2025-08-28
feat(macros): add WithEmbeddedLocation variant and handling

PatrykWalach:
I've made ClientPointerDeclaration use WithLocation, but client declarations should use WithEmbeddedLocation since, they always have a location. This will remove some assertions. Selectables will remain to use WithLocation.


rbalicki2:
LGTM, lets suppress the clippy lint and land!

PR #608, opened 2025-08-28
wrap client selectable names WithLocation wrapper

PatrykWalach:


crates/isograph_schema/src/create_additional_fields/add_link_fields.rs
```diff
@@ -21,7 +21,7 @@ impl<TNetworkProtocol: NetworkProtocol> Schema<TNetworkProtocol> {
                             .intern()
                             .into(),
                     )),
-                    name: field_name,
+                    name: WithLocation::new(field_name, Location::generated()),
```
PatrykWalach, line 24: Link is Generated

crates/isograph_schema/src/create_additional_fields/expose_field_directive.rs
```diff
@@ -221,7 +221,10 @@ impl<TNetworkProtocol: NetworkProtocol> Schema<TNetworkProtocol> {
 
         let mutation_client_scalar_selectable = ClientScalarSelectable {
             description,
-            name: client_field_scalar_selection_name.unchecked_conversion(),
+            name: WithLocation::new(
+                client_field_scalar_selection_name.unchecked_conversion(),
+                Location::generated(),
```
PatrykWalach, line 226: That's `ImperativelyLoadedField` so it's either `__refetch` or `@exposeField`, so for exposed field we can probably do better, don't care for now

PatrykWalach:
Also all `name.item` can be replaced with `name()`

crates/isograph_schema/src/create_additional_fields/expose_field_directive.rs
rbalicki2: sgtm, as long as goto def/hover doesn't cause the lsp to panic, sgtm. But we basically skip over generated locations, so that's fine

```
             if let Some(span) = #field_expr.location.span() {
                    if span.contains(position) {
                        let new_parent = <#inner_type as ::resolve_position::ResolvePosition>::Parent::#struct_name(self.path(parent).into());
                        return #field_expr.item.resolve(new_parent, position);
                    }
                }
```

rbalicki2 (approved):
LGTM shipit

PR #615, opened 2025-08-29
wrap server entity name in `WithEmbeddedLocation`

PatrykWalach:
Need the location for goto definition of field/pointers parent types

#556


crates/isograph_lang_types/src/isograph_resolved_node.rs
```diff
@@ -46,14 +46,14 @@ fn get_path_using_selection_parent<'a>(
             string_vec.push(format!(
                 "{}.{}",
                 client_field_declaration.inner.parent_type,
-                client_field_declaration.inner.client_field_name
+                client_field_declaration.inner.client_field_name.item
```
PatrykWalach, line 49: This made me realize when wrapping field in `WithLocation` fixing all compiler errors is not enough, because `WithLocation` implements `Display`, so if there's `print!("{}", obj.field)` this won't get caught 😫

PatrykWalach: And this is a leftover from when I was wrapping in `WithLocation` last time

PatrykWalach: This time I've commented out the `Display` trait to get all the errors

rbalicki2: Nice. I have been too lazy to do this, but we should probably do something like https://github.com/isographlabs/isograph/issues/616

PR #617, opened 2025-08-30
goto parent type definition

PatrykWalach:


crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -76,7 +76,24 @@ pub fn on_goto_definition_impl<TNetworkProtocol: NetworkProtocol + 'static>(
...
+                let server_object_entity = match validated_schema
+                    .server_entity_data
+                    .server_object_entity(entity.inner.0.unchecked_conversion())
+                {
+                    Some(server_object_entity) => server_object_entity,
+                    None => {
+                        return Ok(None);
+                    }
+                };
+
```
PatrykWalach, line 89: This is working for fields and entrypoint with directive, if entrypoint doesn't have a directive, it doesn't work 🤔

rbalicki2: interesting, seems to work for me on entrypoints and fields without directives

rbalicki2 (approved):
lgtm shipti

PR #618, opened 2025-08-30
resolve GraphQLAnnotation position

PatrykWalach:
GraphQLAnnotation has a built in span
#556


crates/isograph_lang_parser/src/parse_iso_literal.rs
```diff
@@ -309,7 +309,7 @@ fn parse_client_pointer_target_type(
         ));
     }
 
-    parse_type_annotation(tokens)
+    parse_type_annotation(tokens).map(|with_span| with_span.map(ServerObjectEntityNameWrapper))
```
PatrykWalach, line 312: In the future need to update `parse_type_annotation` to return `ServerObjectEntityNameWrapper`, this will make variable definitions clickable

rbalicki2:
looks like fixtures need to be updated

PR #619, opened 2025-08-30
fix goto def on mac

rbalicki2:


crates/isograph_lsp/src/goto_definition.rs
```diff
@@ -182,17 +182,14 @@ fn isograph_location_to_lsp_location<TNetworkProtocol: NetworkProtocol + 'static
     let path_buf = PathBuf::from(db.get_current_working_directory().lookup())
         .join(location.text_source.relative_path_to_source_file.lookup());
 
-    let path = path_buf.to_str()?;
```
PatrykWalach, line 185: This broke windows

I meant something like this 

```suggestion
    let path = path_buf.to_str()?;
    let normalized_path = if cfg!(windows) {
        Cow::Owned(format!(
            "/{}",
            path.strip_prefix(r"\\?\")
                .unwrap_or(path)
                .replace("\\", "/")
        ))
    } else {
        Cow::Borrowed(path)
    };
    let uri = Uri::from_str(&format!("file://{normalized_path}")).ok()?;
```

rbalicki2: okay updated, landing. (It is exactly what you suggested)

Issue #298, opened 2024-12-28
Link should be generic e.g. Link<Viewer>


rbalicki2:

PatrykWalach:
Hm...
<img width="885" height="860" alt="Image" src="https://github.com/user-attachments/assets/6f54f93b-e691-4e9b-8970-16316e28c51c" />

rbalicki2:
trololol

PatrykWalach:
After your explanation of how Relay handles this, I think I prefer that approach.  
It’s more restrictive:  

- A pointer must target a **concrete type**, **or**  
- A pointer may target an **abstract type**, but in that case:  
  - Only a `Link` of the same abstract type can be returned.  
  - Its concrete subtypes are not allowed, since we don’t have the necessary mappings for them.  

But the types end up a lot simpler

rbalicki2:
I'm not sure I follow. Are you saying that `pointer Query.Foo to Node { currentUser { link } }` will not work, even though `currentUser: User` and `User implements Node`? I think that TypeScript will allow that (as in, you can return a `Link<"User">` when `NodeLink` is required (because it's just a union). And that seems like a reasonable thing to wont.

rbalicki2:
I'm not sure I follow. Are you saying that `pointer Query.Foo to Node { currentUser { link } }` will not work, even though `currentUser: User` and `User implements Node`? I think that TypeScript will allow that (as in, you can return a `Link<"User">` when `NodeLink` is required (because it's just a union). And that seems like a reasonable thing to wont.

Issue #206, opened 2024-10-24
Client pointers: make ID types generic


rbalicki2:

PatrykWalach:
I like using abstract classes for this as they have clearer error message than when using branded types, but branded types allow for using the value as string directly [playground for both](https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABFApgZygYQDYEM1oAUAlIgN4BQiiuARhgE67SIR4GICSAJgDwAqAPnJVqiAA4M4qaCm6I443AEcQKAIQAuRPwDco6nUbMoyOAGUoDGGADmJbYxu391AL4VP1bClMg0KAw8iAC8iADkAIwATOE0aIjgANZgcADuSPhcfABEAKoBDDmCogD0pQACUGgAtCgAHuIo0HUMUgzIABYwCT2Iqab4aDC2YHQ+ZohOdqI+pj2W1naOVs6hiYU8s76ItNhwtjzaPLw5AEL7tsXr-oFbHp4UoJCwCMjoUPwAnk0kItQQBAYBRKVQodbmL4AW1ocGwhByihUahyxFEUB+4JOQnW01siAAZP8xABtJFggC62j0oge2z8m3kYSisXiiTAKXSmQSJ3yhWK9MQC1WyymIvxYVuQW4gr2ByO2VOFwO10ljIoHiAA)

PatrykWalach:
The implementation we use at work is pretty interesting https://github.com/microsoft/TypeScript/issues/202#issuecomment-436900738 it doesn't introduce any extra properties on the object, branded type is still assignable to original type.

rbalicki2:
ooh this is really nice. Thanks for sharing.

I'm okay with `Brand<X, 'User'>` being equal to another `Brand<X, 'User'>`, because that's exactly what we want: a user ID is a valid user ID. The other situation: a user ID must come from the input can be solved with generics, i.e. the input is of type `SelectionSet<TFoo>` and we expect a value of `Brand<TFoo>`. But maybe there's no downside to using symbols, which seems cleaner.

PatrykWalach:
I was testing this now, it's not possible to do
```tsx
type Store = {
  [index: TypeName]: ...
}
```
it's possible to use a record
```tsx
type Store = Record<TypeName, ...>
```
but then it's not working fully
```tsx
const store = {} as Store; // have to cast

for(const typeName in store){
  typeName
  // ^ this is string
}
```

[...]

PR #620, opened 2025-08-30
add `BrandedLink`

PatrykWalach:
This also improves typesafety of updaters

#206, #298

### DIFF
--- a/demos/pet-demo/src/components/__isograph/Query/smartestPet/output_type.ts
+++ b/demos/pet-demo/src/components/__isograph/Query/smartestPet/output_type.ts
@@ -1,4 +1,4 @@
 import type { Link } from '@isograph/react';
 import type React from 'react';
 import { SmartestPet as resolver } from '../../../SmartestPet';
-export type Query__smartestPet__output_type = (Link | null);
+export type Query__smartestPet__output_type = (Link<"Pet"> | null);
--- a/demos/pet-demo/src/components/__isograph/Query/smartestPet/param_type.ts
+++ b/demos/pet-demo/src/components/__isograph/Query/smartestPet/param_type.ts
@@ -6,7 +6,7 @@ export type Query__smartestPet__param = {
       /**
 A store Link for the Pet type.
       */
-      readonly link: Link,
+      readonly link: Link<"Pet">,
       readonly stats: ({
         readonly intelligence: (number | null),
       } | null),


crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -162,6 +162,32 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+            );
+
+        if let Some(concrete_type) = server_object_entity.concrete_type {
+            return format!("Link<\"{concrete_type}\">");
+        }
+
+        let subtypes = server_object_entity
+            .network_protocol_associated_data
+            .subtypes
+            .iter()
```
rbalicki2, line 184: YESSSS

crates/graphql_network_protocol/src/process_type_system_definition.rs
```diff
@@ -220,6 +224,21 @@ pub fn process_graphql_type_system_document(
         }
     }
 
+    // add subtypes to associated_data
+    for (supertype_name, subtypes) in supertype_to_subtype_map.iter() {
+        if let Some((object_outcome, _)) = objects.iter_mut().find(|obj| {
```
rbalicki2: should we `unwrap` here instead of silently ignoring the supertype is not found?

crates/generate_artifacts/src/refetch_reader_artifact.rs
```diff
@@ -134,7 +134,7 @@ fn generate_function_import_statement(read_out_data: String) -> ClientFieldFunct
         {indent}artifact: RefetchQueryNormalizationArtifact,\n\
         {indent}readOutData: any,\n\
         {indent}filteredVariables: any,\n\
-        {indent}rootLink: Link,\n\
+        {indent}rootLink: Link<any>,\n\
```
rbalicki2, line 138: can you add a TODO here? We should know the type, right? `Pet.__refetch` should have a `Link<"Pet">`, right?

Anyway, not worth actually doing (because 🤞  we get rid of refetch) but let's track it

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -88,6 +88,14 @@ export type IsographNetworkFunction = (
   variables: Variables,
 ) => Promise<any>;
 
+// @ts-ignore
+type Brand<A, B extends symbol | string> = infer _ extends B ? A : never;
+
+export type BrandedLink<T extends TypeName> = {
```
rbalicki2: should we name these `Link` and `StoreLink` instead of renaming them in `src/index`?

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -88,6 +88,14 @@ export type IsographNetworkFunction = (
   variables: Variables,
 ) => Promise<any>;
 
+// @ts-ignore
+type Brand<A, B extends symbol | string> = infer _ extends B ? A : never;
```
rbalicki2: can you leave a comment explaining what's going on? Claude is confused:

-----

This TypeScript code is broken. The `infer` keyword can only be used within the `extends` clause of a conditional type, but you're trying to use it as a standalone type alias.

What you probably want is one of these patterns:

**Option 1: Standard brand type**
```typescript
type Brand<A, B extends symbol | string> = A & { readonly __brand: B };
```

**Option 2: Conditional brand (though unclear why)**
```typescript
type Brand<A, B> = B extends symbol | string ? A & { readonly __brand: B } : never;
```

The brand pattern is typically used for nominal typing - creating distinct types that are structurally identical but semantically different, like `UserId` vs `ProductId` where both are numbers but shouldn't be interchangeable.

Your syntax suggests you might be thinking of extracting types from conditional types, but that requires the `infer` to be properly positioned within a conditional type's extends clause, not floating as a type parameter.

libs/isograph-react/src/core/IsographEnvironment.ts
PatrykWalach: yes, renamed

libs/isograph-react/src/core/brand.ts
```diff
@@ -0,0 +1,15 @@
...
+// enabling A to be treated as distinct only when tagged with B, even though B doesn't exist at runtime.
+//
+// Explanation:
+// - B extends `symbol | string` acts as a “brand” identifier.
+// - The type uses a conditional check `infer _ extends B ? A : never` to strip out A when the branding doesn't match.
+// - This yields a branded type system: `Brand<string, "UserId">` is not accidentally assignable to `Brand<string, "ProductId">`.
+//
+// Usage: Helps enforce semantic distinctions (e.g., distinguishing user IDs from product IDs) even when their runtime values are both strings.
+//
+// Caveat: This is purely a compile-time trick—B is erased in emitted JavaScript, so runtime checks must rely on other mechanisms.
```
rbalicki2: nice. Love this explanation. Can we rename `A` to be `BaseType` and `B` to be `Brand` for clarity?

crates/graphql_network_protocol/src/process_type_system_definition.rs
PatrykWalach: yes, should exist by now

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -162,6 +162,32 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+            .map(|name| format!("Link<\"{name}\">"))
```
rbalicki2, line 185: One thing that Relay avoided was ever having runtime or type info along the lines of "these 1000 types implement Node", which is what this is. I think Relay is wrong, and it's worth having, at least on the type level.

So, not a blocker, just commenting.

At some point, adding a new type that implements Node will result in changes to 5000 link types, so we should at least eventually change this to `import { Node } from '__isograph/Node/type.ts'` and `Link<Node>` or something.

rbalicki2:
```
      /**
A store Link for the Node type.
      */
      readonly link: Link<"Viewer"> | Link<"Pet"> | Link<"BlogItem"> | Link<"Image"> | Link<"AdItem">,
```

rbalicki2: Also not a blocker, but the last thing I wonder about is whether this is correct given that new types can be added to the schema. So maybe we need a case ` | FutureLinkType`, because you can write an exhaustive match that breaks when a new type is added to the backend.

But we can add that as a follow up

rbalicki2: And we should support a directive `@closed` on interfaces and unions, which avoids adding this. Because some unions and interfaces are known to be closed!

PatrykWalach: > One thing that Relay avoided was ever having runtime or type info along the lines of "these 1000 types implement Node"

I'm sure if I'd selected `__typename` on `Query.node` then relay would generate the type, right?

rbalicki2: No, relay would generate `__typename: string`

rbalicki2: And if we have `__typename ... on User { name )` relay would generate `{ __typename: "User", name: string } | { __typename: "%future type" }`

PR #625, opened 2025-08-30
[LSP] fix overlow error on hover

### LINKED ISSUE #624: Goto def doesn't work for src/components/AllInvoicesRoute.tsx in https://github.com/Benanna2019/Isograph-Tanstack-Invoices/tree/main

PatrykWalach:
This line was cauing `attempt to subtract with overflow` error

This maybe fixes #624


rbalicki2:
BAHAHA I think on that line I fixed half a bug (i.e. fixed the LHS to be `char`) but not the RHS

I literally cannot code without newtypes 😭 . Let's wrap char and line with newtypes

PR #637, opened 2025-09-01
add opentelemetry

ch1ffa:
This PR adds ability to publish traces and spans to an open telemetry compatible collector (e.g. [jaeger](https://www.jaegertracing.io)).

To activate telemetry you should add the following to the isograph config file options section:
```json
"opentelemetry": {
    "enable_tracing": true,
    "collector_endpoint": "http://localhost:4317",
    "service_name": "pet-demo"
}
```

The simplest way to test it is to run `jaeger` in docker:
```bash
docker run --rm --name jaeger \
  -p 16686:16686 \
  -p 4317:4317 \
  -p 4318:4318 \
  -p 5778:5778 \
  -p 9411:9411 \
  cr.jaegertracing.io/jaegertracing/jaeger:2.9.0
```

its default grpc entrypoint is `http://localhost:4317`, UI is accessible via `http://localhost:16686`. There you can select provided service name and operation name (e.g. `compile`) to see available traces which have been sent during compilation.

Screenshot 2025-08-31 at 15 18 21

### DIFF
--- a/crates/isograph_config/src/compilation_options.rs
+++ b/crates/isograph_config/src/compilation_options.rs
@@ -40,6 +40,7 @@ pub struct CompilerConfigOptions {
     pub module: JavascriptModule,
     pub generated_file_header: Option<GeneratedFileHeader>,
     pub persisted_documents: Option<PersistedDocumentsOptions>,
+    pub opentelemetry: Option<OpenTelemetryOptions>,
 }
 
 #[derive(Default, Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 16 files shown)


crates/isograph_config/src/compilation_options.rs
```diff
@@ -356,6 +378,18 @@ fn create_persisted_documents(
     })
 }
 
+fn create_opentelemetry(
```
rbalicki2, line 381: nit: `open_telemetry` here and in other places

crates/isograph_config/src/compilation_options.rs
```diff
@@ -105,6 +106,13 @@ pub struct PersistedDocumentsOptions {
     pub include_extra_info: bool,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OpenTelemetryOptions {
+    pub enable_tracing: bool,
```
rbalicki2, line 111: I think I am fine with this, because folks may want to always include the collector endpoint and service name, while disabling/enabling tracing for some reason. But, to be clear, this effectively is equivalent to not having the `opentelemetry` key, right?

ch1ffa: Yeah, just for convenience to keep config values here if someone wants to disable it temporarily

rbalicki2: sgtm, it makes sense for a big co to have this

Issue #633, opened 2025-08-31
iso.ts has wrong types for pointers


rbalicki2:
- We do not enforce the return type of pointers in iso.ts. e.g.

```
type IdentityWithParam<TParam extends object> = <TClientFieldReturn>(
  clientField: (param: TParam) => TClientFieldReturn
) => (param: TParam) => TClientFieldReturn;

export function iso<T>(
  param: T & MatchesWhitespaceAndString<'pointer Query.smartestPet', T>
): IdentityWithParam<Query__smartestPet__param>;
```

This essentially allows the pointer to have any return type (which is correct for a client field).

PR #645, opened 2025-09-04
Restrict return types for pointers

PatrykWalach:
Follow up to #620

#633

### DIFF
--- a/crates/generate_artifacts/src/iso_overload_file.rs
+++ b/crates/generate_artifacts/src/iso_overload_file.rs
@@ -9,7 +9,7 @@ use isograph_schema::{
     EntrypointDeclarationInfo, NetworkProtocol, Schema,
 };
 
-use crate::generate_artifacts::ISO_TS_FILE_NAME;
+use crate::generate_artifacts::{print_javascript_type_declaration, ISO_TS_FILE_NAME};
 
 fn build_iso_overload_for_entrypoint<TNetworkProtocol: NetworkProtocol>(
     validated_client_field: &ClientScalarSelectable<TNetworkProtocol>,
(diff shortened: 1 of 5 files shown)


PatrykWalach:
I should also restrict components to return ReactNode in this PR or follow up

EDIT: or maybe we are doing this already, have to check

rbalicki2 (approved):
lgtm shipit

crates/generate_artifacts/src/iso_overload_file.rs
```diff
@@ -89,14 +106,15 @@ pub(crate) fn build_iso_overload_artifact<TNetworkProtocol: NetworkProtocol>(
     file_extensions: GenerateFileExtensionsOption,
     no_babel_transform: bool,
 ) -> ArtifactPathAndContent {
-    let mut imports = "import type { IsographEntrypoint } from '@isograph/react';\n".to_string();
+    let mut imports =
+        "import type { IsographEntrypoint, Link } from '@isograph/react';\n".to_string();
```
rbalicki2: let's add `ts-ignore` here, since the `Link` field can be unused and cause TS to complain

rbalicki2: haven't pushed yet (on a flight, can't push, will push when I land), but this is fixed on master: we now generate tsconfig files that allow unused locals and parameters.

rbalicki2:
rebased, should probably pass CI now

crates/generate_artifacts/src/iso_overload_file.rs
```diff
@@ -46,13 +46,24 @@ fn build_iso_overload_for_client_defined_type<TNetworkProtocol: NetworkProtocol>
...
         "import {{ type {}__param }} from './{}/{}/param_type{}';\n",
         client_type.type_and_field().underscore_separated(),
         client_type.type_and_field().type_name,
         client_type.type_and_field().field_name,
         file_extensions.ts()
     );
+
+    if let SelectionType::Object(client_pointer) = client_type {
+        import = format!(
+            "{}import {{ type {}__link__output_type }} from './{}/link/output_type{}';\n",
```
PatrykWalach: There's an issue, this can be imported more than once

rbalicki2: will we import the same link multiple times if there are multiple pointers to the same type?

rbalicki2: If so, might be best to just have a `BTreeSet` that we populate

PR #646, opened 2025-09-05
use isograph database fields for source files

ch1ffa:


crates/pico/src/execute_memoized_function.rs
```diff
@@ -193,10 +193,7 @@ fn any_dependency_changed<Db: Database>(db: &Db, derived_node: &DerivedNode<Db>)
 fn source_node_changed_since<Db: Database>(db: &Db, key: Key, since: Epoch) -> bool {
     match db.get_storage().internal.get_source_node(key) {
         Some(source) => source.time_updated > since,
-        None => panic!(
-            "Source node not found. This may occur if \
-            a `SourceId` is used after the source node has been removed."
-        ),
+        None => true,
```
ch1ffa, line 196: if we remove source, registered parent derived_node still references it and we need to invalidate it

crates/pico_macros/src/db_macro.rs
```diff
@@ -1,52 +1,123 @@
...
+            continue;
+        }
+
+        if is_phantom_data(&f.ty) {
+            continue;
+        }
+
+        let field_ty = &f.ty;
+        let counter_ident =
+            format_ident!("__{}Counter", field_ident.to_string().to_case(Case::Pascal));
```
rbalicki2: til about to_case, that's neat

crates/pico_macros/src/db_macro.rs
```diff
@@ -1,52 +1,123 @@
...
+    let mut aux_struct_defs = Vec::new();
+
+    for f in fields.iter() {
+        let field_ident = f.ident.as_ref().unwrap();
+
+        if field_ident == "storage" {
+            continue;
+        }
+
+        if is_phantom_data(&f.ty) {
```
rbalicki2: makes sense, and I don't think what I'm about to say is a blocker. But should this be configurable and explicit? I would imagine that you might want to not expose some private fields.

crates/isograph_schema/src/isograph_database.rs
```diff
@@ -67,22 +73,82 @@ impl<TNetworkProtocol: NetworkProtocol + 'static> IsographDatabase<TNetworkProto
...
+        file: RelativePathToSourceFile,
+    ) -> Option<SourceId<OpenFileSource>> {
+        self.open_file_map.0.get(&file).cloned()
+    }
+
+    pub fn remove_schema_extension(
+        &mut self,
+        relative_path: RelativePathToSourceFile,
+    ) -> Option<SourceId<SchemaSource>> {
+        self.get_standard_sources_tracked_mut()
```
rbalicki2: `self.standard_sources` is still pub, can we avoid that? We access `.standard_sources` in `source_files`, but should we use `.get_standard_sources_mut`? I haven't thought this through

crates/isograph_schema/src/isograph_database.rs
```diff
@@ -67,22 +73,82 @@ impl<TNetworkProtocol: NetworkProtocol + 'static> IsographDatabase<TNetworkProto
...
+            .schema_extension_sources
+            .remove(&relative_path)
+            .inspect(|&source_id| self.remove(source_id))
```
rbalicki2, line 98: til about `.inspect`!

crates/isograph_schema/src/isograph_database.rs
```diff
@@ -6,16 +6,22 @@ use std::{
...
 use isograph_config::CompilerConfig;
 use pico::{Database, SourceId, Storage};
-use pico_macros::{Db, Singleton, Source, memo};
+use pico_macros::{Db, Source};
 
 use crate::NetworkProtocol;
 
 #[derive(Default, Debug, Db)]
 pub struct IsographDatabase<TNetworkProtocol: NetworkProtocol> {
     pub storage: Storage<Self>,
```
rbalicki2: storage doesn't need to be pub

crates/pico_macros/src/db_macro.rs
```diff
@@ -1,52 +1,123 @@
...
+        getter_impls.push(quote! {
+            #[inline]
+            pub fn #get_ident(&self) -> &#field_ty {
+                let _ = self.storage.get_singleton::<#counter_ident>();
+                &self.#field_ident
+            }
+
+            #[inline]
+            pub fn #get_ident_mut(&mut self) -> &mut #field_ty {
+                let next = self.storage.get_singleton::<#counter_ident>().copied().unwrap_or_default().increment();
```
rbalicki2: `.map(|x| x.increment())` is oh-so-slightly more efficient! No need to increment if its the first access

rbalicki2 (commented):
Partial review... will keep digging

crates/pico/tests/tracking_fields.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    // Instead we have to use tracked iter together with tracked insert
+    // (first call here, so value is correct anyway)
+    assert_eq!(get_values_tracked(&db).len(), 3);
+
+    let input_id_4 = db.set(Input {
+        key: "key4",
+        value: "database".to_string(),
+    });
+    db.get_map_tracked_mut()
```
rbalicki2: seems like we're doing a lot of `let_source_id = db.set(source); db.map_tracked_mut.set(source_id)`. I wonder if we should (perhaps as a follow up) put that in a separate function or something, so that we're less likely to mess up

crates/pico/src/source.rs
```diff
@@ -37,13 +37,26 @@ impl<T> Hash for SourceId<T> {
     }
 }
 
+impl<T> Default for SourceId<T> {
+    fn default() -> Self {
+        Self {
+            key: Default::default(),
```
rbalicki2, line 43: Just wondering — is this an "invalid" key? (Which shouldn't be a problem, we never attempt to convert the key to a value, i.e. "unhash" the key)

crates/pico/src/execute_memoized_function.rs
rbalicki2: Is this a bug on the main branch right now?

ch1ffa: Kind of. It's a leak, we never remove a source now and GC doesn't check for sources (purposely)

crates/pico_macros/src/db_macro.rs
```diff
@@ -1,52 +1,123 @@
...
+
+        if is_phantom_data(&f.ty) {
+            continue;
+        }
+
+        let field_ty = &f.ty;
+        let counter_ident =
+            format_ident!("__{}Counter", field_ident.to_string().to_case(Case::Pascal));
+        let get_ident = format_ident!("get_{field_ident}_tracked");
+        let get_ident_mut = format_ident!("get_{field_ident}_tracked_mut");
```
rbalicki2: should we also add `get_{field_ident}_untracked` and make the fields private?

crates/pico/src/source.rs
ch1ffa: Yes, this is invalid key. Not sure if we ever need to check for it, but we can using is_empty

crates/pico_macros/src/db_macro.rs
```diff
@@ -1,52 +1,156 @@
...
+        field_ident: syn::Ident,
+        field_ty: Type,
+    }
+
+    let mut tracked = Vec::new();
+    let mut errors = Vec::new();
+
+    for f in fields.named.iter() {
+        let field_ident = f.ident.as_ref().unwrap();
+        if has_attr(&f.attrs, "tracked") {
```
rbalicki2, line 31: not a blocker, but can we use `deluxe` (which is another version of a library you recommended I believe, and it's used in resolve position macro)? Maybe it's just extracting an empty tuple but it would be great to be consistent and make it easy to add additional props when needed

crates/isograph_schema/src/isograph_database.rs
```diff
@@ -67,22 +76,88 @@ impl<TNetworkProtocol: NetworkProtocol + 'static> IsographDatabase<TNetworkProto
...
     }
 
-    pub fn get_open_file_map(&self) -> &OpenFileMap {
-        self.get_singleton::<OpenFileMap>()
-            .expect("Expected OpenFileMap to have been set")
+    pub fn get_open_file(
+        &self,
+        file: RelativePathToSourceFile,
+    ) -> Option<SourceId<OpenFileSource>> {
+        self.open_file_map.0.get(&file).cloned()
```
rbalicki2: Is my understanding correct? because we're inside of `IsographDatabase`, we have access to the private `open_file_map`, so we're able to do this. And it's an untracked get.

crates/isograph_compiler/src/compiler_state.rs
```diff
@@ -27,7 +26,6 @@ impl<TNetworkProtocol: NetworkProtocol> CompilerState<TNetworkProtocol> {
         let mut db = IsographDatabase::default();
         db.set(current_working_directory);
```
rbalicki2, line 27: In the future, can we make `current_working_directory` a property of isograph database? Should we?

crates/pico/src/view.rs
```diff
@@ -0,0 +1,65 @@
...
+impl<'a, Db: Database, T, C: Counter> View<'a, Db, T, C> {
+    pub fn new(db: &'a Db, projector: Projector<Db, T>) -> Self {
+        Self {
+            db,
+            projector,
+            phantom: PhantomData,
+        }
+    }
+
+    pub fn tracked(&'a self) -> &'a T {
```
rbalicki2, line 29: Can we add a comment please?

```
/// Provides tracked access to the underlying field. On a tracked map, tracked access is
/// required for correctness if adding or removing an item from the underlying map
/// affects the result. For example, when iterating, use tracked access.
```

And for untracked

```
/// Provides untracked access to the underlying field. On a tracked map, untracked access is
/// only correct if every accessed value is later tracked, and unrelated additions to or
/// removals from the underlying map do not affect the result. For example, if the map's values
/// are themselves [`SourceId`](SourceId) and you are accessing a single, specific value by
/// key, it is safe to use an untracked access.
///
/// This may improve performance, as it means that unrelated additions and removals to the map
/// will not invalidate this function call in pico.
```

and likewise in ProjectorMut

crates/pico/tests/tracking_fields.rs
```diff
@@ -0,0 +1,116 @@
...
+        .insert("key".to_string(), input_id_1);
+    // First calculation is always correct, because we call inner_fn here
+    assert_eq!(get_values_untracked(&db).len(), 1);
+
+    let input_id_2 = db.set(Input {
+        key: "key2",
+        value: "isograph".to_string(),
+    });
+    // If we insert a new node without tracking and call untracked iter()
+    // we got the previous value
```
rbalicki2: `we get the previous value`

crates/pico/src/view.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub fn tracked(&'a mut self) -> &'a mut T {
+        let next = self
+            .db
+            .get_singleton::<C>()
+            .map_or(C::default(), |c| c.increment());
+        self.db.set(next);
+        (self.projector)(self.db)
+    }
+
+    pub fn untracked(&'a mut self) -> &'a mut T {
```
rbalicki2: I don't think we need an `untracked` method in `ProjectorMut`. We only use it (afaict, haven't checked it out locally) when initializing the DB. And it isn't a big deal to use tracked there, and everywhere else it's probably a mistake

ch1ffa: I think it can be used when we update `schema_source_id` in watch mode. We don't need to invalidate the whole `standard_sources` struct here, but it works only because `schema` is not a hashmap.

crates/isograph_compiler/src/compiler_state.rs
ch1ffa: Interesting that we can store it as a plain struct (not interned) and internal generated `Counter` will represent tracked state for it. But `singleton` do the same, so idk what is a preferred way.

crates/isograph_schema/src/isograph_database.rs
ch1ffa: it's a leftover. I would write `untracked` here for clarity. But yeah, for methods we have access to private fields and it is easy to do something wrong here

crates/isograph_compiler/src/compiler_state.rs
rbalicki2: I think fields on db are better than singletons, because you don't have to deal with the invalid state of the db existing, but the singleton not having been set, so I would be supportive of getting rid of singletons tbh. I don't think they're needed anywhere

crates/isograph_schema/src/isograph_database.rs
rbalicki2: okay cool then let's call `tracked` here for clarity

PR #648, opened 2025-09-07
Gracefully support Query.node in @exposeField; clearer behavior when missing

randallb:
Summary

- Fixes crash when auto-generated refetch fields use `field: "node.asType"` but the schema lacks `Query.node`.
- Treats `node` as rooted at the Query type and checks for existence.
- If `Query.node` is absent, skips generating the refetch field instead of erroring (and returns a more specific error variant internally).

Motivation

- On commits around 0.0.0-main-620f4c6f, any object implementing `Node` triggers a refetch field via `@exposeField` (`field: "node.as<Obj>"`). In schemas without `Query.node`, compilation fails with "Invalid field `node` in @exposeField directive".
- This breaks real-world schemas that implement `Node` but intentionally do not expose `Query.node`.

Changes

- create_additional_fields/expose_field_directive.rs:
  - Resolve top-level "node" against Query root when present.
  - Emit a PrimaryDirectiveFieldNotFound error if `Query.node` is absent (clearer than a generic InvalidField).
- isograph_compiler/create_schema.rs:
  - Gracefully skip generating the refetch field when the error indicates that `Query.node` is missing; continue compiling the rest of the schema.
- Dev quality-of-life: add flake.nix and .envrc for a reproducible dev shell with cargo/deno/node.

Notes

- This preserves the refetch field behavior for projects that have `Query.node`, while making compilation resilient for projects that don’t.
- Alternatively, the compiler could hard-require `Query.node` whenever any type implements `Node`; this PR aims to be permissive and avoid breakages while improving the error.

Testing

- Verified with a minimal schema reproducer and with a larger application. In the reproducer without `Query.node`, the compiler proceeds without failing.

Related gist/repro: https://gist.github.com/randallb/df13780a215c74f3f9e05d81f4fe6e03

### DIFF
--- a/crates/tests/tests/refetch_query_node_present.rs
+++ b/crates/tests/tests/refetch_query_node_present.rs
@@ -1,0 +1,108 @@
+use std::{fs, path::PathBuf};
+
+use common_lang_types::{ClientScalarSelectableName, CurrentWorkingDirectory, SelectableName, ServerObjectEntityName};
+use graphql_network_protocol::GraphQLNetworkProtocol;
+use intern::string_key::Intern;
+use isograph_compiler::{
+    create_schema::create_schema,
+    get_validated_schema::process_iso_literals_for_schema,
+    source_files::initialize_sources,
+};
+use isograph_config::{
+    absolute_and_relative_paths,
+    compilation_options::{CompilerConfig, CompilerConfigOptions, GenerateFileExtensionsOption},
+};
+use isograph_schema::IsographDatabase;
+
+fn write_file(path: &PathBuf, contents: &str) {
+    if let Some(dir) = path.parent() {
+        fs::create_dir_all(dir).expect("failed to create directories");
+    }
+    fs::write(path, contents).expect("failed to write file");
+}
+
+#[test]
+fn refetch_node_present_generates_refetch_field_on_object() {
+    // Create a temp workspace
+    let tmp_root = std::env::temp_dir().join("isograph-refetch-node-present-test");
+    let _ = fs::remove_dir_all(&tmp_root);
+    fs::create_dir_all(&tmp_root).expect("failed to create tmp root");
+
+    // Project structure
+    let project_root = tmp_root.join("project");
+    let schema_path = project_root.join("schema.graphql");
+    let artifact_dir = project_root.join("__isograph");
+    fs::create_dir_all(&artifact_dir).expect("failed to create artifact dir");
+
+    // Schema with Node implementer AND Query.node defined
+    let schema = r#"
+schema { query: Query }
+
+interface Node { id: ID! }
+
+type Item implements Node { id: ID!, name: String }
+
+type Query {
+  node(id: ID!): Node
+}
+"#;
+    write_file(&schema_path, schema);
+
+    // Prepare DB and config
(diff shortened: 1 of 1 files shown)


randallb:
(generated with gpt5 pro)

randallb:
per patryk:

https://github.com/isographlabs/isograph/blob/f6a929bffd8f2fa0d4fb53c50caf5f4b1ab0f7c4/crates/graphql_network_protocol/src/process_type_system_definition.rs#L393
here we need to add another condition, object_implements_node && Query.node is "valid" don't know if we can do it at this step

crates/isograph_compiler/src/create_schema.rs
```diff
@@ -86,10 +86,30 @@ pub fn create_schema<TNetworkProtocol: NetworkProtocol + 'static>(
 
     for (parent_object_entity_name, expose_as_fields_to_insert) in expose_as_field_queue {
         for expose_as_field in expose_as_fields_to_insert {
-            let unprocessed_scalar_item = unvalidated_isograph_schema
-                .create_new_exposed_field(expose_as_field, parent_object_entity_name)?;
-
-            unprocessed_items.push(SelectionType::Scalar(unprocessed_scalar_item));
+            match unvalidated_isograph_schema
+                .create_new_exposed_field(expose_as_field, parent_object_entity_name)
```
rbalicki2, line 90: let's have `create_new_exposed_field` return a `Result<Option<...>>` and leave this code unchanged

crates/isograph_schema/src/create_additional_fields/expose_field_directive.rs
```diff
@@ -78,8 +78,52 @@ impl<TNetworkProtocol: NetworkProtocol> Schema<TNetworkProtocol> {
...
+            }
+        } else {
+            parent_object_entity_name
+        };
+
+        // If the field is `node`, ensure that the Query root has a corresponding field.
+        if field == "node" {
+            let Some((query_name, _)) = self.find_query() else {
+                return Err(WithLocation::new(
+                    CreateAdditionalFieldsError::PrimaryDirectiveFieldNotFound {
```
rbalicki2, line 101: and here, let's return `None`

crates/tests/tests/refetch_query_node_absent.rs
```diff
@@ -0,0 +1,101 @@
...
+    write_file(&schema_path, schema);
+
+    // Prepare DB and config
+    let mut db: IsographDatabase<GraphQLNetworkProtocol> = Default::default();
+    let cwd_str = tmp_root
+        .canonicalize()
+        .expect("canonicalize cwd")
+        .to_string_lossy()
+        .to_string();
+    db.set(CurrentWorkingDirectory(cwd_str.intern().into()));
```
rbalicki2, line 69: we don't need to use the `tmp` folder, we can a bunch of calls to `db.set(...)` instead. Happy to figure out how to make that work unless you're interested in it. But not a blocker, the test lgtm

randallb:
I just solved on my own by just implementing it.

rbalicki2:
Okay! I'm glad you're not blocked! FWIW not requiring node fields is something that @tbezman will tackle soon!

PR #649, opened 2025-09-08
[plugin_swc] Upgrade to 0.4.0 with swc_core 3.0.0, @swc/core 1.7.42

Pablinho:
Update swc_core


libs/isograph-swc-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@isograph/plugin-swc",
-  "version": "0.3.1",
+  "version": "0.4.0",
```
rbalicki2: let's keep this at 0.3.1, I'll bump it to 0.4.0 when we bump all the packages.

rbalicki2 (commented):
how do I test this? as in, does it work with the latest nextjs?

PR #662, opened 2025-09-16
fix client pointer `output_type.ts`

PatrykWalach:
Let's infer pointer output type here.

 In the future we can use it to map types with typescript, for example if pointer returns array of length two `[Link, Link]`, we map it so the loadable would return an array of length two as well `[{ ... }, { ... }]`


rbalicki2 (approved):
SGTM shipit

PR #665, opened 2025-09-16
fix client pointer equality check

### LINKED ISSUE #664: Pointers and areEqualWithDeepComparison
When we call mergeObjectsUsingReaderAst, we attempt to set .name on a function:

```
{kind: 'ErrorEncounteredInWithErrorHandling', error: TypeError: Cannot assign to read only property 'name' of function '(args, 
            // TODO get …}
error
: 
TypeError: Cannot assign to read only property 'name' of function '(args, // TODO get the associated type for FetchOptions from the loadably selected field ...<omitted>... }' at mergeObjectsUsingReaderAst (webpack-`
```

This is probably due to a selection on a client pointer including a field named name. This is now surfaced via the 'ErrorEncounteredInWithErrorHandling' message. 

I haven't confirmed, but one can repro this by checking out https://github.com/rbalicki2/follow-the-money and changing suspectName and caseName to name, then clicking the "set debts to zero" button. It doesn't re-execute the pointer, because the subscription isn't fired, because of this error.

PatrykWalach:
Fixes #664

### DIFF
--- a/libs/isograph-react/src/core/areEqualWithDeepComparison.ts
+++ b/libs/isograph-react/src/core/areEqualWithDeepComparison.ts
@@ -72,6 +72,11 @@ export function mergeObjectsUsingReaderAst(
     switch (field.kind) {
       case 'Scalar':
       case 'Linked':
+        if (field.kind === 'Linked' && field.refetchQueryIndex != null) {
+          // client pointers are functions, so we can't merge them
+          canRecycle = false;
+          break;
+        }
         const key = field.alias ?? field.fieldName;
         // @ts-expect-error
         const oldValue = oldItemObject[key];


libs/isograph-react/src/core/areEqualWithDeepComparison.ts
```diff
@@ -72,6 +72,10 @@ export function mergeObjectsUsingReaderAst(
     switch (field.kind) {
       case 'Scalar':
       case 'Linked':
+        if (field.kind === 'Linked' && field.refetchQueryIndex != null) {
+          // client pointers are functions, so we can't merge them
```
PatrykWalach, line 76: I think we should do `canRecycle = false;` here,  stuff like `LoadablySelectedField` depends on `id` so they're stable but pointers generally depend on other fields, so we should always assume they're different

rbalicki2: A loadably selected field never changes — so I think `continue` is correct here.

(Ignoring array pointers... probably more of the same). Non-array pointers are a bit more complicated. The returned value is a function `(args) => [stableId1, fetcher]`. `stableId1` is a combination of the `typename`, `link`, `fieldName`, and `stableStringifyArgs(args + readOutParameters)`. *But* the args are provided by the user, so what will determine if we need to re-read is the `typename`, `link`, `fieldName` and `stableStringifyArgs(readOutParameters)` (call this `stableId2`). We need to expose `stableId2` on the returned object, i.e. not require you to pass args.

`stableId1` is still necessary! If you're deduping network requests, e.g. in useClientSideDefer, it's necessary.

So I see several paths:
- change the read-out type of a loadable field. Don't make it a function, but instead, make it a tuple `[stableId, applyArgs]` or an object `{ stableId, fetcher }`. (And we should use the same data structure twice, i.e. either `[StableId2, (args) => [stableId1, Fetcher]]` or { stableId: stableId2, fetcher: (args) => { stableId: stableId1, fetcher: () => ItemAndCleanupFn<...> }` ⭐
- put the stable Id on the fecher, i.e. have `loadableFieldFunction[stableIdSymbol] = stableId2`.

I prefer the ⭐ option. Objects seem cleaner than tuples, and I would support using objects everywhere, tbh

Edit: `type WithStableId = { stableId, item: T }` seems like the right type to use?

rbalicki2: So, I think we can do `continue` and `canRecycle = false` for the time being, and file ☝️  as an issue

rbalicki2: #668

PR #672, opened 2025-09-19
fix subscriptions last read data not being updated

PatrykWalach:
Closes ##327

### DIFF
--- a/libs/isograph-react/src/core/cache.ts
+++ b/libs/isograph-react/src/core/cache.ts
@@ -330,6 +330,8 @@ export function callSubscriptions(
 
             if (mergedItem !== subscription.encounteredDataAndRecords.item) {
               subscription.callback(newEncounteredDataAndRecords);
+              subscription.encounteredDataAndRecords =
+                newEncounteredDataAndRecords;
             }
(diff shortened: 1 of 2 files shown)


libs/isograph-react/src/core/cache.ts
```diff
@@ -330,6 +330,8 @@ export function callSubscriptions(
 
             if (mergedItem !== subscription.encounteredDataAndRecords.item) {
               subscription.callback(newEncounteredDataAndRecords);
+              subscription.encounteredDataAndRecords =
+                newEncounteredDataAndRecords;
```
PatrykWalach, line 334: If callback throws an errors we might be in a invalid state, not sure if we should interpret it as - they didn't get the data so we should not update

rbalicki2: That's the subscription's fault. We shouldn't break the `cache.ts` file's behavior if someone registered a callback that throws.

We should wrap `callback` in `withErrorHandling`. I'll do that as a follow up

rbalicki2 (approved):
SHIPT

PR #683, opened 2025-09-23
Register dependency on memoref deref

ch1ffa:
This allows us to track memoized functions that do not have access to sources and do memoization over `MemoRef` arguments. We also deduplicate registration in the parent memoized function if it happens immediately after function call.

### DIFF
--- a/crates/pico/tests/basic_multi_function_chain.rs
+++ b/crates/pico/tests/basic_multi_function_chain.rs
@@ -96,6 +96,7 @@ fn sequential_functions_with_memoref_param() {
         value: "bsdf".to_string(),
     });
 
+    assert_eq!(*first_letter(&db, id), 'b');
     assert_eq!(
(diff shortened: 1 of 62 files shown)


crates/pico/tests/basic_multi_function_chain.rs
```diff
@@ -70,6 +71,51 @@ fn multi_function_chain_with_irrelevant_change() {
     assert_eq!(CAPITALIZED_LETTER_COUNTER.load(Ordering::SeqCst), 1);
 }
 
+#[test]
+fn sequential_functions_with_memoref_param() {
```
rbalicki2, line 75: would you mind landing a PR above this one that demonstrates the broken behavior, then stacking this on that one?

ch1ffa: done

PR #684, opened 2025-09-23
add helper methods to project MemoRef on some types

ch1ffa:
Sometimes we need to use Ok value from `MemoRef<Result<T,E>>` without cloning because it is supposed to be passed as an argument to another memoized function as `MemoRef<T>` for performance reasons. This PR allows us to transform `MemoRef<Result<T,E>>` to `Result<MemoRef<T>,E>` without cloning the whole `Result` struct - only `Error` is cloned which is usually used with `?` operator.

Same applies to `MemoRef<Option<T>>` and `MemoRef<(T1, T2)>`. And we also can chain this calls (up to 4) like `try_ok()?.split()`.

I have to remove optimisation when `MemoRef` skips interning as param, but it's not a big deal - it's cheap to hash and copy.

NOTE: stacked on #683

### DIFF
--- a/crates/pico/src/memo_ref.rs
+++ b/crates/pico/src/memo_ref.rs
@@ -127,7 +127,12 @@ impl<T: 'static> Deref for MemoRef<T> {
             NodeKind::Derived(self.derived_node_id),
             revision.time_updated,
         );
-        let mut any_ref: &dyn Any = value;
+        if self.projectors_len == 0 {
+            return value
+                .downcast_ref::<T>()
+                .expect("MemoRef: projector chain produced unexpected type");
+        }
+        let mut any_ref = value;
         let len = self.projectors_len as usize;
         for step in &self.projectors[..len] {
             any_ref = (step)(any_ref);
@@ -146,12 +151,11 @@ impl<T: 'static, E: 'static + Clone> MemoRef<Result<T, E>> {
                     db: self.db,
                     derived_node_id: self.derived_node_id,
                     projectors: self.projectors,
-                    projectors_len: self.projectors_len,
+                    projectors_len: self.projectors_len + 1,
                     phantom: PhantomData,
                 };
-                let idx = next.projectors_len as usize;
+                let idx = self.projectors_len as usize;
                 next.projectors[idx] = step_result_ok::<T, E>;
-                next.projectors_len += 1;
                 Ok(next)
             }
             Err(err) => Err(err.clone()),
@@ -167,12 +171,11 @@ impl<T: 'static> MemoRef<Option<T>> {
                     db: self.db,
                     derived_node_id: self.derived_node_id,
                     projectors: self.projectors,
-                    projectors_len: self.projectors_len,
+                    projectors_len: self.projectors_len + 1,
                     phantom: PhantomData,
                 };
-                let idx = next.projectors_len as usize;
+                let idx = self.projectors_len as usize;
                 next.projectors[idx] = step_option_some::<T>;
-                next.projectors_len += 1;
                 Some(next)
             }
             None => None,
@@ -186,23 +189,21 @@ impl<T0: 'static, T1: 'static> MemoRef<(T0, T1)> {
             db: self.db,
             derived_node_id: self.derived_node_id,
             projectors: self.projectors,
-            projectors_len: self.projectors_len,
+            projectors_len: self.projectors_len + 1,
             phantom: PhantomData,
         };
-        let idx = left.projectors_len as usize;
+        let idx = self.projectors_len as usize;
         left.projectors[idx] = step_tuple_0::<T0, T1>;
-        left.projectors_len += 1;
 
         let mut right = MemoRef::<T1> {
             db: self.db,
             derived_node_id: self.derived_node_id,
             projectors: self.projectors,
-            projectors_len: self.projectors_len,
+            projectors_len: self.projectors_len + 1,
             phantom: PhantomData,
         };
-        let idx = right.projectors_len as usize;
+        let idx = self.projectors_len as usize;
         right.projectors[idx] = step_tuple_1::<T0, T1>;
-        right.projectors_len += 1;
 
         (left, right)
     }
--- a/crates/pico/tests/try_some.rs
+++ b/crates/pico/tests/try_some.rs
@@ -15,9 +15,9 @@ fn try_some_projects_some_value() {
         value: "asdf".to_string(),
     });
 
-    let Some(result) = maybe_first_letter(&db, id).try_some() else {
-        panic!("expected Some")
-    };
+    let result = maybe_first_letter(&db, id)
+        .try_some()
+        .expect("expected Some");
     assert_eq!(*result, 'a');
 }
 
@@ -50,9 +50,7 @@ fn try_some_never_clones_value() {
     let db = TestDatabase::default();
 
     let some_ref = some_value(&db);
-    let Some(inner) = some_ref.try_some() else {
-        panic!("expected Some")
-    };
+    let inner = some_ref.try_some().expect("expected Some");
     let _ = &*inner;
 
     assert!(none_value(&db).try_some().is_none());


crates/pico/src/memo_ref.rs
```diff
@@ -1,13 +1,66 @@
...
 use intern::InternId;
 
-use crate::{DatabaseDyn, DerivedNodeId, ParamId, dependency::NodeKind};
+type MemoRefProjectorStep = for<'a> fn(&'a dyn Any) -> &'a dyn Any;
+
+const MAX_PROJECTOR_STEPS: usize = 4;
+
+#[inline(always)]
+fn step_identity(value: &dyn Any) -> &dyn Any {
+    value
```
rbalicki2, line 17: this can panic, right? The len should always stop before we reach a step_identity, right?

crates/pico/src/memo_ref.rs
```diff
@@ -63,6 +127,83 @@ impl<T: 'static> Deref for MemoRef<T> {
...
+    }
+}
+
+impl<T0: 'static, T1: 'static> MemoRef<(T0, T1)> {
+    pub fn split(self) -> (MemoRef<T0>, MemoRef<T1>) {
+        let mut left = MemoRef::<T0> {
+            db: self.db,
+            derived_node_id: self.derived_node_id,
+            projectors: self.projectors,
+            projectors_len: self.projectors_len,
```
rbalicki2: it would be cleaner, I think, to just say `self.projectors_len + 1`. Wdyt?

crates/pico/src/memo_ref.rs
```diff
@@ -1,13 +1,66 @@
...
+        .expect("MemoRef<(..)>: underlying value has unexpected type");
+    t1 as &dyn Any
+}
 
 #[derive(Debug)]
 pub struct MemoRef<T> {
     pub(crate) db: *const dyn DatabaseDyn,
     pub(crate) derived_node_id: DerivedNodeId,
+    projectors: [MemoRefProjectorStep; MAX_PROJECTOR_STEPS],
+    projectors_len: u8,
```
rbalicki2, line 63: can we add some comments here explaining what's happening?

crates/pico/src/memo_ref.rs
```diff
@@ -1,13 +1,66 @@
...
+        .downcast_ref::<(T0, T1)>()
+        .expect("MemoRef<(..)>: underlying value has unexpected type");
+    t1 as &dyn Any
+}
 
 #[derive(Debug)]
 pub struct MemoRef<T> {
     pub(crate) db: *const dyn DatabaseDyn,
     pub(crate) derived_node_id: DerivedNodeId,
+    projectors: [MemoRefProjectorStep; MAX_PROJECTOR_STEPS],
```
rbalicki2, line 62: I think that the common case is that `projectors_len` is 0. So, perhaps it would be more performant to have an `projectors: Option<...>` (and no `projectors_len`)? (Do you think this is true? I'm not 100% convinced)

If it's indeed true that the common case is 0/`None`, then perhaps we don't need a fixed-length array, but can instead have a `Vec<...>`.

Oh also, `Vec` of length zero doesn't heap allocate anything, so perhaps we just have `projectors: Vec<...>`?

crates/pico/src/memo_ref.rs
```diff
@@ -63,6 +127,83 @@ impl<T: 'static> Deref for MemoRef<T> {
             NodeKind::Derived(self.derived_node_id),
             revision.time_updated,
         );
-        value.downcast_ref::<T>().unwrap()
+        let mut any_ref: &dyn Any = value;
+        let len = self.projectors_len as usize;
+        for step in &self.projectors[..len] {
+            any_ref = (step)(any_ref);
+        }
+        any_ref
```
rbalicki2, line 140: hm. Can we optimize the case where `len == 0` and avoid the `any_ref = any_ref as dyn Any; any_ref.downcast_ref()`?

crates/pico/tests/try_ok.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
+enum FirstLetterError {
+    #[error("empty string")]
+    EmptyString,
+}
+
+#[derive(Debug, Clone, Error, PartialEq, Eq)]
+enum ProcessInputError {
```
rbalicki2, line 52: not that I'm opposed to it per se, but does deriving Error get us anything in this test?

crates/pico/tests/try_ok.rs
```diff
@@ -0,0 +1,125 @@
...
+fn projected_memo_correctly_restored() {
+    let mut db = TestDatabase::default();
+
+    let id = db.set(Input {
+        key: "key",
+        value: "asdf".to_string(),
+    });
+
+    let result = process_input(&db, id).unwrap();
+    assert_eq!(*consume_result(&db, result), 'a');
```
rbalicki2, line 35: What are we testing in this test that we aren't testing in `try_ok_projects_ok_value`?

crates/pico/tests/try_ok_split.rs
```diff
@@ -0,0 +1,34 @@
...
+        .try_ok()
+        .expect("expected Ok result before split")
+        .split();
+
+    let _ = &*left;
+    let _ = &*right;
+}
+
+#[memo]
+fn ok_pair_value(_db: &TestDatabase) -> Result<(PanicOnClone, PanicOnClone), ()> {
```
rbalicki2, line 23: QQ: what need do we have for tuples? Is that a specific, concrete thing we need? (Not opposed, just wondering)

crates/pico/tests/try_some.rs
```diff
@@ -0,0 +1,78 @@
...
+fn maybe_first_letter(db: &TestDatabase, input_id: SourceId<Input>) -> Option<char> {
+    db.get(input_id).value.chars().next()
+}
+
+#[test]
+fn try_some_never_clones_value() {
+    let db = TestDatabase::default();
+
+    let some_ref = some_value(&db);
+    let Some(inner) = some_ref.try_some() else {
```
rbalicki2: nit: `.expect(...)`?

crates/pico/src/memo_ref.rs
```diff
@@ -27,12 +80,23 @@ impl<T> PartialEq for MemoRef<T> {
...
+}
+
 #[allow(clippy::unnecessary_cast)]
 impl<T: 'static + Clone> MemoRef<T> {
     pub fn new(db: &dyn DatabaseDyn, derived_node_id: DerivedNodeId) -> Self {
         Self {
             db: db as *const _ as *const dyn DatabaseDyn,
             derived_node_id,
+            projectors: [step_identity; MAX_PROJECTOR_STEPS],
+            projectors_len: 0,
```
rbalicki2, line 99: Coming back to this... if the typical thing is `len == 0` (is it?) then, perhaps we can take inspiration from how iterators work: https://doc.rust-lang.org/src/core/iter/adapters/map.rs.html#63 they are structs that wrap the parent, and have `impl MemoRef<Item = T>` everywhere? Then, we need only `Map` and `FlatMap` structs that wrap the underlying memo ref.

Is that doable? Worse than this solution? I don't know. I'm just pattern-matching on how stuff is done elsewhere in Rust

rbalicki2 (commented):
more nits/comments.

crates/pico/src/memo_ref.rs
ch1ffa: array is always initialised with `step_identity` function pointers. But because we use `len: 0` it's never called. If somehow there is a weird bug that increases len, then `step_identity` is almost noop. The only optimisation I believe we can apply here is using `MaybeUninit`, but it is tiny.

crates/pico/src/memo_ref.rs
ch1ffa: we need `len` to properly iterate over projectors chain and avoid calling unnecessary `identity_step`. I don't think `Option` gives us anything here, there is still a check for None/Some and it is the same as check for `len` in for loop.

crates/pico/src/memo_ref.rs
ch1ffa: I think we can check for `len == 0` and avoid additional pointer copy and bound check.

crates/pico/tests/try_ok_split.rs
ch1ffa: we use it after `parse_graphql_schema`, because we need to clone one part and pass the second one as `memo_ref`

crates/pico/tests/try_ok.rs
ch1ffa: Yeah, we can just use a struct here. Because it is a `Result` I just without any thought used `Error` lol

crates/pico/tests/try_ok.rs
ch1ffa: we check if memo_ref that were passed as an argument is correctly cloned from the store with proper projector

PR #690, opened 2025-09-26
feat: allow loadables on the root query type

tbezman:
Using `@loadable` on a field defined on the root `Query` type failed before this PR with 

```bash
(base) ➜  isograph-play git:(main) ✗ bun iso
   INFO  Starting to compile.


thread 'main' panicked at /home/runner/work/isograph/isograph/crates/generate_artifacts/src/reader_ast.rs:533:14:
Expected refetch strategy. This is indicative of a bug in Isograph.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: "iso" exited with code 101
```

This is because the only refetch strategy that was implemented was to reselect data from the server via the `node(id)` query. 

To solve this, we add support for a new strategy `RefetchFromRoot` and update all callsites of `RefetchStrategy` to properly support this new strategy.

*Followups*
- [ ] Ensure selection sets with only loadable fields include a `__typename` at a bare minimum to prevent empty selection sets.
- [ ] Update the type for `useConnectionSpecPagination` to support a nullable connection type to conform with the Relay spec.


crates/generate_artifacts/src/generate_artifacts.rs
```diff
@@ -305,11 +290,42 @@ fn get_artifact_path_and_content_impl<TNetworkProtocol: NetworkProtocol>(
...
                             };
-                            let variable_definitions_iter = client_scalar_selectable
-                                .variable_definitions
-                                .iter()
-                                .map(|variable_definition| &variable_definition.item)
-                                .chain(std::iter::once(&id_var));
+
+                            let (wrapped_map, variable_definitions_iter): (_, Vec<_>) =
+                                match client_scalar_selectable.refetch_strategy.as_ref().expect(
+                                    "Expedted a refetch strategy. \
```
rbalicki2: type: `expected`

crates/isograph_schema/src/refetch_strategy.rs
```diff
@@ -25,7 +25,7 @@ pub enum RefetchStrategy<
             TSelectionTypeSelectionLinkedFieldAssociatedData,
         >,
     ),
-    // RefetchFromRoot
+    RefetchFromRoot(RootOperationName),
```
rbalicki2: we never use the `RootOperationName`. So I think this is broken for refetching things on `Mutation`. Can we add two test cases to `pet-demo`, where we refetch something that's on query and on mutation?

(If I have time... probably tomorrow, I can modify the pet demo to fetch from Mutation etc etc)

rbalicki2 (commented):
Overall, lgtm, lets just add the test cases, make sure that mutations use the mutation keyword, and fix the typo

rbalicki2:
LGTM shipittt

PR #697, opened 2025-10-02
Retain queries if data in store

### LINKED ISSUE #696: Retain query if data in store
Right now we `retainQuery` only after network request (in `makeNetworkRequest`).
We should `retainQuery` if the data is already in the store (in `maybeMakeNetworkRequest`)

PatrykWalach:
Closes #696


libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -351,3 +354,47 @@ function readDataForOnComplete<
...
+    variables,
+    root,
+  };
+  retainQuery(environment, retainedQuery);
+  return retainedQuery;
+}
+
+function unretainAndGarbageCollectIfUndisposed(
+  environment: IsographEnvironment,
+  status: NetworkRequestStatus,
```
rbalicki2: let's pass `status: UndisposedNetworkStatus`

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -60,7 +60,23 @@ export function maybeMakeNetworkRequest<
       );
     }
     case 'No': {
-      return [wrapResolvedValue(undefined), () => {}];
+      let status: NetworkRequestStatus = {
```
rbalicki2: let's make this a separate function and call it below as well

rbalicki2 (approved):
amazing, thanks for making the changes!!

PR #699, opened 2025-10-02
Throw graphql errrors for each component

PatrykWalach:


libs/isograph-react/src/core/cache.ts
```diff
@@ -486,22 +540,54 @@ export function insertEmptySetIfMissing<K, V>(map: Map<K, Set<V>>, key: K) {
...
   variables: Variables,
+  errorsByPath: ErrorsByPath,
+  path: (string | number)[],
 ): RecordHasBeenUpdated {
   const networkResponseKey = getNetworkResponseKey(astNode);
   const networkResponseData = networkResponseParentRecord[networkResponseKey];
   const parentRecordKey = getParentRecordKey(astNode, variables);
+  const existingValue = targetStoreRecord.record[parentRecordKey];
+
+  if (networkResponseData === null) {
```
PatrykWalach: This is the key logic we update errors only when network response is `null`

PatrykWalach:
Closed in favor of https://github.com/isographlabs/isograph/pull/808

PR #703, opened 2025-10-04
fix: implement reachable vars for inline fragments

tbezman:
Reachable variables was only partially implemented.


rbalicki2 (approved):
Hell yeah! Glad we found this bug

PR #707, opened 2025-10-04
support pointer returning array of ids

PatrykWalach:


libs/isograph-react/src/core/read.ts
```diff
@@ -914,6 +859,122 @@ export function readLinkedFieldData(
   return data;
 }
 
+function isClientPointer(
+  field: ReaderLinkedField,
+): field is ReaderClientPointer {
```
rbalicki2, line 865: nifty

libs/isograph-react/src/core/read.ts
```diff
@@ -914,6 +859,122 @@ export function readLinkedFieldData(
...
+  variables: Variables,
+  nestedRefetchQueries: RefetchQueryNormalizationArtifactWrapper[],
+  readData: <TReadFromStore>(
+    ast: ReaderAst<TReadFromStore>,
+    root: StoreLink,
+  ) => ReadDataResult<object>,
+): ReadDataResult<unknown> {
+  const refetchReaderParams = readData(
+    [
+      {
```
rbalicki2, line 882: should this be part of the reader AST somehow? Either of the parent reader AST (where the client pointer was selected) or of the client pointer's reader AST?

libs/isograph-react/src/core/read.ts
```diff
@@ -754,8 +711,79 @@ export function readLinkedFieldData(
...
+              '. Link is ' +
+              JSON.stringify(item),
+            nestedReason: result,
+            recordLink: result.recordLink,
+          };
+        }
+        results.push(result.data);
+      }
+
+      const result = readData(field.selections, link);
```
rbalicki2, line 765: Should we call this code, even if it's a client pointer? As in, won't the result array contain `readClientPointerData(...).data` and `readData(...).data`?

libs/isograph-react/src/core/read.ts
```diff
@@ -39,6 +39,7 @@ import {
 import {
   ReaderAst,
   type LoadablySelectedField,
+  type ReaderClientPointer,
```
rbalicki2, line 42: test

rbalicki2: wow, TIL you can review commits individually

PatrykWalach: Wow

libs/isograph-react/src/core/read.ts
PatrykWalach: Good catch, needs `continue;`, I'll change firstCheckin pointer to return two items, to make sure this is working as intended

libs/isograph-react/src/core/read.ts
PatrykWalach: yes eventually it should be

libs/isograph-react/src/core/read.ts
PatrykWalach: I've done the demo, it works as expected now

rbalicki2:
Looks like if we navigate to /smartest, it will suspend indefinitely, with:

```
{
    "kind": "DoneReading",
    "response": {
        "kind": "MissingData",
        "reason": "Missing data for null on root undefined",
        "nestedReason": {
            "kind": "MissingData",
            "reason": "No record for root undefined",
            "recordLink": [
                {
                    "__link": "Checkin__0",
                    "__typename": "Checkin"
                },
                {
                    "__link": "Checkin__1",
                    "__typename": "Checkin"
                }
            ]
        },
        "recordLink": [
            {
                "__link": "Checkin__0",
                "__typename": "Checkin"
            },
            {
                "__link": "Checkin__1",
                "__typename": "Checkin"
            }
        ]
    },
    "fieldName": "smartestPet",
    "root": {
        "__link": "0",
        "__typename": "Pet"
    }
}
```

investigating

rbalicki2:
Oops, forgot to run `watch-libs`. Never midn!

PR #717, opened 2025-10-11
fix Incorrect normalization AST generated for refetch query with abstract types #663

### LINKED ISSUE #663: Incorrect normalization AST generated for refetch query with abstract types
See rb/random-page.

We have a client pointer `to Node` called `firstNode` that we use like this:

```
    firstNode {
      asPet {
        nickname
      }
    }
```

The follow up query looks like:

```
export default 'query Query__firstNode($id: ID!) {\
  node____id___v_id: node(id: $id) {\
    ... on Node {\ # HERE
      __typename,\
      id,\
      ... on Pet {\
        __typename,\
        id,\
        nickname,\
      },\
    },\
  },\
}';
```

This results in a normalization AST containing `{ kind: "InlineFragment", type: "Node", selections: ... }` that is applied when we are iterating "# HERE" in the query.

Unfortunately, that means that we are checking the typename against the string "Node". The typename is `Pet`, though! So, we should find a way to avoid generating this intermediate layer here.

This is probably indicative of the fact that the `selection_map_wrapped` function is poorly designed.

------

A workaround is to have `pointer Query.firstNode to Pet`, then everything works fine. See `rb/fix-random-page`

PatrykWalach:
Closes #663

### DIFF
--- a/demos/pet-demo/src/components/SmartestPet.tsx
+++ b/demos/pet-demo/src/components/SmartestPet.tsx
@@ -1,7 +1,7 @@
 import { LoadableFieldReader } from '@isograph/react';
 import { Card, CardContent, Container, Stack } from '@mui/material';
 import { Suspense } from 'react';
-import { iso } from './__isograph/iso';
+import { iso } from '@iso';
 import { useNavigateTo } from './routes';
 
 export const SmartestPetRoute = iso(`
--- a/demos/pet-demo/src/components/SmartestPetLoader.tsx
+++ b/demos/pet-demo/src/components/SmartestPetLoader.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { iso } from './__isograph/iso';
+import { iso } from '@iso';
 import { ErrorBoundary } from './ErrorBoundary';
 import { FullPageLoading } from './routes';
 import { FragmentRenderer, useLazyReference } from '@isograph/react';


crates/generate_artifacts/src/imperatively_loaded_fields.rs
PatrykWalach: This PR is almost what we want, one extra thing that has to change is this comparison, it makes sense for scalar_selectable, but not for the object_selectables.

This works for the current example because `target_type` is different than `parent_type` `Pet.firstCheckin to ICheckin` but for `pointer ICheckin.toCheckin to ICheckin` it would not work

rbalicki2: good catch

crates/isograph_schema/src/create_merged_selection_set.rs
```diff
@@ -860,6 +862,17 @@ fn merge_client_object_field<TNetworkProtocol: NetworkProtocol>(
             parent_object_entity_name,
             newly_encountered_client_object_selectable_id,
         )));
+
+    // this is theoretically wrong, we should be adding this to the client pointer
+    // traversal state instead of it's parent, but it works out the same
```
PatrykWalach, line 868: This is because we cannot have a client pointer entrypoint, therefore we can always insert to the parent, if we had client pointer entrypoints we wouldn't generate the type for the `Link`

PR #724, opened 2025-10-18
fix `ClientPointer` args not in a correct place

PatrykWalach:


rbalicki2:
good catch, thank you!!

PR #727, opened 2025-10-18
fix: make types of useClientSideDefer correct

tbezman:
before this PR, you could have a loadable field that took args, but `useClientSideDefer` would not complain if you didn't pass them. 

This PR fixes this by getting rid of the overloads and using a conditional type to ensure you land with the right call signature.

### DIFF
--- a/libs/isograph-react/src/core/FragmentReference.ts
+++ b/libs/isograph-react/src/core/FragmentReference.ts
@@ -17,8 +17,8 @@ export type VariableValue =
 
 export type Variables = { readonly [index: string]: VariableValue };
 
-export type UnknownTReadFromStore<TParameters extends object = object> = {
-  parameters: TParameters;
+export type UnknownTReadFromStore = {
+  parameters: object;
   data: object;
   startUpdate?: StartUpdate<object>;
 };


libs/isograph-react/src/react/LoadableFieldReader.tsx
```diff
@@ -8,25 +8,56 @@ import { type LoadableField } from '../core/reader';
...
-  fetchOptions?: FetchOptions<TResult>;
-  networkRequestOptions?: Partial<NetworkRequestReaderOptions>;
-  children: (arg: TResult) => TChildrenResult;
-}): TChildrenResult {
+    TChildrenResult
+  >,
+): TChildrenResult {
   const { fragmentReference } = useClientSideDefer(
     props.loadableField,
+    // @ts-expect-error
```
tbezman, line 50: TypeScript is too dumb to reconcile the types here. The signature of the function is still correct.

libs/isograph-react/src/react/LoadableFieldRenderer.tsx
```diff
@@ -9,25 +9,60 @@ import { type LoadableField } from '../core/reader';
...
-  networkRequestOptions?: Partial<NetworkRequestReaderOptions>;
-  additionalProps: Omit<TProps, keyof JSX.IntrinsicAttributes>;
-}): TChildrenResult {
+    TChildrenResult,
+    TProps
+  >,
+): TChildrenResult {
   const { fragmentReference } = useClientSideDefer(
     props.loadableField,
+    // @ts-expect-error
```
tbezman, line 51: TypeScript is too dumb to reconcile the types here. The signature of the function is still correct.

libs/isograph-react/src/react/LoadableFieldReader.tsx
rbalicki2: sgtm

PR #728, opened 2025-10-18
supress-unused-variables-error

PatrykWalach:
I did validate by updating a demo.
Is this part of the demo being used during presentations?

Should I validate params starting with '_' are actually unused?

### DIFF
--- a/demos/pet-demo/src/components/PetDetailDeferredRoute.tsx
+++ b/demos/pet-demo/src/components/PetDetailDeferredRoute.tsx
@@ -15,12 +15,13 @@ import {
 
 export const PetDetailDeferredRouteComponent = iso(`
   field Query.PetDetailDeferredRoute(
-    $id: ID !
+    $id: ID !,
+    $_bar: Int
   ) @component {
     pet(
       id: $id
     ) {
-      PetDetailDeferredRouteInnerComponent
+      PetDetailDeferredRouteInnerComponent(_foo: $_bar)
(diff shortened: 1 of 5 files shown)


rbalicki2:
Hm. I wonder if we should model this as a directive `@allowUnused`, or `@expose` or something. The reason for modeling it as a directive, rather than a naming convention, is that what looks like an unused variable is actually used by the client field, via `parameters`, it `parameters._foo` doesn't scream "you must use `_foo`"

Thinking out loud. Perhaps we only pass the parameters that are actually marked with `@expose`. Then, we can say that every param must be used or marked with `@expose` (or both.)

I see a potential perf benefit to only exposing params marked with `@expose`. Perhaps we can avoid re-rendering if the param changes, but the fragment is unchanged. For example, a param that is only passed to a child `@component` field should not require a re-render of the parent component. Is there any reason we can't get it to work like that?

https://github.com/obi1kenobi/trustfall uses `@output` for exposing fields (not params). I don't particularly like this name, tbh. `@exposeParam`?

PatrykWalach:
Also renaming inside iso literal requires updating all field callsites, it would be nice to be able to alias variables

rbalicki2:
For renames, I imagine the language server can help with local renames, and for global renames, we have some directive that opts parameters into being global. The language server can also help with renames there

PatrykWalach:
Closed in favor of some directive

PR #730, opened 2025-10-19
add TS tests for LoadableFields

PatrykWalach:


libs/isograph-react/src/react/LoadableFieldReader.tsx
```diff
@@ -17,10 +18,7 @@ type MaybeRequiredArgs<
   TReadFromStore extends UnknownTReadFromStore,
   TProvidedArgs extends object,
 > =
-  ArgsWithoutProvidedArgs<TReadFromStore, TProvidedArgs> extends Record<
-    PropertyKey,
-    never
-  >
+  {} extends ArgsWithoutProvidedArgs<TReadFromStore, TProvidedArgs>
```
PatrykWalach, line 21: This works because it's not possible to assign empty object to one with required props but it's possible to assign to object with optional props
```tsx
// @ts-expect-error this doesn't work
let variables: { foo: "bar" } = {}

// but this does
let variables: { foo?: "bar" } = {}
```

rbalicki2 (approved):
hell yeah, thanks for adding this. shipit

Issue #427, opened 2025-02-24
Optimistic store rewrite


rbalicki2:
# Optimistic store rewrite

## Big picture

`startUpdate` (later, `startOptimisticUpdate`?) and optimistic responses will make use of the same machinery: a stack of mergeable updates. `startUpdate` by itself will be substantially simpler than doing the whole thing. But let's make sure that, while working on `startUpdate`, we do stuff in a way that's compatible with the bigger picture.

There can be any number of optimistic responses, and they can be interspersed with network responses. Let's model that as a doubly-linked list of optimistic nodes.

## Example: making optimistic responses

The state of the store might be as follows

```
base
// optimistic update made
base <- opt1
// another optimistic response
base <- opt1 <- opt2
// opt1 cleared
base <- opt2
// network response received
base <- opt2 <- NR
// network response received, and is merged into the previously received network response
base <- opt2 <- NR
// opt2 is cleared, network response is merged into parent
base
```

## Modeling the stack

Something like

```ts
type NetworkResponseNode = {
  kind: 'NetworkResponseNode';
  childNode: OptimisticNode | null;
  parent: OptimisticNode | null;
  data: DataLayer;
};
type OptimisticNode = {
  kind: 'OptimisticNode';
  childNode: OptimisticNode | NetworkResponseNode | null;
  parentNode: OptimisticNode | NetworkResponseNode | BaseNode;
  data: DataLayer;
};
```

- So, we are ensuring that: a `NetworkResponseNode` must be the parent
- No two `NetworkResponseNode`s can be adjacent

We could refine this to ensure that the base node's data has `Query: { __ROOT: {} }`. I don't know if it's worth it.

## The Isograph environment

The environment only needs a pointer the bottom-most node in the stack.

## Reading

- Reading is always done from the bottom of the stack. If the scalar or linked field is present in the current item and is not `undefined`, then use that value, otherwise, try again in the next item in the stack

## Proxies

- `startUpdate` (and optimistic updaters) are passed a function with signature `(proxy: TUpdatableData) => void`. Call this function an `updater`
- This `proxy` (call it a `StoreProxy`) starts with an empty `DataLayer`.
- Writes will mutate that `DataLayer`.
  - If we write `proxy.user.name = "John"` and in the stack, that user already has the name `"John"`, **then we still write `user.name = "John` to the `DataLayer`**.
[...]

PR #734, opened 2025-10-22
optimistic layer

PatrykWalach:
#427

### DIFF
--- a/libs/isograph-react/src/core/optimisticProxy.ts
+++ b/libs/isograph-react/src/core/optimisticProxy.ts
@@ -47,19 +47,11 @@ export function getMutableStoreRecordProxy(
   childMostStoreLayer: StoreLayer,
   link: StoreLink,
 ): StoreRecord {
-  // Note: this is lazily initialized, to optimize for the (common?) case where
-  // the first thing we do is assign a value.
-  let firstStoreLayerWithRecord: StoreLayer | undefined | null = undefined;
-
   return new Proxy<StoreRecord>(
     {},
     {
       get(_, propertyName) {
-        let currentStoreLayer =
-          firstStoreLayerWithRecord === undefined
-            ? findFirstStoreLayerWithRecord(childMostStoreLayer, link)
-            : firstStoreLayerWithRecord;
-
+        let currentStoreLayer: StoreLayer | null = childMostStoreLayer;
         while (currentStoreLayer !== null) {
           const storeRecord =
             currentStoreLayer.data[link.__typename]?.[link.__link];
@@ -76,11 +68,7 @@ export function getMutableStoreRecordProxy(
         }
       },
       has(_, propertyName) {
-        let currentStoreLayer =
-          firstStoreLayerWithRecord === undefined
-            ? findFirstStoreLayerWithRecord(childMostStoreLayer, link)
-            : firstStoreLayerWithRecord;
-
+        let currentStoreLayer: StoreLayer | null = childMostStoreLayer;
         while (currentStoreLayer !== null) {
           const storeRecord =
             currentStoreLayer.data[link.__typename]?.[link.__link];
@@ -98,10 +86,6 @@ export function getMutableStoreRecordProxy(
         return false;
       },
       set(_, p, newValue) {
-        // Because we are setting a value on the childMost layer, that becomes
-        // the first layer with a record corresponding to link.
-        firstStoreLayerWithRecord = childMostStoreLayer;
-
         return Reflect.set(
           getOrInsertRecord(childMostStoreLayer.data, link),
           p,
@@ -523,22 +507,4 @@ function compareData(
       encounteredIds.get(typeName)?.delete(id);
     }
   }
-}
-
-/**
- * Starting from the childMost layer, search in the parent-wise direction,
- * until we find a layer containing a record corresponding to link.
- */
-function findFirstStoreLayerWithRecord(
-  childMostStoreLayer: StoreLayer | null,
-  link: StoreLink,
-): StoreLayer | null {
-  let currentStoreLayer: StoreLayer | null = childMostStoreLayer;
-  while (
-    currentStoreLayer != null &&
-    currentStoreLayer.data[link.__typename]?.[link.__link] == null
-  ) {
-    currentStoreLayer = currentStoreLayer.parentStoreLayer;
-  }
-  return currentStoreLayer;
 }


libs/isograph-react/src/core/garbageCollection.ts
```diff
@@ -45,22 +46,27 @@ export function retainQuery(
 
 export function garbageCollectEnvironment(environment: IsographEnvironment) {
   const retainedIds: RetainedIds = {};
+  let node: OptimisticLayer = environment.store;
```
PatrykWalach: garbage collect from `BaseNode` only, not sure if that's ok

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -54,37 +60,45 @@ export function createStartUpdate<TReadFromStore extends UnknownTReadFromStore>(
...
+          kind: 'StartUpdateError',
+          error: e,
+        }));
+        throw e;
+      } finally {
+        logMessage(environment, () => ({
+          kind: 'StartUpdateComplete',
+          updatedIds: mutableUpdatedIds,
+        }));
+        callSubscriptions(environment, mutableUpdatedIds);
```
PatrykWalach: this needs to be moved, `callSubscriptions` will be called multiple times, during `replaceOptimisticNodeWithNetworkResponseNode`

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,269 @@
...
+    optimisticNode.childNode.kind === 'NetworkResponseNode'
+  ) {
+    mergeDataLayer(optimisticNode.childNode.data, data);
+
+    makeRootNode(environment, optimisticNode.childNode);
+    reexecuteUpdates(environment, optimisticNode.parentNode);
+  } else if (optimisticNode.parentNode?.kind === 'NetworkResponseNode') {
+    const networkResponseNode = optimisticNode.parentNode;
+    mergeDataLayer(data, networkResponseNode.data);
+    networkResponseNode.data = data;
```
PatrykWalach: this is probably wrong

base <- opt1 <- NR

we apply in other

base <- NR <- opt1

rbalicki2: oh you're right, great catch. If we have `opt1` that hasn't been disposed, then we should preserve the order: `base <- opt1 <- NR`, not `base <- NR <- opt1`. If `opt1` is disposed, then we should combine the stack at that time to be just `base`

PatrykWalach: this is actually correct, opt1 data is overridden by NR

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: how do we track encountered ids? 
we can convert `DataLayer` to `EncounteredIds` but then we don't take into account if data is different

PatrykWalach: If we have some operations we want to compare with the previous version, instead of comparing with a different operation, if the optimistic mutation is accurate we shouldn't have to rerender

```
base <- opt1 <- opt2 <- NR
                 ^      ^
base <- opt1 <-  NR  <- NR
```

PatrykWalach: So this works for optimistic updates, for adding new nodes we use the old algorithm of comparing with the rest of the tree

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -163,7 +173,7 @@ function readUpdatableData<TReadFromStore extends UnknownTReadFromStore>(
   mutableState: MutableInvalidationState,
   mutableUpdatedIds: EncounteredIds,
 ): ReadDataResultSuccess<ExtractUpdatableData<TReadFromStore>> {
-  let storeRecord = environment.store[root.__typename]?.[root.__link];
+  let storeRecord = readOptimisticRecord(environment, root);
```
PatrykWalach: This doesn't include this new `dataLayer`

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: this is implemented now

rbalicki2: nice! will review when I have a chance (probably no sooner than Friday night)

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,352 @@
...
+    }
+  }
+}
+
+function reexecuteUpdates(
+  environment: IsographEnvironment,
+  node: OptimisticLayer | null,
+  mutableEncounteredIds: EncounteredIds,
+) {
+  while (node && node?.kind !== 'OptimisticNode') {
```
PatrykWalach: The loops should be merged only if optimistic node is in front of base `base <- opt1` 
in case it's in the middle we cannot merge, we have to keep `startUpdates`

`base < opt1 < opt2 <- startUpdate`
`base < opt1 < NR <- startUpdate`

rbalicki2: Not sure I follow, but basically there should be a linked list as follows

```
type PersistedStackItem =
  { kind: 'Persisted', parent: null | OptimisticStackItem } // startUpdate, NR, base
type OptimisticStackItem =
  { kind: 'Optimistic', parent: PersistedStackItem | OptimisticStackItem } // optimistic response, revertibleStartUpdate
```

Basically, this means that we cannot ever have `base <- NR`, and so if we go from `base <- opt1` and that becomes `base <- NR` we are forced to merge

PatrykWalach: What you are talking about is guaranteed by the types, this loop is supposed to merge everything into base if until we encounter another optimistic update

`base <- opt1 < NR <- opt2`

`base  <- opt2`

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+
+export function getOrInsertRecord(dataLayer: DataLayer, link: StoreLink) {
+  const recordsById = (dataLayer[link.__typename] ??= {});
+  return (recordsById[link.__link] ??= {});
+}
+
+export function readOptimisticRecord(
+  environment: IsographEnvironment,
+  link: StoreLink,
+) {
```
rbalicki2: can we provide return types for exported functions?

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+  return new Proxy<StoreRecord>(
+    {},
+    {
+      get(_, p) {
+        let node: OptimisticLayer | null = environment.store;
+
+        while (node !== null) {
+          const storeRecord = node.data[link.__typename]?.[link.__link];
+          if (storeRecord != undefined) {
+            const value = Reflect.get(storeRecord, p, storeRecord);
```
rbalicki2: TIL reflect

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+) {
+  return new Proxy<StoreRecord>(
+    {},
+    {
+      get(_, p) {
+        let node: OptimisticLayer | null = environment.store;
+
+        while (node !== null) {
+          const storeRecord = node.data[link.__typename]?.[link.__link];
+          if (storeRecord != undefined) {
```
rbalicki2: this should be `!== null`

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+type BaseNode = {
+  readonly kind: 'BaseNode';
+  parentNode: OptimisticNode | null;
+  childNode: null;
+  readonly data: IsographStore;
+};
+
+type NetworkResponseNode = {
+  readonly kind: 'NetworkResponseNode';
+  parentNode: OptimisticNode | StartUpdateNode | null;
```
rbalicki2: why do we need this to be a doubly linked list? Hm. It's because we need to be able to remove optimistic updates and remove them or replace them network responses.

I think we only need to include references to parents if the node is an `optimisticNode`, right? As in, it's only used when it is removed/replaced, and we'll never do that for NR/StartUpdate

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+
+export function readOptimisticRecord(
+  environment: IsographEnvironment,
+  link: StoreLink,
+) {
+  return new Proxy<StoreRecord>(
+    {},
+    {
+      get(_, p) {
+        let node: OptimisticLayer | null = environment.store;
```
rbalicki2: I can't think of a great name for this, but IMO `.store` is a bit confusing now, as it's actually a pointer to the tip of the store.

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+        }
+        return false;
+      },
+    },
+  );
+}
+
+type BaseNode = {
+  readonly kind: 'BaseNode';
+  parentNode: OptimisticNode | null;
```
rbalicki2: IMO the parentNode and childNode fields are reversed, relative to what I am imagining. As in, I imagine that the `Base` is the "most parent" Node. In my other comments, I try to use your terminology.

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+  }));
+
+  callSubscriptions(environment, encounteredIds);
+}
+
+function mergeDataLayer(target: DataLayer, source: DataLayer): void {
+  for (const typeName in source) {
+    target[typeName] ??= {};
+    for (const id in source[typeName]) {
+      target[typeName][id] ??= {};
```
rbalicki2:
```js
const target = target[typeName][id] ??= {};
Object.assign(target, ...)
```

this can also be simplified IMO with the use of Object.entries

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+    makeRootNode(environment, networkResponseNode);
+    reexecuteUpdates(environment, optimisticNode.parentNode, encounteredIds);
+  }
+}
+
+export type OptimisticLayer =
+  | OptimisticNode
+  | NetworkResponseNode
+  | StartUpdateNode
+  | BaseNode;
```
rbalicki2: This is more complicated than I expected! I didn't expect us to need four layer types. But that is in fact the correct way to model this!

rbalicki2 (commented):
Partial review. I'll replace `parent` and `child` and continue the review... it's confusing for me to think about like this.

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -125,14 +126,17 @@ export type DataId = string;
 
 export const ROOT_ID: DataId & '__ROOT' = '__ROOT';
 
-export type IsographStore = {
+export type DataLayer = {
```
rbalicki2: Let's clean up these names. I don't think we should use `Layer` for both `DataLayer` and `OptimisticLayer`. `OptimisticLayer` is a wrapper around `DataLayer`.

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: I called this the other way around than you, because it makes sense if `.store` is the `root` of the tree

Also this is a linked list so `next` and `prev` might make more sense, and `.store` is the `head`

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -125,14 +126,17 @@ export type DataId = string;
...
   [index: TypeName]: {
     [index: DataId]: StoreRecord | null;
   } | null;
+};
+
+export interface IsographStore extends DataLayer {
```
rbalicki2: `BaseNodeData` and `NodeData` (above)? It's confusing that `environment.store` is not a `Store`, and `node.data` is a `Store`

We could rename this when we export it, if `BaseNodeData` is too technical. But IMO it's fine.

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+};
+
+type FirstUpdate = () => WithEncounteredIds<DataLayer>;
+type DataUpdate = () => Pick<WithEncounteredIds<DataLayer>, 'data'>;
+
+type StartUpdateNode = {
+  readonly kind: 'StartUpdateNode';
+  childNode: OptimisticNode | NetworkResponseNode | null;
+  parentNode: OptimisticNode | NetworkResponseNode;
+  data: DataLayer;
```
rbalicki2: Hm... so we have a `DataLayer` because that's the result of calling `.startUpdate`. Can we add a comment to indicate that? Or rename the field somehow?

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+  }
+  return node;
+}
+
+function reexecuteUpdates(
+  environment: IsographEnvironment,
+  node: OptimisticNode | NetworkResponseNode | StartUpdateNode | null,
+  mutableEncounteredIds: EncounteredIds,
+) {
+  while (node !== null) {
```
rbalicki2: Do we need to reassign `environment.store` here? Can we just go from `env.store -> env.store.parent` etc until we encounter null?

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,363 @@
...
+
+function compareData(
+  oldData: DataLayer,
+  newData: DataLayer,
+  encounteredIds: EncounteredIds,
+) {
+  if (oldData === newData) {
+    for (const [typeName, ids] of encounteredIds.entries()) {
+      for (const id of ids) {
+        encounteredIds.get(typeName)?.delete(id);
```
rbalicki2: If we have `StartUpdate <- NR` and they both contain `(User, ID_1)`, then `StartUpdate` can change `user.name` and `NR` not affect that, and we will remove `ID_1` from the encounteredIds, which seems incorrect.

rbalicki2 (changes requested):
okay that's enough reviewing for now. Amazing work!!! This is dope. This is going to be so clean, and will be a good foundation for:
- deferred merging
- [multi version concurrently control](https://en.wikipedia.org/wiki/Multiversion_concurrency_control) i.e. support for transitions
- optimistic start update
- making optimistic updates permanent

I renamed everything as needed, but maybe there are a few places I missed things. But I like the naming now.

LMK when it's ready for re-review

libs/isograph-react/src/core/optimisticProxy.ts
rbalicki2: Ah sorry, I went ahead and renamed it 😂  to ease my ability to review. Let's keep it as it is now. The mental model I have is that `environment.store: StoreLayer` is the outermost layer, and then you go inward (which is towards the parent, but maybe it would be clearer to call that `inner`.)

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: yes, we need to pop all the nodes behind the optimistic update that's being replaced,
then append them one by one, otherwise we would be reading invalid data

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: That's good catch, so we need to merge startUpdate node + all nodes after into single dataLayer, then compare them

PatrykWalach:
```
base <- opt1 <- opt2 <- NR < startUpdate
              |       merge these       |
```

PatrykWalach: should be working now

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: done

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,400 @@
...
+  | OptimisticStoreLayer
+  | NetworkResponseStoreLayer
+  | StartUpdateStoreLayer
+  | BaseStoreLayer;
+
+function compareData(
+  oldData: StoreLayerData,
+  newData: StoreLayerData,
+  encounteredIds: EncounteredIds,
+): void {
```
PatrykWalach, line 476: If in optimistic updater we update field unrelated to network response, then in the response it's not present so we have to revert this field. In this case we have to add the id of parent object to `EncounteredIds`.

PatrykWalach: done

PatrykWalach: I assume the same can happen with the whole record, but in that case, we don't want to call the subscriptions

libs/isograph-react/src/core/optimisticProxy.ts
rbalicki2: sorry I'm an idiot, it should be `!== undefined`, since null means we deleted it, and that's allowed. And you do that below. So LGTM.

Okay, so I think right now we don't expose the ability to delete a record, just to set fields to null. We should add that ability, but right now, if you e.g. receive `{ node: null }` as a network response, we'll set the field to null, but don't have some way to find the record and remove it.

We should probably add that. But LGTM as of the current state of the codebase!

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,415 @@
...
+          }
+          node = node.parentStoreLayer;
+        }
+      },
+      has(_, p) {
+        let node: StoreLayer | null = environment.store;
+
+        while (node !== null) {
+          const storeRecord = node.data[link.__typename]?.[link.__link];
+          if (storeRecord != undefined) {
```
rbalicki2: `!= undefined` should be linted against, tbh, `!=` should only (I think) be used with `null`

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,412 @@
...
+      set.add(id);
+    }
+  }
+
+  for (const [typeName, newRecords] of Object.entries(newData)) {
+    if (!newRecords) {
+      continue;
+    }
+    const oldRecords = oldData[typeName];
+    outer: for (const [id, newRecord] of Object.entries(newRecords)) {
```
rbalicki2, line 492: nit: `newRecords:`

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,412 @@
...
+
+type FirstUpdate = () => WithEncounteredIds<StoreLayerData>;
+type DataUpdate = () => Pick<WithEncounteredIds<StoreLayerData>, 'data'>;
+
+type StartUpdateStoreLayer = {
+  readonly kind: 'StartUpdateStoreLayer';
+  childStoreLayer: OptimisticStoreLayer | NetworkResponseStoreLayer | null;
+  parentStoreLayer: OptimisticStoreLayer | NetworkResponseStoreLayer;
+  data: StoreLayerData;
+  startUpdate: DataUpdate;
```
rbalicki2: let's call this `executeStartUpdate` or something for clarity

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,405 @@
...
+    | StartUpdateStoreLayer
+    | NetworkResponseStoreLayer
+    | BaseStoreLayer;
+  data: StoreLayerData;
+  startUpdate: DataUpdate;
+};
+
+export function addNetworkResponseStoreLayer(
+  environment: IsographEnvironment,
+  normalizeData: FirstUpdate,
```
PatrykWalach: This taking a function doesn't seem right, because we cannot track which fields changed

libs/isograph-react/src/core/startUpdate.ts
PatrykWalach: with the approach of reading from a given node, this is fixed

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: managed to make it work by resetting optimistic and appending a new node

rbalicki2:
I'm overthinking things, with respect to the doubly linked list. We can just iterate to the final item, creating a call stack (or realized stack), then recalculate, and go backwards until we reach the outer item. Regardless, it's an O(n^2) operation, n for iterating the list. Then each layer has to iterate the list as well. I think N will be relatively small most of the time, so it's not a big deal, and they added simplicity of it being a singly linked list is worth it

I'll sketch this out when I'm at a computer too

--------

So, `base <- opt1 <- SU1 <- opt2 <- SU2`, if we're detaching/replacing `opt1`, we:
- starting at `SU2`, we recurse until we reach `opt1`, keeping track of the parent layer as a parameter, detach/replace it (we can do this without keeping parent pointers, even on `opt1`).
- if we replaced, then we have `base <- NR <- SU1 <- ...`, and the callstack is at `NR`
- at this point, we go backwards and merge/update. We can merge `NR` by looking at its parent, merge `SU` because its parent is `base`/`NR`, leave `opt2` unchanged, then recalculate `SU2` because its parent is an `opt`

This has O(n^2) perf, same as the doubly linked list, though I suppose it might use some extra memory because it constructs the callstack. But again, low N.

rbalicki2 (commented):
oops, sorry, forgot to submit this review

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,407 @@
...
+): void {
+  let encounteredIds: EncounteredIds;
+  switch (environment.store.kind) {
+    case 'NetworkResponseStoreLayer':
+    case 'BaseStoreLayer': {
+      encounteredIds = normalizeData(environment.store);
```
rbalicki2: This function could be reusable if we didn't do the `environment.store =` stuff. This function does the following:
- given a parent and a network response
- merge the NR with the parent if the parent is a network response layer/base layer
- otherwise, set the network response's parent
- return either NR or the parent

The caller then does something with the new parent, i.e. `environment.store = newParent`.

When we replace `opt1` with `NR` in `... <- opt1 <- SU`, we also do that, except the caller does something with `newParent`, which is either `SU.parent = newParent` (if `newParent` is `opt` or `NR`), or `const nextNewParent = addStartUpdateNode(SU, newParentBase)`, if parent is `base`.

----

I think the same thing is true for `addStartUpdateNode`. This will either merge `SU` and return the parent, or return `SU`, and we can assign/merge that guy.

LMK if that doesn't make sense. I have some inkling that we can make this extremely clean!

PatrykWalach: do you think we should use a sentinel node?

rbalicki2: I'm not sure we need one, I think:
- a given layer should only show up once, so removing by identity is sufficient, and
- the "remove/replace/add and recalculate" method should do everything, i.e. we should just do something like

`environment.store = remove(environment.store, itemToRemove)` or `= removeAndReplace(environment.store, itemToRemove, itemToReplace)`

PatrykWalach: When you get some time to re-review, let me know if what I did in the last commit is what you've wanted

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,386 @@
...
+};
+
+export function addNetworkResponseStoreLayer(
+  parent: StoreLayer,
+  normalizeData: DataUpdate,
+): StoreLayer {
+  switch (parent.kind) {
+    case 'NetworkResponseStoreLayer':
+    case 'BaseStoreLayer': {
+      normalizeData(parent);
```
rbalicki2: this function is so clean!

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,386 @@
...
+    case 'BaseStoreLayer': {
+      startUpdate(parent);
+      return parent;
+    }
+    case 'StartUpdateStoreLayer': {
+      const node = parent;
+
+      const prevStartUpdate = node.startUpdate;
+      node.startUpdate = () => {
+        prevStartUpdate(node);
```
rbalicki2, line 200: yesss I love that we're merging these

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,386 @@
...
+    for (const id in source[typeName]) {
+      target[typeName][id] ??= {};
+      Object.assign(target[typeName][id], source[typeName][id]);
+    }
+  }
+}
+
+export function addStartUpdateStoreLayer(
+  parent: StoreLayer,
+  startUpdate: DataUpdate,
```
rbalicki2: is `startUpdate(optimisticLayer)` invalid? Can we restrict the types here to not accept any layer type, but instead, accept just a base store layer/start update store layer?

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,386 @@
...
+export function addOptimisticStoreLayer(
+  parent: StoreLayer,
+  startUpdate: DataUpdate,
+) {
+  switch (parent.kind) {
+    case 'BaseStoreLayer':
+    case 'StartUpdateStoreLayer':
+    case 'NetworkResponseStoreLayer':
+    case 'OptimisticStoreLayer': {
+      const data = {};
```
rbalicki2: nit: no need to make `data` a separate variable

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,386 @@
...
+  }
+}
+
+function replaceOptimisticStoreLayerWithNetworkResponseStoreLayer(
+  environment: IsographEnvironment,
+  optimisticNode: OptimisticStoreLayer,
+  normalizeData: (storeLayer: StoreLayer) => void,
+): void {
+  const oldData = optimisticNode.data;
+  //  we cannot replace optimistic node with network response directly
```
rbalicki2: nit: spacing

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,386 @@
...
+    mergeDataLayer(newData, node.data);
+
+    node = node.childStoreLayer;
+  }
+}
+
+function replaceOptimisticStoreLayerWithNetworkResponseStoreLayer(
+  environment: IsographEnvironment,
+  optimisticNode: OptimisticStoreLayer,
+  normalizeData: (storeLayer: StoreLayer) => void,
```
rbalicki2: Can `normalizeData` here accept `StoreLayerData` instead of `StoreLayer`?

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,386 @@
...
+): void {
+  const oldData = optimisticNode.data;
+  //  we cannot replace optimistic node with network response directly
+  // because of the types so we have to:
+  //  1. reset the optimistic node
+  optimisticNode.data = {};
+  // 2. append the network response as child
```
rbalicki2: Can we:

- store the optimistic layer's parent layer P1 and child layer C1
- merge the network response into the parent, getting a new parent P2 (which has type `BaseStoreLayer | NetworkResponseStoreLayer`)
- set the new parent P2's child to be C1... But this won't typecheck, so we need to actually do `p2.child = mergeChildNodes(c1, p2)`
- re-execute updates, but starting from `p2.child`.

That seems a bit simpler than what is going on here. I haven't really read this in a huge amount of detail yet, though. So maybe we're doing the same thing.

PatrykWalach: > merge the network response into the parent, getting a new parent P2 (which has type BaseStoreLayer | NetworkResponseStoreLayer)

so we can only replace optimistic update if the parent is a `BaseStoreLayer | NetworkResponseStoreLayer`?

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: we need the `StoreLayer` to read through all the parent layers, while we `normalizeData` we compare new data with old  data to track changed records ids

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: done

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: done

libs/isograph-react/src/core/optimisticProxy.ts
rbalicki2: Sorry, I meant call the function that adds and merges the network data. So if the parent is a base node/network node, it gets merged, but otherwise gets added

PatrykWalach: The downside to this is right now I can compare `opt1.data` with `NR.data`, but with this I will apply data to `base` without an extra node and will have to compare `base.data+opt1.data` with `base.data+NR.data`

libs/isograph-react/src/core/optimisticProxy.ts
rbalicki2: Oops, following up here.

If we have `{ User: { 1: { name: "foo" } }` in base, and `{ User: { 1: null } }` in a start update layer (we set `data.user = null`), then this will continue to read through the null layer to the base layer.

So I think

- this should return `StoreRecord | null` and `new Proxy<StoreRecord | null>`???
- we should return `null` from `get` here

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,403 @@
...
+      startUpdate(node);
+      parent.childStoreLayer = node;
+
+      return {
+        node,
+        revert: (
+          environment: IsographEnvironment,
+          normalizeData: (storeLayer: StoreLayer) => EncounteredIds,
+        ): void => {
+          replaceOptimisticStoreLayerWithNetworkResponseStoreLayer(
```
rbalicki2: Let me know if I'm misunderstanding. I think that we're going to call revert in two circumstances:
- when we revert and normalize the non-optimistic network response
- when we revert and normalize nothing (e.g. if the network response errored out)

I think in the latter case, we will create an empty network layer for no reason. So, perhaps we should:
- make `normalizeData` nullable (or something), and
- only create that layer if `normalizeData != null`? (Or solve this some other way.)

Thoughts?

rbalicki2 (commented):
more review
edit: lol I didn't need to submit this review, my bad

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: We have optimistcUpdate but no optimistcResponse is that alright?

PatrykWalach: I guess optimisticResponse can be inital data of OptimistcUpdateNode

rbalicki2: I'm not sure I follow

PatrykWalach: Relay has two different props that can be used: optimisticResponse that accepts a raw response object and an optimistic updater where you can make updates with proxy

rbalicki2: yeah, at least for now, lets support optimistic updates but not optimistic responses. To support optimistic responses, we at least need a raw response type. We can punt on that for now

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,403 @@
...
+): StoreRecord {
+  return new Proxy<StoreRecord>(
+    {},
+    {
+      get(_, p) {
+        let node: StoreLayer | null = storeLayer;
+
+        while (node !== null) {
+          const storeRecord = node.data[link.__typename]?.[link.__link];
+          if (storeRecord != undefined) {
```
rbalicki2: `!== null`

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: done

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -166,13 +167,30 @@ export function makeNetworkRequest<
...
       const root = { __link: ROOT_ID, __typename: artifact.concreteType };
       if (status.kind === 'UndisposedIncomplete') {
-        normalizeData(
-          environment,
-          normalizationAst.selections,
-          networkResponse.data ?? {},
-          variables,
-          root,
+        const encounteredIds: EncounteredIds = new Map();
+        environment.store = addNetworkResponseStoreLayer(
```
rbalicki2: Since we're returning `storeLayer` (line 173), it would be simpler to do:

```
// returns either environment.store or a new node that points to it
const newChildStoreLayer = ensureChildStoreLayerAcceptsData(environment.store);
normalizeData(newChildStoreLayer);
environment.store = newChildStoreLayer;
```

Sorry, I know I suggested what you have here!

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,411 @@
...
+            const value = Reflect.has(storeRecord, p);
+            if (value) {
+              return true;
+            }
+          }
+          node = node.parentStoreLayer;
+        }
+        return false;
+      },
+      set(_, p, newValue) {
```
rbalicki2, line 88: Maybe we should have two functions: `getMutableStoreRecordProxy` and `getStoreRecordProxy`, and call `getStoreRecordProxy` from `read` and `getMutable` from `cache`?

Because right now, the setter is defined but not valid in `read`, since the childmost store layer could be an optimistic layer or a startUpdate layer.

libs/isograph-react/src/core/startUpdate.ts
```diff
@@ -163,7 +183,7 @@ function readUpdatableData<TReadFromStore extends UnknownTReadFromStore>(
   mutableState: MutableInvalidationState,
   mutableUpdatedIds: EncounteredIds,
 ): ReadDataResultSuccess<ExtractUpdatableData<TReadFromStore>> {
-  let storeRecord = environment.store[root.__typename]?.[root.__link];
+  let storeRecord = getStoreRecordProxy(storeLayer, root);
   if (storeRecord == null) {
```
rbalicki2: this can never be null, now

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,416 @@
...
+  link: StoreLink,
+): StoreRecord {
+  return new Proxy<StoreRecord>(
+    {},
+    {
+      get(_, p) {
+        let node: StoreLayer | null = storeLayer;
+
+        while (node !== null) {
+          const storeRecord = node.data[link.__typename]?.[link.__link];
```
rbalicki2: Right now, every time we access a field, we re-iterate the layers. Let's keep track of the first storeLayer that contained the store record, as in:

```
let targetStoreLayer: StoreLayer | undefined | null = undefined;

get(_, p) {
  if (targetStoreLayer === undefined) { ... }
  // we still have to do the while loop again
}
set(_,p) {
  targetStoreLayer = storeLayer; // if we set, the record gets added to the childmost layer
}
```

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,424 @@
...
+ * with a network response.
+ *
+ * After we do this, we must re-execute all child startUpdate and optimistic
+ * layers (since their data may have changed.) We also keep track of changed
+ * records, in order to call affected subscriptions.
+ */
+export function revertOptimisticStoreLayerAndMaybeReplace(
+  environment: IsographEnvironment,
+  optimisticNode: OptimisticStoreLayer,
+  normalizeData: null | ((storeLayer: StoreLayerWithData) => void),
```
rbalicki2, line 365: Can this just be `data | null`?

rbalicki2 (commented):
more review

rbalicki2:
okay, finally, I feel good about this. Let's make the final changes above ☝️  merge main and we should be g2g.

I *really* like how `revertOptimisticStoreLayerAndMaybeReplace` turned out with respect to the types forcing us to the do right thing. e.g. `mergeLayersWithDataIntoBaseLayer` is so cool

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: Yeah

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: done

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: but `normalizeData()` accepts a node so I then we have to create a new BaseNode before reverting

```tsx
const node = { kind: "BaseNode", ... };
normalizeData(node, ...);
revertOptimisticStoreLayerAndMaybeReplace(..., node.data);
```

rbalicki2: SGTM.

Let's tighten this up as a follow up. I think `normalizeData` should not require a layer, since it only uses `storeLayer.data` in:

```
return Reflect.set(
          getOrInsertRecord(storeLayer.data, link),
          p,
          newValue,
        );
```
and we will always `getOrCreate`, i.e. never actually reach for a parent layer.

rbalicki2: I think that `normalizeData` is two actually separate functions:
- normalize network response, and
- normalize startUpdate

The former doesn't require a layer, the latter does

PatrykWalach: it uses layer to compare the old data with the new one to track changed ids

rbalicki2: Ah I see, sgtm

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: done

This has a caveat, if we have two proxies to the same record, setting on one proxy will break the other.
Because of this in startUpdate I recreate a proxy in each getter - the getters in start update have caching so this should be fine

libs/isograph-react/src/core/optimisticProxy.ts
PatrykWalach: I don't understand what's the issue here, we don't use setters when reading

libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -0,0 +1,506 @@
...
+      continue;
+    }
+    const targetRecordById = (target[typeName] ??= {});
+    for (const [id, sourceRecord] of Object.entries(sourceById)) {
+      if (sourceRecord === null) {
+        targetRecordById[id] = null;
+        continue;
+      }
+      const targetRecord = (targetRecordById[id] ??= {});
+      Object.assign(targetRecord, sourceRecord);
```
PatrykWalach, line 181: if we do `{ ...targetRecord, ...sourceRecord }`, then we can check for `===` in `compareData` to see if the record is unchanged, not sure if that's faster

PatrykWalach: We also could track unchanged as in separate object

PR #747, opened 2025-11-07
do not suspend in useResult for components

### LINKED ISSUE #746: `useResult` suspends when returnig components
Should it not suspend, but return components that suspend when rendered?

prerequisite:
- `ReaderArtifact` contains field name needed to store the component, it might need to be moved.

PatrykWalach:
Closes #746

---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #747 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #794

### DIFF
--- a/libs/isograph-react/src/core/IsographEnvironment.ts
+++ b/libs/isograph-react/src/core/IsographEnvironment.ts
@@ -231,7 +231,11 @@ export function getOrLoadReaderWithRefetchQueries(
   readerWithRefetchQueries:
     | ReaderWithRefetchQueries<any, any>
     | ReaderWithRefetchQueriesLoader<any, any>,
-) {
+): {
+  readerWithRefetchQueries: PromiseWrapper<ReaderWithRefetchQueries<any, any>>;
+  fieldName: string;
+  readerArtifactKind: 'EagerReaderArtifact' | 'ComponentReaderArtifact';
+} {
   switch (readerWithRefetchQueries.kind) {
     case 'ReaderWithRefetchQueries':
       return {


crates/artifact_content/src/eager_reader_artifact.rs
```diff
@@ -103,7 +99,7 @@ pub(crate) fn generate_eager_reader_artifacts<TNetworkProtocol: NetworkProtocol
             {}{reader_output_type}\n\
             > = {{\n\
             {}kind: \"EagerReaderArtifact\",\n\
-            {}fieldName: \"{eager_reader_name}\",\n\
+            {}typeAndField: \"{type_and_field}\",\n\
```
PatrykWalach: Actually I probably should do it the other way around - replace the `typeAndField` with `fieldName` everywhere.
We used to need the type when the cache was single layer, but now, each `FragmentReference` has `root.__typename` which is less prone to collisions, because in `typeAndField` the type could be abstract.
Keeping `typeAndField` just makes stableIdForFragment longer, for no reason.

demos/pet-demo/src/components/__isograph/BlogItem/BlogItemDisplay/resolver_reader.ts
```diff
@@ -41,7 +41,7 @@ const readerAst: ReaderAst<BlogItem__BlogItemDisplay__param> = [
     ],
     entrypoint: {
       kind: "EntrypointLoader",
-      typeAndField: "BlogItem__BlogItemMoreDetail",
```
PatrykWalach, line 44: I guess here we want the abstract type and the concrete: 
- abstract to cache `.loader()`
- concrete for the fragment

libs/isograph-react/src/react/useResult.ts
```diff
@@ -28,25 +28,23 @@ export function useResult<
...
 
-  maybeUnwrapNetworkRequest(
-    fragmentReference.networkRequest,
-    networkRequestOptions,
-  );
-  const readerWithRefetchQueries = readPromise(
-    fragmentReference.readerWithRefetchQueries,
-  );
-
   switch (readerWithRefetchQueries.readerArtifact.kind) {
```
PatrykWalach: I guess I've missed an important fact - the kind of resolver needs to be known upfront

libs/isograph-react/src/core/read.ts
```diff
@@ -376,7 +381,7 @@ export function readLoadablySelectedFieldData(
           } else {
             const isographArtifactPromiseWrapper = getOrLoadIsographArtifact(
               environment,
-              field.entrypoint.typeAndField,
+              `${root.__typename}/${field.entrypoint.fieldName}`,
```
PatrykWalach: TODO

libs/isograph-react/src/tests/__isograph/Query/normalizeUndefinedField/resolver_reader.ts
```diff
@@ -29,7 +29,7 @@ const artifact: EagerReaderArtifact<
   Query__normalizeUndefinedField__output_type
 > = {
   kind: "EagerReaderArtifact",
-  fieldName: "Query.normalizeUndefinedField",
+  fieldName: "normalizeUndefinedField",
```
PatrykWalach, line 32: This should be part of last PR, I think this is not part of build-demos so CI didn't caught it

rbalicki2: weird. okay will push change to main

rbalicki2: thanks for splitting this into two PRs btw

rbalicki2 (approved):

PR #751, opened 2025-11-13
allow to store references

### LINKED ISSUE #740: Pico return references
- Memoized functions can return references
- Memoized functions do not accept references as parameters, only DB can be a reference
    - we should always pass small, copiable objects (like names or IDs), not large objects that we worry about cloning

## unchanged stuff:

- params should remain hash / WithSerializable
- GC should work as is? As in, we GC from the root query and (lower priority) GC the related derived nodes, so the last thing to be GC'd will be the "underlying truth", i.e. in `get_first_item_in_selection_set -> get_selection_set -> get_entity -> parse_graphql_schema`, we will GC from the left side, and all references point "to the right"

ch1ffa:
Closes #740 (probably)


crates/pico/src/memo_ref.rs
```diff
@@ -21,48 +27,100 @@ impl<T> Clone for MemoRef<T> {
...
         let (value, revision) = storage
             .get_derived_node_value_and_revision(self.derived_node_id)
-            .unwrap();
+            .expect("Derived node not found. This is indicative of a bug in Pico.");
         storage.register_dependency_in_parent_memoized_fn(
             NodeKind::Derived(self.derived_node_id),
             revision.time_updated,
         );
-        value.downcast_ref::<T>().unwrap()
+        match self.kind {
```
rbalicki2, line 72: should this call 
```
storage.register_dependency_in_parent_memoized_fn(
            NodeKind::Derived(self.derived_node_id),
            revision.time_updated,
        );
self.lookup(...)
```

instead of re-implementing everything?

crates/pico/src/database.rs
```diff
@@ -359,30 +359,49 @@ impl<Db: Database> Default for Storage<Db> {
...
+    intern_from_param(db, param_id, MemoRefKind::Value)
 }
 
 pub fn intern_ref<Db: Database, T: Clone + Hash + DynEq + 'static>(
     db: &Db,
     value: &T,
 ) -> MemoRef<T> {
-    let param_id = intern_borrowed_param(db, value);
-    intern_from_param(db, param_id)
+    let param_id = hash(value).into();
```
rbalicki2, line 369: should we be hashing the `RawPtr::from_ref` instead of `value`? (I don't know)

crates/pico/src/memo_ref.rs
ch1ffa: Not sure I follow. It's not exactly the same. For `tracked` we need revision, so duplicate code goes only after `match self.kind`

rbalicki2: awesome. I didn't really read it too closely 😂

Issue #177, opened 2024-10-17
generate raw response types


rbalicki2:
If we are going to write data into the store with `normalizeData`, we need types to guide us, especially since the names of fieldsare going to be quite weird! We should probably have something akin to raw response types for this.

Note that the values of variables are part of the key, so ... that's actually not possible to do with a statically generated type :/. But maybe there's an alternative solution.

Anyway, this is a placeholder so we don't forget

rbalicki2:
Done! Thank you @PatrykWalach 🎉 🎉

PR #764, opened 2025-11-20
generate raw response types

PatrykWalach:
#177

---
This is **part 3 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #764 👈 
- <kbd>&nbsp;2&nbsp;</kbd> #766
- <kbd>&nbsp;1&nbsp;</kbd> #765

### DIFF
--- a/crates/artifact_content/src/raw_response_type.rs
+++ b/crates/artifact_content/src/raw_response_type.rs
@@ -1,5 +1,4 @@
 use intern::Lookup;
-use isograph_lang_types::TypeAnnotation;
 use isograph_schema::{
     IsographDatabase, MergedSelectionMap, MergedServerSelection, NetworkProtocol,
     server_object_selectable_named, server_scalar_entity_javascript_name,
@@ -41,7 +40,6 @@ pub fn generate_raw_response_type_inner<TNetworkProtocol: NetworkProtocol>(
     for item in selection_map.values() {
         match &item {
             MergedServerSelection::ScalarField(scalar_field) => {
-                raw_response_type_inner.push_str(indent);
                 let normalization_alias = scalar_field.normalization_alias();
                 let name = normalization_alias
                     .as_deref()
@@ -80,15 +78,16 @@ pub fn generate_raw_response_type_inner<TNetworkProtocol: NetworkProtocol>(
                             ),
                     },
                 );
-
-                let is_optional = matches!(
-                    server_scalar_selectable.target_scalar_entity,
-                    TypeAnnotation::Union(_)
-                );
+                // TODO: make sure undefined is treated as null when normalizing data
+                // https://github.com/isographlabs/isograph/issues/776
+                // let is_optional = matches!(
+                //     server_scalar_selectable.target_scalar_entity,
+                //     TypeAnnotation::Union(_)
+                // );
 
                 raw_response_type_inner.push_str(&format!(
-                    "{indent}{name}{}: {},\n",
-                    if is_optional { "?" } else { "" },
+                    "{indent}{name}: {},\n",
+                    // if is_optional { "?" } else { "" },
                     print_javascript_type_declaration(&raw_type)
                 ));
             }
@@ -130,15 +129,16 @@ pub fn generate_raw_response_type_inner<TNetworkProtocol: NetworkProtocol>(
                             raw_response_type_declaration.push_str(&format!("{indent}}}"));
                             raw_response_type_declaration
                         });
-
-                let is_optional = matches!(
-                    server_object_selectable.target_object_entity,
-                    TypeAnnotation::Union(_)
-                );
+                // TODO: make sure undefined is treated as null when normalizing data
+                // https://github.com/isographlabs/isograph/issues/776
+                // let is_optional = matches!(
+                //     server_object_selectable.target_object_entity,
+                //     TypeAnnotation::Union(_)
+                // );
 
                 raw_response_type_inner.push_str(&format!(
-                    "{indent}{name}{}: {},\n",
-                    if is_optional { "?" } else { "" },
+                    "{indent}{name}: {},\n",
+                    // if is_optional { "?" } else { "" },
                     print_javascript_type_declaration(&raw_type)
                 ));
             }
--- a/demos/github-demo/src/isograph-components/__isograph/Query/PullRequest/raw_response_type.ts
+++ b/demos/github-demo/src/isograph-components/__isograph/Query/PullRequest/raw_response_type.ts
@@ -1,14 +1,14 @@
 export type Query__PullRequest__rawResponse = {
-  repository____owner___v_repositoryOwner____name___v_repositoryName?: ({
+  repository____owner___v_repositoryOwner____name___v_repositoryName: ({
     id: string,
-    pullRequest____number___v_pullRequestNumber?: ({
+    pullRequest____number___v_pullRequestNumber: ({
       id: string,
       bodyHTML: string,
       comments____last___l_10: {
-        edges?: (ReadonlyArray<({
-          node?: ({
+        edges: (ReadonlyArray<({
+          node: ({
             id: string,
-            author?: ({
+            author: ({
               __typename: string,
               login: string,
             } | null),
@@ -23,7 +23,7 @@ export type Query__PullRequest__rawResponse = {
   viewer: {
     id: string,
     avatarUrl: string,
-    name?: (string | null),
+    name: (string | null),
   },
 }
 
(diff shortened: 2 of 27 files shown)


libs/isograph-react/src/tests/__isograph/Query/subquery/raw_response_type.ts
```diff
@@ -0,0 +1,9 @@
+export type Query__subquery__rawResponse = {
+  query: {
+    node____id___v_id: {
+      __typename: "Node",
```
PatrykWalach: needs FIX

crates/artifact_content/src/entrypoint_artifact.rs
```diff
@@ -307,20 +312,29 @@ pub(crate) fn generate_entrypoint_artifacts_with_client_field_traversal_result<
...
         file_name: *NORMALIZATION_AST_FILE_NAME,
-        type_and_field: ParentObjectEntityNameAndSelectableName {
+        type_and_field: Some(ParentObjectEntityNameAndSelectableName {
             parent_object_entity_name: type_name,
             selectable_name: field_name,
-        }
-        .some(),
+        }),
+    });
+    path_and_contents.push(ArtifactPathAndContent {
```
rbalicki2, line 319: should we unconditionally generate raw response types? I am not sure, tbh.

Reasons to do so:
- it's probably not that much more work than generating the normalization AST
- it also doesn't cause more code churn than the normalization AST does, as they change at the same time

Reasons not to:
- it's probably rarely used
- it exposes things like `node____id___v_id` and we should hide that kind of implementation detail from the user!

I don't feel strongly either way. What do you think?

crates/artifact_content/src/raw_response_type.rs
```diff
@@ -0,0 +1,130 @@
...
+        db,
+        &mut raw_response_type,
+        selection_map,
+        indentation_level + 1,
+    );
+    raw_response_type.push_str(&format!("{}}}\n", indent));
+    raw_response_type
+}
+
+pub fn generate_raw_response_type_inner<TNetworkProtocol: NetworkProtocol>(
```
rbalicki2, line 30: I'm really happy that this is just a pretty simple traversal. Looks very clean.

crates/artifact_content/src/raw_response_type.rs
```diff
@@ -0,0 +1,130 @@
...
+                generate_raw_response_type_inner(
+                    db,
+                    raw_response_type,
+                    &linked_field.selection_map,
+                    indentation_level + 1,
+                );
+                raw_response_type.push_str(&format!("{indent}}},\n"));
+            }
+            MergedServerSelection::ClientPointer(_) => {}
+            MergedServerSelection::InlineFragment(_) => {
```
rbalicki2: This would be true for GraphQL in general, but we don't support anything besides abstract -> concrete type refinements via `asConcreteType` fields, so we can generate a discriminated union here.

rbalicki2 (commented):
Hell yeah. I can't wait to have a `normalizeData` API that has types that we can sent to Randall 🤩

demos/pet-demo/src/components/__isograph/Viewer/NewsfeedPaginationComponent/raw_response_type.ts
```diff
@@ -0,0 +1,20 @@
+export type Viewer__NewsfeedPaginationComponent__rawResponse = {
+  node____id___v_id: {
+    __typename: "Viewer",
+    id: string,
+    newsfeed____skip___v_skip____limit___v_limit: {
+      __typename: "AdItem",
+      id: string,
+    } | {
+      __typename: "BlogItem",
```
PatrykWalach, line 9: This one looks good 🔥

demos/pet-demo/src/components/__isograph/Query/OnlyOneRootLoadablePet/raw_response_type.ts
```diff
@@ -0,0 +1,3 @@
+export type Query__OnlyOneRootLoadablePet__rawResponse = {
```
PatrykWalach, line 1: this is missing __typename 😠

crates/artifact_content/src/raw_response_type.rs
PatrykWalach: I've made it much more complicated with fragments

demos/pet-demo/src/components/__isograph/Query/OnlyOneRootLoadablePet/raw_response_type.ts
PatrykWalach: It's okay we don't generate __typename here, because it's not in the normalization ast

demos/pet-demo/src/components/__isograph/Mutation/MututalBestFriendSetterMutation/raw_response_type.ts
```diff
@@ -0,0 +1,17 @@
...
+    pet: {
+      id: string,
+      best_friend_relationship: {
+        best_friend: {
+          id: string,
+          firstName: string,
+          lastName: string,
+          picture: string,
+        },
+        picture_together?: (string | null),
```
rbalicki2: 1. not a blocker if it's not trivial to fix, but
2. we should confirm that if this isn't provided, we don't suspend. If we write `store.picture_together = data.picture_together`, we would write `undefined`, at which point this would cause us to suspend

```
export function readScalarFieldData(
  field: ReaderScalarField,
  storeRecord: StoreRecord,
  root: StoreLink,
  variables: Variables,
): ReadDataResult<
  string | number | boolean | StoreLink | DataTypeValue[] | null
> {
  const storeRecordName = getParentRecordKey(field, variables);
  const value = storeRecord[storeRecordName];
  // TODO consider making scalars into discriminated unions. This probably has
  // to happen for when we handle errors.
  if (value === undefined) {
    return {
      kind: 'MissingData',
      reason: 'No value for ' + storeRecordName + ' on root ' + root.__link,
      recordLink: root,
    };
  }
  return { kind: 'Success', data: value };
}
```

3. if it does cause us to suspend, I will inform Jerred about this, and he can use Isograph with that issue until we fix it

crates/graphql_network_protocol/src/process_type_system_definition.rs
```diff
@@ -388,7 +399,7 @@ fn process_object_type_definition(
                 ),
                 // This is bad data modeling, and we should do better.
                 javascript_type_override: concrete_type.map(|parent_concrete_type| {
-                    format!("'{parent_concrete_type}'").intern().into()
+                    format!("\"{parent_concrete_type}\"").intern().into()
```
rbalicki2, line 391: good catch, fixed in main fyi

demos/github-demo/src/isograph-components/__isograph/Query/PullRequest/raw_response_type.ts
```diff
@@ -0,0 +1,29 @@
...
+    id: string,
+    pullRequest____number___v_pullRequestNumber: {
+      id: string,
+      bodyHTML: string,
+      comments____last___l_10: {
+        edges: {
+          node: {
+            id: string,
+            author: {
+              __typename: string,
```
rbalicki2, line 12: 🤔  can we make this a union of concrete strings of types that implement `Actor`? Not a blocker

demos/pet-demo/src/components/__isograph/Query/Newsfeed/raw_response_type.ts
```diff
@@ -0,0 +1,19 @@
+export type Query__Newsfeed__rawResponse = {
+  viewer: {
+    id: string,
+    newsfeed____skip___l_0____limit___l_6: {
+      __typename: "AdItem",
+      id: string,
```
rbalicki2, line 6: so good

demos/pet-demo/src/components/__isograph/Query/OnlyOneRootLoadablePet/raw_response_type.ts
rbalicki2: okay sgtm

demos/pet-demo/src/components/__isograph/Query/HomeRoute/raw_response_type.ts
```diff
@@ -0,0 +1,10 @@
+export type Query__HomeRoute__rawResponse = {
+  pets: {
```
rbalicki2: this should be an array

Issue #779, opened 2025-11-23
Optimize artifact writing to disk


lucasmachadorj:
We currently rewrite the entire artifact set on every compilation. 
It represents a considerable amount of the compilation time.

The proposed optimization is to hash the artifacts and write to disk only the modified ones. 

The steps involved are the following:

1. **Hash current artifacts**
  a. When generating artifacts, compute each artifact hash based on the file content.
  b. Maintain a lookup table of `{ filepath → hash }` under __isograph
 
2.  **Check for changes before writing**  
  a. For each regenerated artifact, compare its hash against the stored hash
  b. Only write the file if:
       * No existing hash is found → new file; or
       * Hash differs → modified file
 
3. **Prune orphaned artifacts** 
   a. After generation, find existing filepaths in the mapping that were not produced during the current run.
 
4. **Save new lookup table to disk**

rbalicki2:
LETS GOOO

PR #784, opened 2025-11-23
Optimize write artifacts to disk in watch mode

lucasmachadorj:
This PR creates a FileSystemState that is used to optimize the performance of write-to-disk artifacts in watch mode.
The first compilation, changes in config, or not running compilation in watch mode (`compile_and_print`) makes the artifacts directory being deleted and recreated.

While in watch mode, on every reload, it creates a new file state based on artifact filepaths, their content and hashes and compare the hashes from old and new states to create a vector of filesystem operations (create dir, create file, delete fir, delete file) which are properly handled by `write_artifacts.rs` 

This is the implementation of the issue #779

### DIFF
--- a/crates/artifact_content/src/file_system_state.rs
+++ b/crates/artifact_content/src/file_system_state.rs
@@ -1,7 +1,7 @@
 use pico::Index;
 use std::{
     collections::{HashMap, HashSet},
-    path::PathBuf,
+    path::Path,
 };
 
 use crate::operation_text::hash;
@@ -12,6 +12,7 @@ use common_lang_types::{
 use isograph_config::PersistedDocumentsHashAlgorithm;
 
 #[derive(Debug, Clone, Default)]
+#[expect(clippy::type_complexity)]
 pub struct FileSystemState {
     root_files: HashMap<ArtifactFileName, (Index<FileContent>, ArtifactHash)>,
     nested_files: HashMap<
@@ -21,7 +22,7 @@ pub struct FileSystemState {
 }
 
 impl FileSystemState {
-    pub fn recreate_all(state: &Self, artifact_directory: &PathBuf) -> Vec<FileSystemOperation> {
+    pub fn recreate_all(state: &Self, artifact_directory: &Path) -> Vec<FileSystemOperation> {
         let mut operations: Vec<FileSystemOperation> = Vec::new();
         operations.push(FileSystemOperation::DeleteDirectory(
             artifact_directory.to_path_buf(),
@@ -57,7 +58,7 @@ impl FileSystemState {
         operations
     }
 
-    pub fn diff(old: &Self, new: &Self, artifact_directory: &PathBuf) -> Vec<FileSystemOperation> {
+    pub fn diff(old: &Self, new: &Self, artifact_directory: &Path) -> Vec<FileSystemOperation> {
         let mut operations: Vec<FileSystemOperation> = Vec::new();
 
         let mut new_server_object_entity_name_set = HashSet::new();
@@ -72,19 +73,14 @@ impl FileSystemState {
             for (new_selectable, new_files) in new_selectable_map {
                 new_selectable_set.insert((*new_server_object_entity_name, *new_selectable));
                 let new_selectable_path = new_server_object_path.join(new_selectable);
+                let old_files_for_selectable =
+                    old_selectables_for_object.and_then(|s| s.get(new_selectable));
 
-                let should_create_dir = old_selectables_for_object
-                    .and_then(|s| s.get(new_selectable))
-                    .is_none();
-
-                if should_create_dir {
+                if old_files_for_selectable.is_none() {
                     operations.push(FileSystemOperation::CreateDirectory(
                         new_selectable_path.clone(),
                     ));
                 }
-
-                let old_files_for_selectable =
-                    old_selectables_for_object.and_then(|s| s.get(new_selectable));
 
                 for (new_file_name, (new_index, new_hash)) in new_files {
                     let new_file_path = new_selectable_path.join(new_file_name);
@@ -166,6 +162,7 @@ impl FileSystemState {
     }
 }
 
+#[expect(clippy::type_complexity)]
 impl From<&[ArtifactPathAndContent]> for FileSystemState {
     fn from(artifacts: &[ArtifactPathAndContent]) -> Self {
         let mut root_files = HashMap::new();
@@ -232,7 +229,7 @@ mod tests {
         ArtifactPathAndContent {
             artifact_path: ArtifactPath {
                 type_and_field,
-                file_name: ArtifactFileName::from(file_name.intern()),
+                file_name: file_name.intern().into(),
             },
             file_content: FileContent::from(content.to_string()),
         }
--- a/crates/isograph_compiler/src/write_artifacts.rs
+++ b/crates/isograph_compiler/src/write_artifacts.rs
@@ -1,4 +1,7 @@
-use std::{fs, path::PathBuf};
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
 use common_lang_types::{
     ArtifactPathAndContent, Diagnostic, DiagnosticResult, FileSystemOperation,
@@ -9,14 +12,14 @@ use artifact_content::FileSystemState;
 #[tracing::instrument(skip_all)]
 pub(crate) fn get_file_system_operations(
     paths_and_contents: &[ArtifactPathAndContent],
-    artifact_directory: &PathBuf,
+    artifact_directory: &Path,
     file_system_state: &mut Option<FileSystemState>,
 ) -> Vec<FileSystemOperation> {
     let new_file_system_state = paths_and_contents.into();
     let operations = match file_system_state {
         None => FileSystemState::recreate_all(&new_file_system_state, artifact_directory),
         Some(file_system_state) => FileSystemState::diff(
-            &file_system_state,
+            file_system_state,
             &new_file_system_state,
             artifact_directory,
         ),
@@ -38,7 +41,7 @@ pub(crate) fn apply_file_system_operations(
                 if path.exists() {
                     fs::remove_dir_all(path.clone()).map_err(|e| {
                         unable_to_do_something_at_path_diagnostic(
-                            &path,
+                            path,
                             &e.to_string(),
                             "delete directory",
                         )
@@ -49,7 +52,7 @@ pub(crate) fn apply_file_system_operations(
                 count += 1;
                 fs::create_dir_all(path.clone()).map_err(|e| {
                     unable_to_do_something_at_path_diagnostic(
-                        &path,
+                        path,
                         &e.to_string(),
                         "create directory",
                     )
@@ -62,7 +65,7 @@ pub(crate) fn apply_file_system_operations(
                     .file_content;
                 fs::write(path.clone(), content.as_bytes()).map_err(|e| {
                     unable_to_do_something_at_path_diagnostic(
-                        &path,
+                        path,
                         &e.to_string(),
                         "write contents of file",
                     )
@@ -70,7 +73,7 @@ pub(crate) fn apply_file_system_operations(
             }
             FileSystemOperation::DeleteFile(path) => {
                 fs::remove_file(path.clone()).map_err(|e| {
-                    unable_to_do_something_at_path_diagnostic(&path, &e.to_string(), "delete file")
+                    unable_to_do_something_at_path_diagnostic(path, &e.to_string(), "delete file")
                 })?;
             }
         }


demos/pet-demo/src/components/__isograph/.artifact_lookup_cache.json
```diff
@@ -0,0 +1,249 @@
+{
```
rbalicki2: Hm... IMO we shouldn't write this file to disk. I think it's sufficient to
1. store this in memory
2. recreate the __isograph folder entirely whenever we restart the compiler
3. optimize for watch mode

In the future, I do want to save state between compilations, but 1. it shouldn't live in `__isograph`, but instead somewhere else, and 2. it should be done via something like serde, so that there's no manual maintenance

lucasmachadorj:
In pet-demo:

`write_artifacts_to_disk` first writing time:
Screenshot 2025-11-24 at 22 24 48

`write_artifacts_to_disk` after adding a query field to `PetDetailDeferredRouteComponent` in watch mode: `pnpm iso --watch`

Screenshot 2025-11-24 at 22 25 10

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,71 @@
+use std::collections::{BTreeSet, HashMap};
+
+use crate::operation_text::hash;
+use isograph_config::PersistedDocumentsHashAlgorithm;
+
+type Filepath = String;
```
rbalicki2: - We should (almost never) use type aliases like this, especially when they are aliases the same underlying type (i.e. multiple aliases to String). It's too easy to accidentally use a Filepath where we intend to use a `FileHash`
- Let's try to reuse the types we have. I just shipped a change to main to split `ArtifactPathAndContent` into two parts: `ArtifactPath` and `FileContent`. I think we should be able to use these instead.
- In particular, `ArtifactPath` contains a `FileName`, which is a string that we expect to appear repeatedly. In cases like that, we intern the string (i.e. replace it with an integer and store it in an array.) It may not matter much here, but in general, strings are expensive and we should try not to use them much.
- Let's also use a newtype wrapper for `FileHash` (i.e. `struct FileHash(String)`). (It doesn't make sense to intern it, I think...)

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,71 @@
...
+        for key in candidate_to_update {
+            if self.get_hashed_content(&key).unwrap() != other.get_hashed_content(&key).unwrap() {
+                to_add.push(key);
+            }
+        }
+        (to_delete, to_add)
+    }
+}
+
+impl Default for FileSystemState {
```
rbalicki2: we can derive `Default`

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -7,62 +8,132 @@ use std::{
...
+            path_and_content.file_content.clone(),
+        );
+        artifact_map.insert(relative_file_path, (absolute_file_path, path_and_content));
+    }
+
+    let mut artifacts_to_disk = ChangedArtifacts::new();
+
+    if file_system_state.is_empty() {
+        artifacts_to_disk.artifacts_to_write = artifact_map
+            .into_iter()
```
rbalicki2: `.values().collect()`

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,71 @@
...
+impl FileSystemState {
+    pub fn new() -> Self {
+        Self {
+            state: HashMap::new(),
+            sorted_keys: BTreeSet::new(),
+        }
+    }
+
+    pub fn insert(&mut self, filename: Filepath, content: FileContent) {
+        let hashed_content = hash(&content, PersistedDocumentsHashAlgorithm::Sha256);
```
rbalicki2: https://stackoverflow.com/questions/2722943/is-calculating-an-md5-hash-less-cpu-intensive-than-sha-family-functions indicates that we should use md5 here

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,71 @@
...
+use crate::operation_text::hash;
+use isograph_config::PersistedDocumentsHashAlgorithm;
+
+type Filepath = String;
+type FileContent = String;
+type FileHash = String;
+
+#[derive(Debug, Clone)]
+pub struct FileSystemState {
+    state: HashMap<Filepath, (FileContent, FileHash)>,
```
rbalicki2: We need to keep track of directories, too, otherwise we will have orphaned directories.

If we
- run pet demo in watch mode, then
- add `type Foo { id: ID! }` to the schema
- then add and remove `export const foo = iso(`field Foo.Bar {}`)();` to a random file,

Then we will find that that `src/components/__isograph/Foo/Bar` exists, but is empty.

-------

So, IMO, it's probably best to have a multi-level hashmap: `Option<TypeAndField> -> artifact name -> artifact hash`.

Then, we should implement a function that converts `Vec<ArtifactPathAndContent>` to `MultiLevelFileSystemHashMap`, and a separate function that compares these two. I would imagine that the comparison would be a vec of operators, like `vec![CreateDirectory(path), WriteFile(path, content), DeleteDirectory(d)]`, and the very first time, the first instruction can be something like `DeleteDirectory(root_artifact_path)`.

rbalicki2 (changes requested):
Amazing!!! This great work. Insane how much faster the compiler is now 😂  I can't believe we didn't do this sooner 😂  I guess there is still low-hanging fruit

I really like the big picture structure here. Let's change some of the ways we're representing things, namely to:
- reuse existing data structures where possible, and
- handle deleting directories, and
- represent the work that needs to be done as a `Vec<Action>` instead of as a struct that we need to interpret.

🙌 
Once you address these changes, I'll re-review.

crates/artifact_content/src/file_system_state.rs
rbalicki2: the more I think about it, the more I think we should have a hash map like so: `HashMap<ServerObjectEntityName, HashMap<ClientSelectableName, HashMap<ArtifactName, ArtifactHash>>>`

rbalicki2: (because that makes it easier to delete directories like `__isograph/Type/`)

ch1ffa: In multi-project perspective we also need a `HashMap<SourceId, Vec<Artifacts>>` or something to keep track of what shared artifacts we have. Maybe this view can help as well

crates/artifact_content/src/file_system_state.rs
ch1ffa: In relay they use a user defined function that generate hashes, so one can use version control system to get file hash of existing file. I think is a good idea to have a enum here and associated hash algorithm so we could extend it later.

And I was thinking to use `blake3` for hashing files

crates/isograph_compiler/src/batch_compile.rs
```diff
@@ -77,18 +80,25 @@ fn print_stats(elapsed_time: Duration, stats: CompilationStats) {
 }
 
 /// This the "workhorse" command of batch compilation.
-#[tracing::instrument(skip(db))]
+#[tracing::instrument(skip(state))]
 pub fn compile<TNetworkProtocol: NetworkProtocol>(
-    db: &IsographDatabase<TNetworkProtocol>,
+    state: &mut CompilerState<TNetworkProtocol>,
```
ch1ffa, line 85: compile could be memoized one day, so maybe we should keep db here, return new file state and perform writing artifacts outside of compile? i.e. compile should be named `create_artifacts` or something. I mean I would avoid having mutable arguments everywhere to keep it "pico ready" so it would be easier to refactor later

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -7,63 +8,134 @@ use std::{
 use common_lang_types::{ArtifactPathAndContent, Diagnostic, DiagnosticResult};
 use intern::string_key::Lookup;
 
-#[tracing::instrument(skip(paths_and_contents, artifact_directory))]
-pub(crate) fn write_artifacts_to_disk(
+use crate::changed_artifacts::ChangedArtifacts;
+use artifact_content::FileSystemState;
+
+#[tracing::instrument(skip(paths_and_contents, artifact_directory, file_system_state))]
```
ch1ffa: nit: let's use `skip_all`. I don't think we ever need to log fn values tbh

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -7,63 +8,134 @@ use std::{
...
+                &absolute_directory.to_path_buf(),
                 &e.to_string(),
                 "create directory",
             )
         })?;
 
-        let absolute_file_path =
-            absolute_directory.join(path_and_content.artifact_path.file_name.lookup());
-        let mut file = File::create(&absolute_file_path).map_err(|e| {
+        let mut file = File::create(&path).map_err(|e| {
```
ch1ffa: Another thing I was planning to do (or maybe we can do it as a follow up), is to write to a tmp file first and then do atomic replace. Some say it is a proper way to update files that are heavily used by another service, e.g. lsp

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.root_files.is_empty() && self.nested_files.is_empty()
+    }
+
+    // Computes filesystem operations needed to transform this state into the target state.
+    // Returns operations to create, update, and delete files and directories. Files are updated
+    // only when their hash differs. If the current state is empty, emits a DeleteDirectory
+    // operation for the artifact root followed by recreation of all files. Empty directories
+    // are automatically removed after their files are deleted.
+    pub fn diff(&self, new: &Self, artifact_directory: &PathBuf) -> Vec<FileSystemOperation> {
```
lucasmachadorj: This function became quite long, but I strove to keep everything quite explicit.
It handles two scenarios: state is empty vs compare new and old states

For an empty state, delete the artifact directory and recreate everything.

Otherwise, operations are pushed in the order: 
- create new directories if they don't exist in old state
- write file if they are new or updated
- delete nested files if they exist in old state but not in new
- delete orphan directories

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -1,69 +1,71 @@
...
+) -> Vec<FileSystemOperation> {
+    let artifacts: Vec<ArtifactPathAndContent> = paths_and_contents.into_iter().collect();
+    let new_file_system_state = FileSystemState::from_artifacts(artifacts);
+    let operations = file_system_state.diff(&new_file_system_state, artifact_directory);
+    *file_system_state = new_file_system_state;
+    operations
+}
 
+#[tracing::instrument(skip_all)]
+pub(crate) fn apply_file_system_operations(
```
lucasmachadorj, line 32: I renamed `write_to_disk` to `apply_file_system_operations` and wonder if we should encapsulate this function and `get_file_system_operations` in `FileSystemState` and remove `write_to_artifacts`.

crates/isograph_compiler/src/write_artifacts.rs
lucasmachadorj: It's interesting. 
We could do it in a follow-up

https://users.rust-lang.org/t/how-to-write-replace-files-atomically/42821

crates/isograph_compiler/src/write_artifacts.rs
rbalicki2: Not sure yet, but it's not a big deal to me if stuff is strewn across multiple files. After we land this, we can figure out what goes where. As in, I'm mostly concerned about getting the big picture logic right, and then we can move stuff around

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -1,69 +1,71 @@
...
+    operations: Vec<FileSystemOperation>,
+) -> DiagnosticResult<usize> {
     let mut count = 0;
```
rbalicki2, line 36: Hm. This isn't really accurately conveying the number of artifacts written. It's more like the number of operations.

I think the most informative thing would be to increment this only for `WriteFile`, and then display `X files modified or written`

(And otherwise, we could just `.len()`)

crates/isograph_compiler/src/batch_compile.rs
rbalicki2: I think you're right that this is a misnamed function. I wasn't able to think of a good name to name this though 😂 . But IMO `get_artifact_paths_and_content` is the only thing that the memoized `compile` would do, and that takes `db`, so we're good on that front (IMO)

crates/common_lang_types/src/file_system_operation.rs
```diff
@@ -0,0 +1,11 @@
+use std::path::PathBuf;
+
+use crate::FileContent;
+
+#[derive(Debug, Clone)]
+pub enum FileSystemOperation {
```
rbalicki2, line 8: `FileSystemOperation<'a> { WriteFile(&'a Path, &'a FileContent), ... }` will allow us to avoid cloning the file content each time!

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+                    .insert(path_and_content.artifact_path.file_name, value);
+            }
+            None => {
+                self.root_files
+                    .insert(path_and_content.artifact_path.file_name, value);
+            }
+        }
+    }
+
+    pub fn from_artifacts(artifacts: Vec<ArtifactPathAndContent>) -> Self {
```
rbalicki2: there's no need to make these functions methods on `FileSystemState`, IMO. It's generally cleaner to have pure functions.

In the `from_artifacts` case, it might be cleanest to have `impl From<Vec<ArtifactPathAndContent>> for FileSystemState`. But IMO `insert` can just be a free fn (if needed, it also doesn't seem too complicated to just inline into `impl From<Vec<...>>`

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+        let mut state = Self::default();
+        for artifact in artifacts {
+            state.insert(&artifact);
+        }
+        state
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.root_files.is_empty() && self.nested_files.is_empty()
```
rbalicki2: It would be cleaner to not check `is_empty`, but instead have `CompilerState` have `Option<FileSystemState>`

crates/artifact_content/src/file_system_state.rs
rbalicki2: Let's have `Option<FileSystemState>` instead of a `FileSystemState` with an `is_empty` function. And then let's split this into two functions, one which handles the `None` case and one which handles the `Some` case.

BTW this principal is articulated here: https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+                        selectable_path.clone(),
+                    ));
+
+                    for (file_name, (content, _)) in files {
+                        let file_path = selectable_path.join(file_name.to_string());
+                        operations.push(FileSystemOperation::WriteFile(file_path, content.clone()));
+                    }
+                }
+
+                for (file_name, (content, _)) in &new.root_files {
```
rbalicki2: this loop is accidentally inside the other loop, it belongs right before `return operations`

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+        let mut operations: Vec<FileSystemOperation> = Vec::new();
+
+        if self.is_empty() {
+            operations.push(FileSystemOperation::DeleteDirectory(
+                artifact_directory.to_path_buf(),
+            ));
+
+            for (server_object, selectables) in &new.nested_files {
+                let server_object_path = artifact_directory.join(server_object.to_string());
+                operations.push(FileSystemOperation::CreateDirectory(
```
rbalicki2, line 36: I think we should skip this `CreateDirectory`, as `CreateDirectory` already creates calls `create_dir_all`, which will:
/// Recursively create a directory and all of its parent components if they
/// are missing.

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+                for (file_name, (content, _)) in &new.root_files {
+                    let file_path = artifact_directory.join(file_name.to_string());
+                    operations.push(FileSystemOperation::WriteFile(file_path, content.clone()));
+                }
+            }
+
+            return operations;
+        }
+
+        let mut new_server_objects: HashSet<&ServerObjectEntityName> = HashSet::new();
```
rbalicki2: We should not generally create references to `&Copy` types, and anything that's been interned (e.g. `ServerObjectEntityName` and `SelectableName`) implements Copy. So instead, let's just do `*server_object_entity_name` wherever necessary

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+        let mut new_selectables: HashSet<(&ServerObjectEntityName, &SelectableName)> =
+            HashSet::new();
+
+        for (server_object, selectables) in &new.nested_files {
+            new_server_objects.insert(server_object);
+            let server_object_path = artifact_directory.join(server_object.to_string());
```
rbalicki2: `to_string()` will allocate a string unnecessarily. We can use `.lookup()` on all interned strings to get the underlying `&'static str`, which is sufficient here.

rbalicki2: Though even cleaner is to `impl AsRef<Path>` for interned strings, which I just landed in master (https://github.com/isographlabs/isograph/commit/7b16c1d312e6eaf0954064889dc8c11995e22d8b), so we can just do `.join(server_object)`.

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+                        operations.push(FileSystemOperation::DeleteFile(file_path));
+                    }
+                }
+
+                if !new_selectables.contains(&(server_object, selectable)) {
+                    operations.push(FileSystemOperation::DeleteDirectory(selectable_path));
+                }
+            }
+
+            if !new_server_objects.contains(server_object) {
```
rbalicki2:
```
if !new_server_objects.contains(server_object) { delete directory } else { for loop for selectables }
```

Likewise above ☝️  for the file level

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+                parent_object_entity_name: ServerObjectEntityName::from(s.intern()),
+                selectable_name: SelectableName::from(sel.intern()),
+            }),
+            _ => None,
+        };
+
+        ArtifactPathAndContent {
+            artifact_path: ArtifactPath {
+                type_and_field,
+                file_name: ArtifactFileName::from(file_name.intern()),
```
rbalicki2: `file_name.intern().into()`

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+pub struct FileSystemState {
+    root_files: HashMap<ArtifactFileName, (FileContent, ArtifactHash)>,
+    nested_files: HashMap<
+        ServerObjectEntityName,
+        HashMap<SelectableName, HashMap<ArtifactFileName, (FileContent, ArtifactHash)>>,
+    >,
+}
+
+impl FileSystemState {
+    pub fn insert(&mut self, path_and_content: &ArtifactPathAndContent) {
```
rbalicki2: this should take ownership of `ArtifactPathAndContent`, so we don't have to clone.

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+use crate::operation_text::hash;
+use common_lang_types::{
+    ArtifactFileName, ArtifactHash, ArtifactPathAndContent, FileContent, FileSystemOperation,
+    SelectableName, ServerObjectEntityName,
+};
+use isograph_config::PersistedDocumentsHashAlgorithm;
+
+#[derive(Debug, Clone, Default)]
+pub struct FileSystemState {
+    root_files: HashMap<ArtifactFileName, (FileContent, ArtifactHash)>,
```
rbalicki2: We unfortunately clone `FileContent` a lot! That's not really ideal, since these can be quite big strings.

We don't need to. Let's have this just contain the `ArtifactHash` and an `Index<FileContent>` (from `Pico`), and in `write_artifacts`, pass `&Vec<ArtifactPathAndContent>` and look up the content in the vec. We can use `expect` to deal with the fact that `vec.get(index)` returns `Option`

Note that we could also avoid cloning by having a single function that:
- accepts the old `FileSystemState` and the new `Vec<ArtifactPathAndContent>`, and
- creates the new `FileSystemState` and `Vec<FileSystemOperation>`.
Then, we can avoid storing the `FileContent` in the `FileSystemState` entirely. But IMO, it's easier to understand if we divide this into two separate steps (first creating the `FileSystemState`, then comparing the two.)

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,530 @@
...
+
+            if should_write {
+                operations.push(FileSystemOperation::WriteFile(
+                    file_path,
+                    new_content.clone(),
+                ));
+            }
+        }
+
+        for (server_object, selectables) in &self.nested_files {
```
rbalicki2: nit: could you rename these `existing_server_object` etc, and above `new_server_object`? Likewise for th other variables. This would make it easier to follow.

rbalicki2:
oh and clippy is complaining https://github.com/isographlabs/isograph/actions/runs/19684421184/job/56406099824?pr=784

crates/artifact_content/src/file_system_state.rs
lucasmachadorj: It's being done above

```
                    if !exist_in_new {
                        let file_path = selectable_path.join(file_name);
                        operations.push(FileSystemOperation::DeleteFile(file_path));
                    }
                }

                if !new_selectables.contains(&(server_object, selectable)) {
                    operations.push(FileSystemOperation::DeleteDirectory(selectable_path));
                }
            }

            if !new_server_objects.contains(server_object) {
                operations.push(FileSystemOperation::DeleteDirectory(server_object_path));
            }
```

crates/artifact_content/src/file_system_state.rs
lucasmachadorj: Hmm the compiler complains: `type must be known at this point`

rbalicki2: can you push your latest and I'll see? It should be able to infer the type

crates/artifact_content/src/file_system_state.rs
lucasmachadorj: I refactored to contain the index and it worked well.
Later I'll see how the alternative would work.

crates/artifact_content/src/file_system_state.rs
rbalicki2: Sorry, I meant that we end up deleting a bunch of individual files, then we delete the directory. We don't need to do that! We can just delete the directory directly. So I'm imagining that we skip the for loop if we're going to delete the directory

crates/common_lang_types/src/file_system_operation.rs
lucasmachadorj: Solved with Index

crates/isograph_compiler/src/write_artifacts.rs
```diff
@@ -1,69 +1,88 @@
...
-) -> DiagnosticResult<usize> {
-    if artifact_directory.exists() {
-        fs::remove_dir_all(artifact_directory).map_err(|e| {
-            unable_to_do_something_at_path_diagnostic(
+    file_system_state: &mut Option<FileSystemState>,
+) -> Vec<FileSystemOperation> {
+    let new_file_system_state = FileSystemState::from(paths_and_contents);
+    let operations = match file_system_state {
+        None => {
+            let operations =
```
rbalicki2: nit: no need to create a variable here or below, just `None => FileSystemState::recreate_all(...)`

crates/artifact_content/src/file_system_state.rs
lucasmachadorj: Got it, you're right! Just landed an improv

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,504 @@
...
+        }
+
+        operations
+    }
+
+    pub fn diff(old: &Self, new: &Self, artifact_directory: &PathBuf) -> Vec<FileSystemOperation> {
+        let mut operations: Vec<FileSystemOperation> = Vec::new();
+
+        let mut new_server_objects: HashSet<ServerObjectEntityName> = HashSet::new();
+        let mut new_selectables: HashSet<(ServerObjectEntityName, &SelectableName)> =
```
rbalicki2: `SelectableName` instead of `&SelectableName`, since `SelectableName` implements `Copy` (and is small)

rbalicki2: You also don't need these type annotations, they're inferred from the `new_selectables.insert(...)` line

rbalicki2:
Can you address the clippy errors? FWIW I'll look into why CI is not auto-running for you. (Perhaps after you accept the invitation to be a contributor it will auto-run?)

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,516 @@
...
+            let new_server_object_path = artifact_directory.join(*new_server_object_entity_name);
+
+            let old_selectables_for_object = old.nested_files.get(new_server_object_entity_name);
+
+            for (new_selectable, new_files) in new_selectable_map {
+                new_selectable_set.insert((*new_server_object_entity_name, *new_selectable));
+                let new_selectable_path = new_server_object_path.join(new_selectable);
+
+                let should_create_dir = old_selectables_for_object
+                    .and_then(|s| s.get(new_selectable))
```
rbalicki2: let's extract `old_selectables_for_object.and_then(|s| s.get(new_selectable))` into a variable and use it here and later

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,516 @@
...
+                parent_object_entity_name: ServerObjectEntityName::from(s.intern()),
+                selectable_name: SelectableName::from(sel.intern()),
+            }),
+            _ => None,
+        };
+
+        ArtifactPathAndContent {
+            artifact_path: ArtifactPath {
+                type_and_field,
+                file_name: ArtifactFileName::from(file_name.intern()),
```
rbalicki2: I'm not sure why the compiler was complaining, but FWIW it's not complaining when I do `sel.intern().into()`

crates/artifact_content/src/file_system_state.rs
```diff
@@ -0,0 +1,516 @@
...
+use common_lang_types::{
+    ArtifactFileName, ArtifactHash, ArtifactPathAndContent, FileContent, FileSystemOperation,
+    SelectableName, ServerObjectEntityName,
+};
+use isograph_config::PersistedDocumentsHashAlgorithm;
+
+#[derive(Debug, Clone, Default)]
+pub struct FileSystemState {
+    root_files: HashMap<ArtifactFileName, (Index<FileContent>, ArtifactHash)>,
+    nested_files: HashMap<
```
rbalicki2, line 18: for these, do `#[expect(clippy::type_complexity)]`, you can find other examples of that in the codebase as well.

you can run `cargo clippy` locally as well

rbalicki2 (changes requested):
just some small added nits, in addition to clippy

Issue #797, opened 2025-11-26
Canonical ID directives


rbalicki2:
# What?

# Rough outline:

## Prefactoring

- [ ] Everywhere we write `"id"` in rust code, change that to `*ID_FIELD_NAME`
- [ ] Modify how directives are deserialized in graphql_network_protocol. We should not deserialize individual directives, but instead, be able to deserialize an object (or an enum, but that's not relevant yet) from a vector of directives. Modify graphql_network_protocol to deserialize an object `struct ServerObjectEntityDirectives { expose_field: Option<ExposeFieldDirective> }`. Behavior should be unchanged.
- [ ] This struct should end up on `ServerObjectEntity`, but we should not use it (yet?)

## Generate the correct ID field
- [ ] Add a field to `ServerSchemaEntityDirectives`, `canonical_id: CanonicalIdDirective`. Test that it works. Then add code to panic if we encounter one. `struct CanonicalIdDirective { field_name: ServerScalarSelectableName }`.
- [ ] Everywhere we use *ID_FIELD_NAME, we should use the canonical id name if it exists. This should mean that we automatically select different fields where we would automatically select `id`, so we should see changes to query text, normalization AST, etc.

## Gather all of the reachable entities
- [ ] When we create a merged selection map (`create_merged_selection_map_for_field_and_insert_into_global_map`), we should also return a `BTreeSet<SelectionType<ServerObjectEntityName, ServerScalarEntityName>>` of reachable entities with non-canonical IDs
- [ ] Then, modify the entrypoint.ts file to include something like

```
import pet__canonical_id from '../../Pet/canonical_id'
const reachableEntities = { Pet: pet__canonical_id };
```
and add that array to the exported entrypoint and modify the types to add that.
- `generate_entrypoint_artifacts` should also accept an `&mut HashSet<SelectionType<ServerObjectEntityName, ServerScalarEntityName>>` and extend that with the encountered entities
- after we generate all entrypoints, generate the canonical ID files for the entities that were reached

## Runtime

- Whenever we encounter an entrypoint, we need to merge the canonical entities into some canonical entity map that is held in the environment.
- Wherever we select `.id`, we should instead call a function `getCanonicalId(environment, typeName): string` which defaults to `id`

## Roll it out

- Make pets have a canonical ID `petId`

PR #801, opened 2025-11-28
Create the canonical id directive

lucasmachadorj:
This PR creates a new field directive (`CanonicalFieldDirective`) and includes it to `ServerObjectEntityDirectives` as optional.

Also, a temporary `panic` if it's used while it's under implementation.

#797


rbalicki2 (approved):
LGTM just run `pnpm generate-fixture-tests` and this should be good to land!

Thank you for doing this!! Lets gooo!!!

PR #803, opened 2025-11-28
Refactor id field name resolution

lucasmachadorj:
Currently, Isograph hardcodes the assumption that all entities use "id" as their identifier field. However, some schemas may use different conventions (e.g., petId, userId). This PR implements the dynamic id field resolution.

#### Add canonical ID field support
- canonical_id_field_name: Option<StringLiteralValue> field to GraphQLSchemaObjectAssociatedData
- Initialized as None for all entity types during schema processing

#### Add network protocol trait method
```
fn get_id_field_name(
    db: &IsographDatabase<Self>,
    entity_name: &ServerObjectEntityName,
) -> ServerScalarSelectableName;
```

#### Update all ID field access points
1. Normalization key generation 
2. ID field lookups
3. Query variable generation: enables queries like query($petId: ID!) instead of always $id
4. Refetch field generation
5. Normalization key methods

Behavior is unchanged for entities without @canonicalId directive

#797

### DIFF
--- a/crates/graphql_network_protocol/src/process_type_system_definition.rs
+++ b/crates/graphql_network_protocol/src/process_type_system_definition.rs
@@ -88,7 +88,6 @@ pub fn process_graphql_type_system_document(
                     },
                     GraphQLObjectDefinitionType::Object,
                     &mut refetch_fields,
-                    None,
                 )?;
 
                 directives
@@ -120,7 +119,6 @@ pub fn process_graphql_type_system_document(
                         },
                         GraphQLObjectDefinitionType::Interface,
                         &mut refetch_fields,
-                        None,
                     )?;
 
                 type_system_entities.push(process_object_type_definition_outcome.object_selected());
@@ -151,7 +149,6 @@ pub fn process_graphql_type_system_document(
                         },
                         GraphQLObjectDefinitionType::InputObject,
                         &mut refetch_fields,
-                        None,
                     )?;
 
                 type_system_entities.push(process_object_type_definition_outcome.object_selected());
@@ -197,7 +194,6 @@ pub fn process_graphql_type_system_document(
                         },
                         GraphQLObjectDefinitionType::Union,
                         &mut refetch_fields,
-                        None,
                     )?;
 
                 type_system_entities.push(process_object_type_definition_outcome.object_selected());
@@ -342,7 +338,6 @@ fn process_object_type_definition(
     associated_data: GraphQLSchemaObjectAssociatedData,
     type_definition_type: GraphQLObjectDefinitionType,
     refetch_fields: &mut Vec<ExposeFieldToInsert>,
-    server_object_entity_directives: Option<ServerObjectEntityDirectives>,
 ) -> DiagnosticResult<(
     ProcessObjectTypeDefinitionOutcome<GraphQLNetworkProtocol>,
     Vec<GraphQLDirective<GraphQLConstantValue>>,
@@ -356,7 +351,7 @@ fn process_object_type_definition(
         name: object_type_definition.name.item,
         concrete_type,
         network_protocol_associated_data: associated_data,
-        server_object_entity_directives: server_object_entity_directives.unwrap_or_default(),
+        server_object_entity_directives: ServerObjectEntityDirectives::default(),
     }
     .with_location(object_type_definition.name.location.into());
 


crates/isograph_schema/src/network_protocol.rs
```diff
@@ -56,6 +57,11 @@ pub trait NetworkProtocol:
         operation_name: ServerObjectEntityName,
         indentation_level: u8,
     ) -> QueryExtraInfo;
+
+    fn get_id_field_name(
```
rbalicki2, line 61: Hm. I like where your head is at re: making this customizable. Does it make sense to make this specific to the network protocol? What network protocol would use a different id field name? Maybe it makes sense to put this into the config? I don't really have a strong opinion here (other than to punt on this until there's a clear need here)

If we do make it a method on NetworkProtocol then let's make it have a default implementation of returning *ID_FIELD_NAME

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -234,12 +246,39 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+            .expect(
+                "Expected entity to exist. \
+                This is indicative of a bug in Isograph.",
+            )
+            .lookup(db);
+
+        match entity
+            .network_protocol_associated_data
+            .canonical_id_field_name
+        {
```
rbalicki2: .map(...).unwrap_or_else(...) is more idiomatic fyi

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -234,12 +246,39 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+
+    fn get_id_field_name(
+        db: &IsographDatabase<GraphQLNetworkProtocol>,
+        server_object_entity_name: &ServerObjectEntityName,
+    ) -> ServerScalarSelectableName {
+        let entity = server_object_entity_named(db, *server_object_entity_name)
+            .as_ref()
+            .expect(
+                "Expected entity to exist. \
+                This is indicative of a bug in Isograph.",
```
rbalicki2: We shouldn't panic here, because this can be called from anywhere, not just after validate.rs is called. So we should return a DiagnosticResult. It's only "safe" to panic from artifact generation code, and even that code needs to stop panicking at some point

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -234,12 +246,39 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+            None => *ID_FIELD_NAME,
+        }
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Hash)]
 pub struct GraphQLSchemaObjectAssociatedData {
     pub original_definition_type: GraphQLSchemaOriginalDefinitionType,
     pub subtypes: Vec<UnvalidatedTypeName>,
+    pub canonical_id_field_name: Option<StringLiteralValue>,
```
rbalicki2: Hm. I think this isn't specific to the network protocol, so let's add the canonical field to the server object entity directly (or rather, add the server entity directive set, and that should have a canonical id method that defaults to *ID_FIELD_NAME)

rbalicki2 (changes requested):
Looks amazing! I'll re review once you move the canonical id source of truth to the server object entity itself. But overall I really love what I see!!

crates/isograph_schema/src/network_protocol.rs
lucasmachadorj: This is one point I wasn't sure, but at first I declared it in the NetworkProtocol trait due `isograph_schema` needs to call `get_id_field_name` but this module shouldn't be protocol dependent IIUC.

```#[derive(Debug, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
pub struct NameAndArguments {
    pub name: SelectableName,
    pub arguments: Vec<ArgumentKeyAndValue>,
}

impl NameAndArguments {
    pub fn normalization_key<TNetworkProtocol: NetworkProtocol>(
        &self,
        db: &IsographDatabase<TNetworkProtocol>,
        parent_object_entity_name: ServerObjectEntityName,
    ) -> NormalizationKey {
        if self.name == TNetworkProtocol::get_id_field_name(db, &parent_object_entity_name) {
            NormalizationKey::Id
        } else {
            NormalizationKey::ServerField(self.clone())
        }
    }
}```

Comment (rbalicki2) [reply crates/isograph_schema/src/network_protocol.rs:61]: Let's just do `get_id_field_name(db, parent_object_entity_name)`. Yeah, it doesn't need to be protocol-dependent

File: crates/graphql_network_protocol/src/graphql_network_protocol.rs [hunk 5]
```diff
@@ -234,6 +242,25 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
             {indent}}}"
         ))
     }
+
+    fn get_id_field_name(
+        db: &IsographDatabase<GraphQLNetworkProtocol>,
+        server_object_entity_name: &ServerObjectEntityName,
```
rbalicki2: Btw you should generally never write a function that takes a reference to a `Copy` struct (and any interned strings are Copy). (One exception is if you need to because you're implementing a trait and the trait requires &T, or if the struct is copy but really really big. Anyway, these aren't big), so let's accept `server_object_entity_name: ServerObjectEntityName`.

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -234,6 +242,25 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+    ) -> DiagnosticResult<ServerScalarSelectableName> {
+        let entity = server_object_entity_named(db, *server_object_entity_name)
+            .as_ref()
+            .map_err(|e| e.clone())?  // Propagate error instead of expect
```
rbalicki2: as_ref + map_err can be replaced by `.clone_err`, which is a function we defined on a trait to simplify this extremely common case

crates/isograph_schema/src/data_model/entities.rs
```diff
@@ -28,10 +32,20 @@ pub struct ServerObjectEntity<TNetworkProtocol: NetworkProtocol> {
     ///
     /// This is a GraphQL-ism! We should get rid of it.
     pub concrete_type: Option<ServerObjectEntityName>,
-
+    pub server_object_entity_directives: ServerObjectEntityDirectives,
     pub network_protocol_associated_data: TNetworkProtocol::SchemaObjectAssociatedData,
 }
 
+impl<TNetworkProtocol: NetworkProtocol> ServerObjectEntity<TNetworkProtocol> {
+    pub fn canonical_id_field_name(&self) -> ServerScalarSelectableName {
```
rbalicki2, line 40: nice, this is great

crates/artifact_content/src/imperatively_loaded_fields.rs
```diff
@@ -204,14 +209,16 @@ pub(crate) fn get_paths_and_contents_for_imperatively_loaded_field<
...
     reachable_variables: &BTreeSet<VariableName>,
     entrypoint: &OwnedClientSelectable<TNetworkProtocol>,
-) -> Vec<VariableDefinition<ServerEntityName>> {
+) -> DiagnosticResult<Vec<VariableDefinition<ServerEntityName>>> {
     reachable_variables
         .iter()
         .flat_map(|variable_name| {
             // HACK
-            if *variable_name == *ID_FIELD_NAME {
+            if *variable_name == TNetworkProtocol::get_id_field_name(db, entity_name).unwrap() {
```
rbalicki2: let's use expect instead of unwrap

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -234,6 +242,24 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
     }
+
+    fn get_id_field_name(
+        db: &IsographDatabase<GraphQLNetworkProtocol>,
+        server_object_entity_name: ServerObjectEntityName,
+    ) -> DiagnosticResult<ServerScalarSelectableName> {
+        let entity = server_object_entity_named(db, server_object_entity_name)
+            .clone_err()?
+            .as_ref()
+            .ok_or_else(|| {
```
rbalicki2, line 250: let's re-use one of the existing diagnostics, `entity_not_defined_diagnostic`... actually I'll land a change to have a method `server_object_entity_named_result`, which returns a `Result` containing exactly that diagnostic.

crates/graphql_network_protocol/src/process_type_system_definition.rs
```diff
@@ -87,6 +88,7 @@ pub fn process_graphql_type_system_document(
                     },
                     GraphQLObjectDefinitionType::Object,
                     &mut refetch_fields,
+                    None,
```
rbalicki2: What we're doing is awkward, but it's okay for now: we accumulate all the GraphQL directives into a struct called directives

```
        let (mut result, mut directives, mut refetch_fields) =
            process_graphql_type_system_document(
```

we later add to this for extensions. Then we attempt to parse this into a struct.

So, IMO, what we should do is to
- in graphql_network_protocol.rs, have a `HashMap<ServerObjectEntityName, ServerObjectEntity>`
- in `process_object_type_definition` when we create the `ServerObjectEntity`, we can create it with `None` (or `Default::default`) for the `ServerObjectEntityDirectives` field
- when we iterate over the directives, call

```
                    let server_object_entity_directives: ServerObjectEntityDirectives =
                        from_graphql_directives(&directives).map_err(|err| {
```

and then mutate the server object entity in the map, and assign the directive set.

I am sorry. This code is messier than I thought! I'm doing a decent amount of cleaning up of this code (broadly speaking), so a bunch of this will be cleaner soon. (Hopefully.)

crates/graphql_network_protocol/src/graphql_network_protocol.rs
lucasmachadorj: Ok, lemme know when you land it

crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -114,46 +114,51 @@ impl NetworkProtocol for GraphQLNetworkProtocol {
...
+            ServerObjectEntityName,
+            &mut ProcessObjectTypeDefinitionOutcome<Self>,
+        > = HashMap::new();
+
+        for item in result.iter_mut() {
+            if let Some(outcome) = item.as_ref_mut().as_object() {
+                entity_map.insert(outcome.server_object_entity.item.name, outcome);
+            }
+        }
+
```
lucasmachadorj, line 127: Reduces the O(n^2) loop complexity below to O(n)

PR #808, opened 2025-11-30
Throw graphql errors at component level 

PatrykWalach:
---
This is **part 2 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #808 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #927

### DIFF
--- a/libs/isograph-react/src/core/makeNetworkRequest.ts
+++ b/libs/isograph-react/src/core/makeNetworkRequest.ts
@@ -225,9 +225,6 @@ export function makeNetworkRequest<
         try {
           fetchOptions?.onError?.();
         } catch {}
-        throw new Error('Network response had errors', {
-          cause: networkResponse,
-        });
       }
 
       const root = { __link: ROOT_ID, __typename: artifact.concreteType };
@@ -256,7 +253,10 @@ export function makeNetworkRequest<
             environment,
             environment.store,
             normalizationAst.selections,
-            networkResponse.data ?? {},
+            {
+              data: networkResponse.data ?? undefined,
+              errors: networkResponse.errors ?? undefined,
+            },
             variables,
             root,
             encounteredIds,
@@ -417,7 +417,10 @@ function readDataForOnComplete<
       environment,
       fragment,
       fakeNetworkRequestOptions,
-    ).item;
+    );
+    if (fragmentResult.kind === 'Errors') {
+      return null;
+    }
     const readerArtifact = resolvedReaderWithRefetchQueries.readerArtifact;
     switch (readerArtifact.kind) {
       case 'ComponentReaderArtifact': {
@@ -445,7 +448,7 @@ function readDataForOnComplete<
       }
       case 'EagerReaderArtifact': {
         return readerArtifact.resolver({
-          data: fragmentResult,
+          data: fragmentResult.item,
           parameters: variables,
           ...(readerArtifact.hasUpdatable
             ? {
@@ -529,7 +532,10 @@ function makeOptimisticUpdate<
     environment,
     environment.store,
     artifact.networkRequestInfo.normalizationAst.selections,
-    optimisticNetworkResponse,
+    {
+      data: optimisticNetworkResponse,
+      errors: undefined,
+    },
     variables,
     root,
     encounteredIds,
--- a/libs/isograph-react/src/core/read.ts
+++ b/libs/isograph-react/src/core/read.ts
@@ -12,13 +12,15 @@ import type {
   ReaderWithRefetchQueries,
   RefetchQueryNormalizationArtifactWrapper,
 } from './entrypoint';
+import type { ReadFieldErrorPath } from './errors';
+import { readDataWithErrors, ReadFieldAggregateError } from './errors';
 import type {
   ExtractData,
   FragmentReference,
   UnknownTReadFromStore,
   Variables,
 } from './FragmentReference';
-import type { IsographEnvironment } from './IsographEnvironment';
+import type { IsographEnvironment, StoreError } from './IsographEnvironment';
 import {
   assertLink,
   getOrLoadIsographArtifact,
@@ -27,9 +29,11 @@ import {
   type DataTypeValue,
   type StoreLink,
   type StoreRecord,
+  type WithErrors,
 } from './IsographEnvironment';
 import { logMessage } from './logging';
 import { maybeMakeNetworkRequest } from './makeNetworkRequest';
+import { isNonEmptyArray, type NonEmptyArray } from './NonEmptyArray';
 import { getStoreRecordProxy } from './optimisticProxy';
 import type { PromiseWrapper } from './PromiseWrapper';
 import {
@@ -49,12 +53,26 @@ import type {
   ReaderScalarField,
 } from './reader';
 import { getOrCreateCachedStartUpdate } from './startUpdate';
-import type { Arguments } from './util';
+import { isArray, type Arguments } from './util';
 
-export type WithEncounteredRecords<T> = {
-  readonly encounteredRecords: EncounteredIds;
-  readonly item: ExtractData<T>;
+export type WithEncounteredRecords<T> =
+  | {
+      readonly kind: 'Success';
+      readonly encounteredRecords: EncounteredIds;
+      readonly item: ExtractData<T>;
+    }
+  | {
+      readonly kind: 'Errors';
+      readonly encounteredRecords: EncounteredIds;
+      readonly errors: ReadFieldAggregateError;
+    };
+
+export type WithReadFieldErrorPath<T> = {
+  readonly path: readonly ReadFieldErrorPath[];
+  readonly errors: T;
 };
+
+export type ReadFieldErrors = WithReadFieldErrorPath<NonEmptyArray<StoreError>>;
 
 export function readButDoNotEvaluate<
   TReadFromStore extends UnknownTReadFromStore,
@@ -79,6 +97,7 @@ export function readButDoNotEvaluate<
     fragmentReference.networkRequest,
     networkRequestOptions,
     mutableEncounteredRecords,
+    [],
   );
 
   logMessage(environment, () => ({
@@ -119,27 +138,36 @@ export function readButDoNotEvaluate<
       });
     }
     throw onNextChangeToRecord(environment, response.recordLink);
+  } else if (response.item.kind === 'Errors') {
+    return {
+      kind: 'Errors',
+      encounteredRecords: mutableEncounteredRecords,
+      errors: new ReadFieldAggregateError(response.item.errors),
+    };
   } else {
     return {
+      kind: 'Success',
       encounteredRecords: mutableEncounteredRecords,
-      item: response.data,
+      item: response.item.value,
     };
   }
 }
 
-export type ReadDataResultSuccess<Data> = {
+export type ReadDataResultSuccess<Item> = {
   readonly kind: 'Success';
-  readonly data: Data;
+  readonly item: Item;
 };
 
-export type ReadDataResult<Data> =
-  | ReadDataResultSuccess<Data>
-  | {
-      readonly kind: 'MissingData';
-      readonly reason: string;
-      readonly nestedReason?: ReadDataResult<unknown>;
-      readonly recordLink: StoreLink;
-    };
+type ReadDataResultMissingData = {
+  readonly kind: 'MissingData';
+  readonly reason: string;
+  readonly nestedReason?: ReadDataResultMissingData;
+  readonly recordLink: StoreLink;
+};
+
+export type ReadDataResult<Item> =
+  | ReadDataResultSuccess<Item>
+  | ReadDataResultMissingData;
 
 function readData<TReadFromStore>(
   environment: IsographEnvironment,
@@ -150,7 +178,8 @@ function readData<TReadFromStore>(
   networkRequest: PromiseWrapper<void, any>,
   networkRequestOptions: NetworkRequestReaderOptions,
   mutableEncounteredRecords: EncounteredIds,
-): ReadDataResult<ExtractData<TReadFromStore>> {
+  path: ReadFieldErrorPath[],
+): ReadDataResult<WithErrors<ExtractData<TReadFromStore>, ReadFieldErrors>> {
   const encounteredIds = insertEmptySetIfMissing(
     mutableEncounteredRecords,
     root.__typename,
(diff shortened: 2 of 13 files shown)


rbalicki2:
holy shit, this is great. I had no idea this was on your roadmap. Will review (though maybe not today)

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -121,18 +124,46 @@ export type DataTypeValue =
...
 
+export type WithErrorsData<T extends DataTypeValue> = {
+  readonly kind: 'Data';
+  readonly value: T;
+  /**
+   * A field can be null with errors and have a value depending on it's nested fields
+   * 1. if a nested field is non null and has a error it bubbles up to this parent field
+   * 2. in another query which is not selecting this nested field the parent field can resolve correctly
+   */
+  readonly errors: PayloadErrors | undefined;
```
PatrykWalach: The comment is incorrect. We basically o it this way because that's how relay does it but this may not be correct. They delete errors only when the response is a non-error null but I don't know the exact reasons for that, this might be for optimisation reasons. Removing this should simplify `read` 🤔 .

If we have query
`node {id name}`
and response `node: null`

then fire another query
`node {id}`
and response `node:{id:1}`
The first query will get stuck in loading state.

We should somehow handle this case in generic way, not only when we have errors

libs/isograph-react/src/core/read.ts
```diff
@@ -122,23 +124,27 @@ export function readButDoNotEvaluate<
...
+
+type ReadDataResultMissingData = {
+  readonly kind: 'MissingData';
+  readonly reason: string;
+  readonly nestedReason?: ReadDataResultMissingData;
+  readonly recordLink: StoreLink;
 };
 
 export type ReadDataResult<Data> =
   | ReadDataResultSuccess<Data>
```
PatrykWalach: This is correct for an record (`readData`), we either miss data or we return object with errors aggregated from the records fields.

For `readScalar`, `readClientField` etc. This is incorrect, they are semantically nullable we either have missing data, success with no errors or null with errors. We should make it a new type - this will make sure we aggregate errors in a typesafe way.

Right now there's a high chance I forgot to append errors somewhere

libs/isograph-react/src/core/IsographEnvironment.ts
PatrykWalach: removed `errors`

libs/isograph-react/src/core/read.ts
PatrykWalach: added `ReadFieldResult`

libs/isograph-react/src/core/areEqualWithDeepComparison.ts
```diff
@@ -135,3 +131,52 @@ export function mergeObjectsUsingReaderAst(
 
   return canRecycle ? oldItemObject : newItemObject;
 }
+
+export function mergeArrays<T extends unknown[]>(oldArray: T, newArray: T): T {
+  if (newArray.length !== oldArray.length) {
+    return newArray;
```
rbalicki2: should we attempt to reuse objects within oldArray if they are the same as those in newArray?

e.g. in `new: [{ foo: 'bar' }]` and `old [{ foo: 'bar' }, 123]`, we should reuse the object `{foo: 'bar'}`, i.e. we might want to do `for (let i = 0; i < Math.min(newArray.length, oldArray.length); i++) { mergeValues(...) }` and copy over everything from `Math.min(newArray.length, oldArray.length)` to `newArray.length`

libs/isograph-react/src/core/areEqualWithDeepComparison.ts
```diff
@@ -135,3 +131,52 @@ export function mergeObjectsUsingReaderAst(
...
+    if (mergedItem !== oldArray[i]) {
+      canRecycle = false;
+    } else {
+      newArray[i] = mergedItem;
+    }
+  }
+  return canRecycle ? oldArray : newArray;
+}
+function mergeValues<T>(oldObject: T, newObject: T): T {
+  if (newObject == null || oldObject == null) {
```
rbalicki2: why are we treating `==` and `undefined` the same? (Maybe this is clear from the rest of the PR)

libs/isograph-react/src/core/areEqualWithDeepComparison.ts
```diff
@@ -135,3 +131,52 @@ export function mergeObjectsUsingReaderAst(
...
+      return newObject;
+    }
+    return mergeArrays(oldObject, newObject);
+  }
+  if (typeof newObject === 'object') {
+    if (typeof oldObject !== 'object') {
+      return newObject;
+    }
+    if (Object.keys(oldObject).length !== Object.keys(newObject).length) {
+      return newObject;
```
rbalicki2: like with arrays, we should probably reuse the parts of `newObject` here with overlapping keys even if we have a different number of keys

libs/isograph-react/src/core/areEqualWithDeepComparison.ts
```diff
@@ -135,3 +131,52 @@ export function mergeObjectsUsingReaderAst(
 
   return canRecycle ? oldItemObject : newItemObject;
 }
+
+export function mergeArrays<T extends unknown[]>(oldArray: T, newArray: T): T {
```
rbalicki2: nit: can we either name this or add a comment to explain that we're doing this without the reader AST

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -121,18 +124,40 @@ export type DataTypeValue =
...
+export interface PayloadErrorExtensions {}
+export type PayloadErrorPath = string | number;
+export type PayloadError = {
+  readonly message: string;
+  readonly locations?: { readonly line: number; readonly column: number }[];
+  readonly path?: PayloadErrorPath[];
+  readonly extensions?: PayloadErrorExtensions;
+};
+export type PayloadErrors = [PayloadError, ...PayloadError[]];
+
```
rbalicki2: neat trick, TIL

libs/isograph-react/src/core/areEqualWithDeepComparison.ts
rbalicki2: FWIW I am not 1000% convinced we need reuse of data for errors... still thinking about it.

libs/isograph-react/src/core/areEqualWithDeepComparison.ts
PatrykWalach: This a refactor, the code is equivalent with the previous one

libs/isograph-react/src/core/areEqualWithDeepComparison.ts
PatrykWalach: I guess we immediately throw them so there's no opportunity to compare the errors

libs/isograph-react/src/core/IsographEnvironment.ts
PatrykWalach: this is quite useful, taken from `fp-ts` and `effect`
```tsx
type NonEmptyArray<T> = [T, ...T[]]

function isNonEmptyArray<T>(arr: T[]): arr is NonEmptyArray<T> {
  return arr.length !== 0;
}
```

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -121,18 +124,40 @@ export type DataTypeValue =
...
 export type DataId = string;
 
 export const ROOT_ID: DataId & '__ROOT' = '__ROOT';
 
+export interface PayloadErrorExtensions {}
+export type PayloadErrorPath = string | number;
+export type PayloadError = {
+  readonly message: string;
+  readonly locations?: { readonly line: number; readonly column: number }[];
+  readonly path?: PayloadErrorPath[];
```
PatrykWalach: We should map path when reading e.g. ["foo___...", "bar___....", "baz___...."] - this path is from root of the query
we read fragment from "foo" so we want to return: ["bar", "baz"].
We can just track the path while we read.

The issue is once we hit null, e.g. "bar" is null, we don't know how to map rest of the path "baz___..." into "baz".


I think we need to map path from `NetworkResponseKey[]` into `ParentRecordKey[]` then we can map it while reading

libs/isograph-react/src/core/IsographEnvironment.ts
rbalicki2: Let's do this (i.e. explicitly name it `NonEmptyArray`) for clarity

PatrykWalach: done

libs/isograph-react/src/tests/normalizeData.test.ts
```diff
@@ -188,6 +216,735 @@ describe('nested Query', () => {
...
+        { __link: ROOT_ID, __typename: errorsEntrypoint.concreteType },
+        new Map(),
+      );
+      expect(store).toStrictEqual<BaseStoreLayerData>({
+        Query: {
+          [ROOT_ID]: {
+            node____id___1: err([
+              {
+                message: 'Not found',
+                path: ['node____id___v_id'],
```
rbalicki2: 1. do we need to store the path at all? 2. if we do, do we need to store it for each error?

----------

I don't think we should expose the path + error message to the user at this stage. Both are too brittle, and it's a mistake to rely on either of them. Innocuous changes (like renaming `viewer -> loggedInUser`) will cause unrelated code to silently stop working.

Right now, the behavior is to unwrap all errors in a fragment. Good! IMO, in the future, I think we should let folks opt out of that (how to give that good DevEx TBD) by allowing them to handle errors at the read level: e.g. `node @allowError { ... }` in the param type would now have `{ kind: 'Error', message: string } | { kind: 'Data', data: ... }`. 

Analogously, we also automatically unwrap all missing fields via throwing a promise.

------

So, given that I don't think it's likely that we'll ever directly expose the path, can we avoid storing it?

(Error message I feel less strongly about, but I also don't think that the error boundary should necessarily see it.)

libs/isograph-react/src/tests/normalizeData.test.ts
```diff
@@ -188,6 +216,735 @@ describe('nested Query', () => {
...
+        nicknameErrorsEntrypoint.networkRequestInfo.normalizationAst.selections,
+        {
+          data: {
+            node____id___v_id: {
+              __typename: 'Economist',
+              id: '1',
+              nickname: null,
+            },
+          },
+          errors: [
```
rbalicki2, line 380: A problem for the future: this is very graphql-specific. I am happy to go with this for now, but IMO, we should require that network/raw response types are of the form: `node: { kind: 'Ok', ... } | { kind: 'Error', message }`, and we should provide a function that transforms the graphql network response shape to the above **before** it is passed to `normalizeData`.

------

That being said, we should also optimize for the common case where there are no errors and allow you to pass fully-unwrapped data to `normalizeData`.

But that's an optimization for the future.

libs/isograph-react/src/tests/normalizeData.test.ts
rbalicki2: And as another note, there are multiple paths to the same object (e.g. node or viewer), so the path should not be exposed except perhaps as part of a fragment, and can't be stored on the store item.

libs/isograph-react/src/tests/normalizeData.test.ts
```diff
@@ -188,6 +216,735 @@ describe('nested Query', () => {
...
+            node____id___1: ok({
+              __typename: 'Economist',
+              __link: '1',
+            }),
+          },
+        },
+      });
+    });
+
+    test('keeps nested errors', () => {
```
rbalicki2: the test description doesn't seem accurate?

libs/isograph-react/src/tests/normalizeData.test.ts
```diff
@@ -188,6 +216,735 @@ describe('nested Query', () => {
...
+          '1': {
+            __typename: ok('Economist'),
+            id: ok('1'),
+            name: ok('Bob'),
+          },
+        },
+      });
+    });
+
+    test('keeps unrelated errors', () => {
```
rbalicki2: likewise, I don't understand this test description

libs/isograph-react/src/tests/normalizeData.test.ts
```diff
@@ -188,6 +216,735 @@ describe('nested Query', () => {
           },
         },
       },
+      errors: undefined,
     } satisfies WithEncounteredRecords<Query__subquery__param>);
   });
 });
+
+export const errors = iso(`
```
rbalicki2, line 229: could we rename this + errorsEntrypoint + the others to be more descriptive? e.g. `errorsEntrypoint` isn't quite accurate... it's only related to errors because we write errors into the store. And unrelated to errors if we don't. And that's true of all entrypoints!

libs/isograph-react/src/tests/normalizeData.test.ts
```diff
@@ -188,6 +216,735 @@ describe('nested Query', () => {
...
+
+      const data = readButDoNotEvaluate(environment, item, {
+        suspendIfInFlight: true,
+        throwOnNetworkError: false,
+      });
+
+      expect(data).toStrictEqual<WithEncounteredRecords<Query__errors__param>>({
+        encounteredRecords: new Map([['Query', new Set([ROOT_ID])]]),
+        item: {
+          node: null,
```
rbalicki2: I don't like GraphQL's error behavior, and don't think we should emulate it. In other words, if `node` is null because there was an error, I think we should nonetheless never let the user see `node: null`. Even if they're calling `readData` directly.

--------

Instead, I think that if a field has an error, we should throw when we call `useResult` or whatnot. So IMO `readData` should have return type:

```
type ReadDataResult = ReadDataResultMissingData | ReadDataSuccess

type ReadDataSuccess = {
  readonly kind: 'Success';
  item: ReadDataDataResult | ReadDataErrorsResult
}

type ReadDataDataResult = { kind: 'Success', item: ... }
type ReadDataErrorsResult = { kind: 'Errors', errors: ... }
```

Note the two-layer structure: first we check whether something has missing data, and then we check whether something errored out or not. I think that's better than flattening it.

libs/isograph-react/src/react/useReadAndSubscribe.ts
```diff
@@ -35,6 +36,13 @@ export function useReadAndSubscribe<
     setReadOutDataAndRecords,
     readerAst,
   );
+
+  if (readOutDataAndRecords.errors) {
+    throw new GraphqlAggregateError(
+      readOutDataAndRecords.errors.map((error) => new GraphqlError(error)),
```
rbalicki2: It might be nice to not create new errors on every render, but instead to have stable references here.

libs/isograph-react/src/tests/normalizeData.test.ts
PatrykWalach: Yeah, this tests is not relevant anymore

rbalicki2:
Perhaps somewhat unrelatedly, I wonder if a NetworkProtocol can either support per-field errors or not. Not sure

libs/isograph-react/src/react/useReadAndSubscribe.ts
PatrykWalach: done

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -356,7 +356,10 @@ function readDataForOnComplete<
       environment,
       fragment,
       fakeNetworkRequestOptions,
-    ).item;
+    );
+    if (fragmentResult.kind === 'Errors') {
+      return null;
```
PatrykWalach: we call `onComplete` with `null`

libs/isograph-react/src/tests/normalizeData.test.ts
PatrykWalach: deleted

libs/isograph-react/src/tests/normalizeData.test.ts
PatrykWalach: Done, this made a couple of things much clearer

libs/isograph-react/src/tests/normalizeData.test.ts
PatrykWalach: deleted the test

libs/isograph-react/src/tests/normalizeData.test.ts
PatrykWalach: ok, I don't store path and message

compute the path when reading, so e.g. you know where to `@catch`

PR #809, opened 2025-11-30
handle whitespace before and after iso literal

PatrykWalach:
if entrypoint literal is long, prettier will format it, making it not recognizable by the compiler

```tsx
// prettier in
const normalizeUndefinedFieldEntrypoint = iso(`entrypoint Query.normalizeUndefinedField`);


// out
const normalizeUndefinedFieldEntrypoint = iso(
  `entrypoint Query.normalizeUndefinedField`,
);
```


rbalicki2 (approved):
LGTM, thank you for catching this!!

PR #810, opened 2025-11-30
optimize intern fn

ch1ffa:
We optimise `intern` fn the same way we did for `inter_ref`, but it is simpler as we don't need to handle a case when `derived_node` exists besides its registration.

### DIFF
--- a/crates/pico/src/database.rs
+++ b/crates/pico/src/database.rs
@@ -413,7 +413,15 @@ pub fn intern<Db: Database, T: Clone + Hash + DynEq + 'static>(db: &Db, value: T
 
             current_epoch
         }
-        Entry::Occupied(occupied) => occupied.get().time_updated,
+        Entry::Occupied(mut occupied) => {
+            let revision = occupied.get_mut();
+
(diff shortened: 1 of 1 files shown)


crates/pico/src/database.rs
```diff
@@ -382,8 +381,47 @@ impl<Db: Database> Default for Storage<Db> {
...
+            vacant.insert(DerivedNodeRevision {
+                time_updated: current_epoch,
+                time_verified: current_epoch,
+                node_index,
+                dependency_index,
+            });
+
+            current_epoch
+        }
+        Entry::Occupied(occupied) => occupied.get().time_updated,
```
rbalicki2: do we not need to set `time_verified = current_epoch`?

rbalicki2 (approved):
Amazing, just one question, otherwise lgtm!!

crates/pico/src/database.rs
ch1ffa: yeah! good catch, forgot about it

ch1ffa: done

Issue #820, opened 2025-12-03
Todo: validate that all arguments point to types that exist


rbalicki2:
This was lost in refactoring :/

  title(foo: Foo): String!

does not cause us to fail to compile

PR #821, opened 2025-12-04
The network protocol should define at most one selectable per (parent_name, selectable_name)

rbalicki2:


rbalicki2:
Closes #820

PR #827, opened 2025-12-06
replace `Array.isArray` with `isArray`

PatrykWalach:
Array.isArray narrows to any[] which can be unsafe, also it doesn't narrow if array is readonly


rbalicki2:
awesome!! Thanks for doing this, and for making a small PR for it

PR #831, opened 2025-12-07
Add strict mode note for TanStack Start

corydeppen:


rbalicki2:
Amazing! Thank you for adding this

PR #836, opened 2025-12-09
make `writeData` return `ItemCleanupPair<FragmentReference>`

### LINKED ISSUE #785: writeData should return ItemCleanupPair<FragmentReference>; the fragment reference should be retained

PatrykWalach:
Closes #785

### DIFF
--- a/libs/isograph-react/src/core/cache.ts
+++ b/libs/isograph-react/src/core/cache.ts
@@ -935,7 +935,7 @@ export function writeData<
   environment: IsographEnvironment,
   entrypoint: IsographEntrypoint<
     TReadFromStore,
-    unknown,
+    TClientFieldValue,
     NormalizationAst,
     TRawResponseType
   >,


docs-website/docs/external-data-sources.md
```diff
@@ -43,63 +47,62 @@ Let's modify this example to integrate with another data source.
...
+} from '@isograph/react-disposable-state';
 
-type State = { kind: 'NotLoaded' } | { kind: 'Loading' } | { kind: 'Loaded' };
+type State = { kind: 'NotLoaded' } | { kind: 'Loading' };
 
 export const Button = iso(`
   field Query.PreloadUserDetailPageButton @component {}`)(({ data }) => {
   const [state, setState] = useState<State>({ kind: 'NotLoaded' });
   const environment = useIsographEnvironment();
+  const { state: fragmentReference, setState: setFragmentReference } =
```
rbalicki2, line 61: in the future, we should rewrite this to have a single call to `useUpdatableDisposableState` instead of two pieces of state, where impossible states are representable. I think it should be doable

rbalicki2 (approved):
YESSSS

docs-website/docs/external-data-sources.md
PatrykWalach: If I was implementing this I'd use a startTransition around make network request

PatrykWalach: The <Suspense in the example will never suspend though. if we accept raw response data wrapped in a WrappedPromise, then we can eliminate the useState

rbalicki2: oh yeah good catch

PR #843, opened 2025-12-11
add optimisticNetworkResponse

PatrykWalach:


libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach, line 302: Not sure if I have to revert and dispose here

rbalicki2: yeah, we should

libs/isograph-react/src/core/read.ts
```diff
@@ -316,7 +316,7 @@ export function readLoadablySelectedFieldData(
     data: (
       args: any,
       // TODO get the associated type for FetchOptions from the loadably selected field
-      fetchOptions?: FetchOptions<any>,
+      fetchOptions?: FetchOptions<any, never>,
```
rbalicki2, line 319: interesting that this type checks... I would think that since in `makeNetworkRequest` etc `TRawResponseType extends NetworkResponseObject`, we would not be able to pass this fetchOptions. I guess never depends weirdly.

Is there any reason we can't support optimistic responses here? (and in other places where we have `FetchOptions<never>`)

libs/isograph-react/src/core/reader.ts
```diff
@@ -177,5 +177,5 @@ export type LoadableField<
   // user-facing API. Users should only interact with LoadableFields via APIs
   // like useClientSideDefer. These APIs should have a nullable fetchOptions
   // parameter, and provide a default value ({}) to the LoadableField.
-  fetchOptions: FetchOptions<TResult>,
+  fetchOptions: FetchOptions<TResult, never>,
```
rbalicki2, line 180: e.g. here, we could allow you to pass a raw response (as ugly as it is) to `LoadableFieldRenderer`

libs/isograph-react/src/core/read.ts
PatrykWalach: this is because I didn't generate raw response type for imperative fields

rbalicki2: okay sgtm

PatrykWalach: The `LoadableField` needs generic for RawResponseType

rbalicki2: ah yeah, okay, we can punt on that for now then

rbalicki2: #844

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: done

rbalicki2: sweet, okay, will eat and then do a deeper review, but 🙌  so good

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -130,19 +138,39 @@ export function retainQueryWithoutMakingNetworkRequest<
         TRawResponseType
       >,
   variables: ExtractParameters<TReadFromStore>,
+  fetchOptions: FetchOptions<TClientFieldValue, TRawResponseType> | null,
 ): ItemCleanupPair<PromiseWrapper<void, AnyError>> {
   let status: NetworkRequestStatus = {
```
rbalicki2: `status` can only be `Disposed | UndisposedIncomplete`, can we reflect that in the types here, instead of having the third, unreachable type (`UndisposedComplete`)?

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -218,26 +255,40 @@ export function makeNetworkRequest<
...
-          kind: 'AfterNormalization',
-          store: environment.store,
-          encounteredIds: encounteredIds,
-        }));
+      if (status.kind === 'UndisposedIncomplete') {
+        if (status.optimistic != null) {
+          revertOptimisticStoreLayerAndMaybeReplace(
+            environment,
+            status.optimistic,
+            null,
```
rbalicki2: I don't think we should pass `null` here, right?

In addition, we should still call `logMessage('AfterNormalization')` and `callSubscriptions` in this branch

libs/isograph-react/src/core/makeNetworkRequest.ts
rbalicki2: I don't see where we are removing the optimistic layer

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -171,18 +199,27 @@ export function makeNetworkRequest<
   readerWithRefetchQueries: PromiseWrapper<
     ReaderWithRefetchQueries<TReadFromStore, TClientFieldValue>
   > | null,
-  fetchOptions: FetchOptions<TClientFieldValue> | null,
+  fetchOptions: FetchOptions<TClientFieldValue, TRawResponseType> | null,
 ): ItemCleanupPair<PromiseWrapper<void, AnyError>> {
```
rbalicki2, line 181: let's later take a pass at cleaning this up. I think it would be clearer to have functions that transition us between the states:
- getUndisposedCompleteStatus(oldStatus)
- getDisposedStatus(oldStatus)

That will give us a nice way to ensure that e.g. whenever we transition away from `UndisposedIncomplete` we revert the optimistic update

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -438,9 +506,62 @@ function fetchNormalizationAstAndRetainArtifact<
...
+    | RefetchQueryNormalizationArtifact
+    | IsographEntrypoint<
+        TReadFromStore,
+        TClientFieldValue,
+        TNormalizationAst,
+        TRawResponseType
+      >,
+  variables: ExtractParameters<TReadFromStore>,
+  optimisticNetworkResponse: TRawResponseType,
+): OptimisticStoreLayer {
```
rbalicki2, line 521: iiiinteresting, we don't actually have a single way to dispose of this (or rather, disposal requires an additional param). If this is a widely used pattern, maybe `ItemCleanupPairWithParamForDestructor` makes sense...

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -438,9 +506,62 @@ function fetchNormalizationAstAndRetainArtifact<
...
+  const root = { __link: ROOT_ID, __typename: artifact.concreteType };
+
+  if (
+    artifact.networkRequestInfo.normalizationAst.kind ===
+    'NormalizationAstLoader'
+  ) {
+    throw new Error(
```
rbalicki2, line 528: no need to do it now, but as a cleanup, we should move this check out of this function, and instead only accept `TNormalizationAst extends NormalizationAst`. https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -438,9 +506,62 @@ function fetchNormalizationAstAndRetainArtifact<
...
+    environment.store,
+    artifact.networkRequestInfo.normalizationAst.selections,
+    optimisticNetworkResponse,
+    variables,
+    root,
+    encounteredIds,
+  );
+
+  logMessage(environment, () => ({
+    kind: 'AfterNormalization',
```
rbalicki2, line 546: Hm... I think we should probably have two separate events: `WroteOptimisticUpdate` and `RevertedOptimisticUpdate`

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: I failed to push, now it's done

rbalicki2: we know computer

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: The plan was to do the types but I forgot, but the error has to be here, because types are difficult -_-

PatrykWalach: done

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: `revertOptimisticStoreLayerAndMaybeReplace` calls `callSubscriptions`

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: We have three states:
- `UndisposedIncomplete` - initial
- `UndisposedComplete` - the optimistic is reverted but the query is retained
- `Disposed` - the query is unretained

I think if `throwOnNetworkError` is false, and we have enough data we will render so we don't want to unretain the data in the `catch` clause.

PatrykWalach: Nevermind, here it's only two

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: if the param is optional instead of null, then it works as expected, right?

rbalicki2: Just to clarify, I meant for this to be a random thought, not a request for changes

That's a good point, too, maybe destructors can be generic over `T` which can be optional if `T` includes undefined

libs/isograph-react/src/core/makeNetworkRequest.ts
PatrykWalach: done

libs/isograph-react/src/core/check.ts
```diff
@@ -15,15 +15,21 @@ export type RequiredShouldFetch = 'Yes' | 'No';
...
+  extends FetchOptionsShared<TReadOutData> {
+  shouldFetch?: ShouldFetch;
+  optimisticNetworkResponse?: TRawResponseType;
+}
+
+export interface RequiredFetchOptions<TReadOutData>
+  extends FetchOptionsShared<TReadOutData> {
   shouldFetch: RequiredShouldFetch;
-} & FetchOptions<TReadOutData>;
+}
```
rbalicki2, line 32: nice

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -130,24 +139,40 @@ export function retainQueryWithoutMakingNetworkRequest<
...
+            environment,
+            artifact,
+            variables,
+            fetchOptions.optimisticNetworkResponse,
+          )
+        : null,
   };
   return [
     wrapResolvedValue(undefined),
     () => {
```
rbalicki2, line 153: hm... again, just thinking out loud... maybe we should wrap all of the dispose functions in a `callOnce(callback)` so that the disposal functions don't need to worry about being called twice. Then we can add something like emitting warnings (in dev?) if the disposal is called multiple times

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -276,27 +344,41 @@ export function makeNetworkRequest<
...
-        unretainAndGarbageCollect(environment, status);
+      if (status.kind === 'UndisposedIncomplete') {
+        status =
+          revertOptimisticStoreLayerAndMaybeReplaceIfUndisposedIncomplete(
+            environment,
+            status,
+            null,
+          );
+      }
+      if (status.kind !== 'Disposed') {
```
rbalicki2, line 336: this is elegant

rbalicki2 (approved):
🎉  AMAZING 🎉

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -130,24 +139,40 @@ export function retainQueryWithoutMakingNetworkRequest<
...
+  let status:
+    | NetworkRequestStatusUndisposedIncomplete
+    | NetworkRequestStatusDisposed = {
+    kind: 'UndisposedIncomplete',
     retainedQuery: fetchNormalizationAstAndRetainArtifact(
       environment,
       artifact,
       variables,
     ),
+    optimistic:
```
rbalicki2: oops, we probably shouldn't write the optimistic update here, right?

rbalicki2: I think so... will push a fix and land

PR #852, opened 2025-12-12
feat: add `PromiseWrapper` utilities

PatrykWalach:
These should help with  handling sync`networkFunction`

### DIFF
--- a/libs/isograph-react/src/core/PromiseWrapper.ts
+++ b/libs/isograph-react/src/core/PromiseWrapper.ts
@@ -128,13 +128,19 @@ type Unwrap<T> = T extends null | undefined
     ? V
     : T;
 
+type UnwrapErr<T> = T extends null | undefined
+  ? never
+  : T extends PromiseWrapper<infer _, infer E>
+    ? E
+    : never;
+
 export function all<T extends readonly unknown[] | []>(
   values: T,
 ): PromiseWrapper<
   {
     -readonly [P in keyof T]: Unwrap<T[P]>;
   },
-  Unwrap<T[number]>
+  UnwrapErr<T[keyof T]>
 > {
   const promise: Promise<any> = Promise.all(
     values.map((value) => {
--- a/libs/isograph-react/src/core/PromiseWrapper.ts
+++ b/libs/isograph-react/src/core/PromiseWrapper.ts
@@ -1,8 +1,5 @@
 export type AnyError = any;
 
-export const NOT_SET = Symbol('NOT_SET');
-export type NotSet = typeof NOT_SET;
-
 export type Result<T, E> =
   | {
       readonly kind: 'Ok';
@@ -18,22 +15,39 @@ export type Result<T, E> =
  * Before the promise is resolved, value becomes non-null.
  */
 export type PromiseWrapper<T, E = any> = {
-  readonly promise: Promise<Exclude<T, NotSet>>;
-  result: Result<Exclude<T, NotSet>, E> | NotSet;
+  readonly promise: Promise<T>;
+  result:
+    | Result<T, E>
+    | {
+        readonly kind: 'Pending';
+      };
 };
 
-export interface PromiseWrapperOk<T, E = any> extends PromiseWrapper<T, E> {
+export interface PromiseWrapperOk<T> extends PromiseWrapper<T, never> {
   result: {
     readonly kind: 'Ok';
-    readonly value: Exclude<T, NotSet>;
+    readonly value: T;
+  };
+}
+
+export interface PromiseWrapperErr<E> extends PromiseWrapper<never, E> {
+  result: {
+    readonly kind: 'Err';
+    readonly error: E;
   };
 }
 
 export function wrapPromise<T>(
-  promise: Promise<Exclude<T, NotSet>>,
+  promise: Promise<T>,
 ): PromiseWrapper<T, unknown> {
   // TODO confirm suspense works if the promise is already resolved.
-  const wrapper: PromiseWrapper<T, any> = { promise, result: NOT_SET };
+  const wrapper: PromiseWrapper<T, any> = {
+    promise,
+    result: {
+      kind: 'Pending',
+    },
+  };
+
   promise
     .then((v) => {
       wrapper.result = { kind: 'Ok', value: v };
@@ -44,9 +58,7 @@ export function wrapPromise<T>(
   return wrapper;
 }
 
-export function wrapResolvedValue<T>(
-  value: Exclude<T, NotSet>,
-): PromiseWrapperOk<T, never> {
+export function wrapResolvedValue<T>(value: T): PromiseWrapperOk<T> {
   return {
     promise: Promise.resolve(value),
     result: {
@@ -56,15 +68,25 @@ export function wrapResolvedValue<T>(
   };
 }
 
+export function wrapRejectedValue<E>(error: E): PromiseWrapperErr<E> {
+  const promise = Promise.reject(error);
+  promise.catch(() => {});
+  return {
+    promise: promise,
+    result: {
+      kind: 'Err',
+      error,
+    },
+  };
+}
+
 export function readPromise<T, E>(p: PromiseWrapper<T, E>): T {
   const { result } = p;
-  if (result !== NOT_SET) {
-    const resultKind = result;
-    if (resultKind.kind === 'Ok') {
-      return resultKind.value;
-    } else {
-      throw resultKind.error;
-    }
+
+  if (result.kind === 'Ok') {
+    return result.value;
+  } else if (result.kind === 'Err') {
+    throw result.error;
   }
 
   throw p.promise;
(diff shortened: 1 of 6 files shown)


rbalicki2:
nice! I'll take a look this weekend

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -56,6 +61,18 @@ export function wrapResolvedValue<T>(
   };
 }
 
+export function wrapRejectedValue<E>(error: E): PromiseWrapperErr<never, E> {
+  const promise = Promise.reject(error);
+  promise.catch(() => {});
```
rbalicki2, line 73: why are we catching with a no-op here? (we also do it elsewhere)

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -18,19 +18,26 @@ export type Result<T, E> =
...
+export interface PromiseWrapperErr<T, E = any> extends PromiseWrapper<T, E> {
+  result: {
+    readonly kind: 'Err';
+    readonly error: E;
   };
 }
 
 export function wrapPromise<T>(
-  promise: Promise<Exclude<T, NotSet>>,
+  promise: Promise<T>,
```
rbalicki2, line 41: love the lack of `Exclude`

libs/isograph-react/src/core/makeNetworkRequest.ts
```diff
@@ -208,15 +210,18 @@ export function makeNetworkRequest<
...
-  const promise = Promise.all([
-    environment.networkFunction(
-      artifact.networkRequestInfo.operation,
-      variables,
-    ),
-    status.retainedQuery.normalizationAst.promise,
-    readerWithRefetchQueries?.promise,
-  ])
-    .then(([networkResponse, normalizationAst, readerWithRefetchQueries]) => {
+  const wrapper = PromiseWrapperThen(
```
rbalicki2: we are edging closer and closer to observables! 🎉

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -18,19 +18,26 @@ export type Result<T, E> =
  * Before the promise is resolved, value becomes non-null.
  */
 export type PromiseWrapper<T, E = any> = {
-  readonly promise: Promise<Exclude<T, NotSet>>;
-  result: Result<Exclude<T, NotSet>, E> | NotSet;
+  readonly promise: Promise<T>;
+  result: Result<T, E> | NotSet;
 };
 
 export interface PromiseWrapperOk<T, E = any> extends PromiseWrapper<T, E> {
```
rbalicki2: do we need `E = any` here? Can `PromiseWrapperOk` just have one type param? Same for `Err`

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -18,19 +18,26 @@ export type Result<T, E> =
...
+    readonly error: E;
   };
 }
 
 export function wrapPromise<T>(
-  promise: Promise<Exclude<T, NotSet>>,
+  promise: Promise<T>,
 ): PromiseWrapper<T, unknown> {
   // TODO confirm suspense works if the promise is already resolved.
   const wrapper: PromiseWrapper<T, any> = { promise, result: NOT_SET };
```
rbalicki2: Can we replace `NOT_SET` with `{ kind: 'NOT_SET' }`? Seems cleaner

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -88,3 +105,115 @@ export function getPromiseState<T, E>(
...
+  );
+}
+
+type Unwrap<T> = T extends null | undefined
+  ? T
+  : T extends PromiseWrapper<infer V>
+    ? V
+    : T;
+
+export function PromiseWrapperAll<T extends readonly unknown[] | []>(
```
rbalicki2: I am not a huge fan of starting this and `PromiseWrapperThen` with an uppercase letter. Thoughts?

Perhaps there can be an object (a class... but without a constructor ideally? Or a constructor that throws) `PromiseWrapper` that has static `then` and `all` methods on it. Perhaps also `wrapPromise` etc.

These functions are all tiny, so it doesn't seem like the end of the world if you bring in all of them instead of tree shaking.

rbalicki2: Separately, I wonder if we should just say that objects are better, and just support stuff of the form `PromiseWrapperAll({ ... })`

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -44,9 +51,7 @@ export function wrapPromise<T>(
   return wrapper;
 }
 
-export function wrapResolvedValue<T>(
-  value: Exclude<T, NotSet>,
-): PromiseWrapperOk<T, never> {
+export function wrapResolvedValue<T>(value: T): PromiseWrapperOk<T, never> {
   return {
     promise: Promise.resolve(value),
```
rbalicki2, line 63: do we need to actually have the resolved promise here, and the rejected promise in the err variant?

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -88,3 +105,115 @@ export function getPromiseState<T, E>(
...
+    ? V
+    : T;
+
+export function PromiseWrapperAll<T extends readonly unknown[] | []>(
+  values: T,
+): PromiseWrapper<
+  {
+    -readonly [P in keyof T]: Unwrap<T[P]>;
+  },
+  Unwrap<T[number]>
```
rbalicki2: do we need to `UnwrapErr` or something here?

libs/isograph-react/src/core/PromiseWrapper.ts
```diff
@@ -88,3 +105,115 @@ export function getPromiseState<T, E>(
...
+        state.promise.then(onfulfilled, onrejected),
+      ) satisfies PromiseWrapper<
+        TResult1 | TResult2,
+        unknown
+      > as PromiseWrapper<TResult1 | TResult2, never>;
+    case 'Err': {
+      const result = onrejected(state.error);
+
+      return {
+        promise: Promise.resolve(result),
```
rbalicki2: Hm... lets think about whether we want to emulate the behavior of promises. I'm not convinced! Promises suck! You can't have `Promise<Promise<T>>` for example. Perhaps we should implement the Rust Result methods?

rbalicki2 (commented):
Love it. Some nits. A bigger overall question of whether `PromiseWrapper` is just a worse version of https://github.com/facebook/relay/blob/0dea0b18c8585a75148f819feacf23f940a1628a/packages/relay-runtime/network/RelayObservable.js#L4, and maybe we should just adopt observables. (It's not! Observables can include streams! So the question is, will we eventually want to add streams of data?)

Otherwise, I think promises are garbage (e.g. no `Promise<Promise<T>>`), and `catch` has weird behavior. I think we definitely want to support `PromiseWrapper<PromiseWrapper<T>, PromiseWrapper<E>>`. So IMO it's worth taking the time to think about the API we want to support. So e.g. `then` is IMO worse than `map` and `mapErr`

And a last thought about whether these would be cleaner as methods classes. I imagine three classes `PromiseWrapperOk`, `Err` and `Pending` which all implement the same interface, and that interface has `map`, `mapErr` etc methods.

And a final last thought about whether we should model this as `MaybePresent<Result<....>>`.

Anyway, this is great, I'm super pumped that we're getting proper foundations

libs/isograph-react/src/core/PromiseWrapper.ts
rbalicki2: gemini has informed me: this is deliberate suppression of unhandled promise rejection warnings.

PatrykWalach: Exactly

libs/isograph-react/src/core/PromiseWrapper.ts
PatrykWalach: I also can use import * as PromiseWrapper

libs/isograph-react/src/core/PromiseWrapper.ts
PatrykWalach: I guess not, we should only use promise.then if it's a Pending promise, otherwise it's a smell

libs/isograph-react/src/core/PromiseWrapper.ts
PatrykWalach: I did `kind: "Pending"`

libs/isograph-react/src/core/PromiseWrapper.ts
PatrykWalach: we do need it so we can pass them to react, technically react supports PromiseLike, so we can use an object with `.then` instead

libs/isograph-react/src/core/PromiseWrapper.ts
PatrykWalach: done

libs/isograph-react/src/core/PromiseWrapper.ts
PatrykWalach: I split this into `mapOk` and `mapErr`, is that what you've meant?

rbalicki2: Yeah, let's do that, and let's have them return the same variants `then` can change a promise's variant, I don't want that. Instead, we should have other functions that allow us to change variants

libs/isograph-react/src/core/PromiseWrapper.ts
PatrykWalach: I like class because mapOk and mapErr can be chained, if we are exposing this to the user I'd make it a class to make all available methods visible

PR #858, opened 2025-12-14
enforce use of == with null and === with other values

PatrykWalach:


libs/isograph-react/src/core/optimisticProxy.ts
```diff
@@ -69,15 +72,16 @@ export function getMutableStoreRecordProxy(
...
-            return false;
-          }
-          if (storeRecord != null) {
+          if (storeRecord !== undefined) {
+            if (storeRecord == null) {
+              return false;
+            }
+
             const value = Reflect.has(storeRecord, propertyName);
-            if (value !== undefined) {
```
PatrykWalach, line 80: There was a bug here

rbalicki2: NICE

rbalicki2 (approved):
lgtm

PR #860, opened 2025-12-14
enable oxlint import/no-cycle

PatrykWalach:
lint for import cycles.

- moved `subscribe()` from `cache.ts -> subscribe.ts` 
- moved `writeData` from `cache.ts -> writeData.ts`
- moved `getOrCreateCacheForArtifact` from `cache.ts ->
  getOrCreateCacheForArtifact.ts`
- moved `stableCopy` and `isArray` from `cache.ts -> util.ts`
- moved `maybeUnwrapNetworkRequest` from `useResult ->
  maybeUnwrapNetworkRequest.ts`
- removed a nasty cycle `read -> componentCache -> useReadAndSubscribe
  -> read`, it required splitting `componentCache`, which removed core dependy on react  `core -> react -> core` which is part of #564

- updated a bunch of imports to use `type`

### DIFF
--- a/libs/isograph-react/src/core/componentCache.ts
+++ b/libs/isograph-react/src/core/componentCache.ts
@@ -1,13 +1,9 @@
-import { useReadAndSubscribe } from '../react/useReadAndSubscribe';
-import { maybeUnwrapNetworkRequest } from '../react/useResult';
 import {
   FragmentReference,
   stableIdForFragmentReference,
 } from './FragmentReference';
-import { IsographEnvironment } from './IsographEnvironment';
-import { logMessage } from './logging';
-import { readPromise } from './PromiseWrapper';
-import { NetworkRequestReaderOptions } from './read';
+import type { IsographEnvironment } from './IsographEnvironment';
+import type { NetworkRequestReaderOptions } from './read';
 import { createStartUpdate } from './startUpdate';
 
 export function getOrCreateCachedComponent(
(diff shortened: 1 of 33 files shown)


libs/isograph-react/src/core/componentCache.ts
```diff
@@ -24,41 +20,10 @@ export function getOrCreateCachedComponent(
...
-            : undefined,
-        },
-        additionalRuntimeProps,
-      );
-    }
-    const idString = `(type: ${fragmentReference.root.__typename}, id: ${fragmentReference.root.__link})`;
-    Component.displayName = `${fragmentReference.fieldName} ${idString} @component`;
-    return Component;
-  })());
+  ] ??= environment.componentFunction(
```
rbalicki2, line 23: since this is only ever a single function, why don't we just import `componentFunction` directly here? Why do we need to store it on the environment?

I am not particularly concerned about the implementation right now, because I anticipate this will change at some point (i.e. I expect `createIsographEnvironment` will be generated by the compiler, and will have different values for vue React Solid etc), just wondering what your thoughts are

libs/isograph-react/src/core/entrypoint.ts
```diff
@@ -5,8 +5,8 @@ import type {
   UnknownTReadFromStore,
 } from './FragmentReference';
 import type { ComponentOrFieldName, TypeName } from './IsographEnvironment';
-import { TopLevelReaderArtifact } from './reader';
-import { Arguments } from './util';
+import type { TopLevelReaderArtifact } from './reader';
+import type { Arguments } from './util';
```
rbalicki2, line 9: would be great if importing something as a value but using it as a type was linted against

libs/isograph-react/src/core/componentCache.ts
PatrykWalach: because then we import `componentFunction` from `react` into `core` we should only import from `core` into `react`

rbalicki2: Sgtm lets ship

libs/isograph-react/src/core/entrypoint.ts
PatrykWalach: we need to use typescript `verbatimModuleSyntax` which is only available in esm. This requires either:
- using a bundler to ship esm and commonjs 
- or shipping esm only, which is quite good these days, you can require esm in commonjs just fine in newer versions of node

PR #861, opened 2025-12-14
enable oxlint rule typescript/consistent-type-imports

PatrykWalach:
It turns out, there is a rule for that, which is great, because I don't think typescript can fix these on save


PatrykWalach:
I should have asked, it's configurable to use single import

PR #866, opened 2025-12-15
enable `typescript/switch-exhaustiveness-check`

PatrykWalach:
This is cool:
- we don't have to remember to add default case
- we can still add them, but I deleted them so we are shipping less code
- it actualy found a bug - we are not visiting fields from inline fragments during garbage collection https://github.com/isographlabs/isograph/actions/runs/20236250770/job/58091580947?pr=866


rbalicki2:
lfgggg I love this. Will review

rbalicki2 (approved):
shipittt

PR #889, opened 2025-12-21
Migrate lazy_static macro to std LazyLock

ecalifornica:
https://rust-lang.github.io/rust-clippy/rust-1.91.0/index.html#non_std_lazy_statics


ecalifornica:
Is this build demos CI failure a sign that I need to set up the front-end tests?

PR #905, opened 2025-12-22
Fix and deny clippy explicit_into_iter_loop

ecalifornica:
A little lint, as a treat.


relay-crates/graphql-cli/examples/print-source.rs
```diff
@@ -45,7 +45,7 @@ fn main() {
         111..111,
     ];
     let printer = SourcePrinter::default();
-    for test_case in test_cases.into_iter() {
+    for test_case in test_cases {
```
rbalicki2, line 48: many such cases

PR #912, opened 2025-12-29
Colorize text pointed to by carats

### LINKED ISSUE #911: Colorize text pointed to by carats
We currently colorize the carats when printing errors, but not the text pointed to by the carat. That would look nice, I think!

kartikmawani:
"Fixes #911 "
This will colorize the carat and the text pointed to by the carat both in red when printing errors

### DIFF
--- a/crates/common_lang_types/src/text_with_carats.rs
+++ b/crates/common_lang_types/src/text_with_carats.rs
@@ -182,6 +182,8 @@ fn text_with_carats_and_line_count_buffer_and_line_numbers(
     // - the carat line containing the end of the span and LINE_COUNT_BUFFER later lines
     // - everything in between
 
+    colored::control::unset_override();
+
     (
         output_lines[(first_line_with_span.saturating_sub(line_count_buffer + 1))
             ..(std::cmp::min(last_line_with_span + line_count_buffer, output_lines.len()))]
@@ -631,6 +633,7 @@ mod test {
             true,
         )
         .0;
+
         let expected = "01234567\u{1b}[91m8\u{1b}[0m\n        \u{1b}[91m^\u{1b}[0m\n\u{1b}[91m01\u{1b}[0m2345678\n\u{1b}[91m^\u{1b}[0m\u{1b}[91m^\u{1b}[0m       \n012345678\n012345678\n012345678";
 
         assert_eq!(output, expected);


rbalicki2:
Amazing!! Will review when I'm back at a computer. Thank you so much for doing this!!!

Could you ensure that cargo fmt results in no changes? (CI will probably catch this)

Also could you add some tests for this (perhaps just run the existing tests once without color, once with color, and have two sets of expected results)

kartikmawani:
> Amazing!! Will review when I'm back at a computer. Thank you so much for doing this!!!
> 
> Could you ensure that cargo fmt results in no changes? (CI will probably catch this)
> 
> Also could you add some tests for this (perhaps just run the existing tests once without color, once with color, and have two sets of expected results)

the CI checks are now passing  cargo fmt on the latest commit.would you be okay with a simpler unit test in text_with_carats.rs.I can just force color mode on for that specific test   and verify that the output string includes the correct color codes.

rbalicki2:
Yeah, sure, if you just want to enable color on some tests, that's fine. Let's keep both scenarios, though (color and non). But yeah, all I want is to have a test that ensures that we get the expected string (which includes color codes)

crates/common_lang_types/src/text_with_carats.rs
```diff
@@ -589,4 +605,21 @@ mod test {
...
+            None,
+            Span::new(31, 33),
+            3,
+            true,
+        )
+        .0;
+
+        colored::control::unset_override();
+
+        assert!(output.contains("\u{1b}[91m"));
```
rbalicki2: could you assert that it exactly matches the expected text, please?

crates/common_lang_types/src/text_with_carats.rs
```diff
@@ -589,4 +605,21 @@ mod test {
...
+        let output = text_with_carats_and_line_count_buffer_and_line_numbers(
+            &input_with_lines(10),
+            None,
+            Span::new(31, 33),
+            3,
+            true,
+        )
+        .0;
+
+        colored::control::unset_override();
```
rbalicki2: why do we need this? do tests not print in color? I would rather not have these lines, but if they're necessary, let's add an explanatory comment and move this into `text_with_carats_and_line_count_buffer_and_line_numbers`

crates/common_lang_types/src/text_with_carats.rs
```diff
@@ -116,19 +114,35 @@ fn text_with_carats_and_line_count_buffer_and_line_numbers(
...
+                    highlighted.bright_red().to_string()
+                } else {
+                    highlighted.to_string()
+                },
+                suffix
+            );
+            output_lines.push(colored_source);
             // a line may be entirely empty, due to containing only a \n. We probably want to avoid
             // printing an empty line underneath. This is weird and probably buggy!
+            let end_of_carats = end_idx as u32;
```
rbalicki2: instead of having two variables (`end_of_carats` and `end_idx`) which have the same value but different representations, can we have a single one, and just cast it as needed?

e.g. if `start_of_carats` became `let start_of_carats = (actual_span.start as usize).saturating_sub(start_of_line);` we probably could do with fewer variables and less casting.

I'm not concerned with the perf implications of usize vs u32 here, I just want this code to be understandable later

(and despite my previous comment about multiple PRs, feel free to make this change in this same PR)

crates/common_lang_types/src/text_with_carats.rs
```diff
@@ -116,19 +114,35 @@ fn text_with_carats_and_line_count_buffer_and_line_numbers(
         };
 
         if should_print_carats {
-            let start_of_carats = actual_span.start.saturating_sub(start_of_line as u32);
             let line_len = line_content.len() as u32;
-            let end_of_carats = std::cmp::min(
+            let start_of_carats = actual_span.start.saturating_sub(start_of_line as u32);
```
rbalicki2: nit: no worries here, but it's slightly easier to review if you don't re-order lines like this. Before submitting PRs, it's a good practice to do `git diff main` and see where there are minor changes like this that add visual noise but don't materially affect the PR. (Likewise, if there are refactors that you can extract, that's good to make as a separate PR, but so far I see none of those.)

crates/common_lang_types/src/text_with_carats.rs
```diff
@@ -116,19 +114,35 @@ fn text_with_carats_and_line_count_buffer_and_line_numbers(
...
+
+            let prefix = &line_content[0..start_idx];
+            let highlighted = &line_content[start_idx..end_idx];
+            let suffix = &line_content[end_idx..];
 
+            let colored_source = format!(
+                "{}{}{}",
+                prefix,
+                if colorize_carats {
+                    highlighted.bright_red().to_string()
```
rbalicki2: `highlighted.bright_red()` and `highlighted.normal()`

crates/common_lang_types/src/text_with_carats.rs
```diff
@@ -589,4 +605,21 @@ mod test {
...
+            Span::new(31, 33),
+            3,
+            true,
+        )
+        .0;
+
+        colored::control::unset_override();
+
+        assert!(output.contains("\u{1b}[91m"));
+    }
```
rbalicki2, line 640: could we also do this for at least one test that spans across multiple lines? (Which looks like it works correctly!!! 🎉 )

Image

rbalicki2: though this is an incorrect error lol

rbalicki2 (commented):
Amazing!!!! 🙌  Thank you so much for doing this!!!

Just a few nits + let's add a few tests for the multiline case, and we're g2g. This is soooo nice!!!!

kartikmawani:
Thanks for the detailed review! I've addressed all the points discussed

  I  refactored the loop to calculate start_of_carats and end_of_carats directly as usize. This removed the redundant end_idx variable and the back-and-forth casting between u32 and usize.

Switched to using .normal() for the uncolored branches to match the ColoredString type returned by .bright_red().  

I moved colored::control::unset_override(); inside the main function so it doesn't need to be called in tests.

 Added the requested multiline test case  to verify cross-line highlighting.
Switched all tests to use strict assert_eq! with the exact expected output strings, rather than just checking .contains().

 noted on the "visual noise" regarding line ordering.I'll keep that in mind for future PRs to keep the diffs cleaner

Issue #335, opened 2025-01-24
[epic] Stateful resolvers


rbalicki2:
Something like this

```
iso(`
  field Pet.weight @stateful @mutable {
    height #server fields
    length
  }
`)(({ data, rerender }) => {
  // only called once
  const localStorage = {};

  browser.subscribeToWeight(() => {
    localStorage.weight = (...)
    rerender();
  })

  // these functions are called later repeatedly
  return {
    get: () => localStorage.weight,
    set: ({data, newValue}) => {
      if (newValue == null) {

      }
      localStorage.weight = newValue;
    },
    destructor: () => {
      browser.unsubscribe();
    }
  };
}, reconcileFunction);

```

rbalicki2:
this is extremely redundant with #79 and other ones

rbalicki2:
Possible breakdown of tasks:

# `MultiCallbackSet` optimization

- Use `MultiCallbackSet` to optimize the case where the same component (id + component name + arguments) is rendered multiple times. There should only be a single FragmentSubscription.
- Invariant: there should never be duplicate FragmentSubscriptions (but with different callbacks)
- This `MultiCallbackSet` is an object that contains one or more callbacks, each of which is disposable. When the last callback is removed (and disposed), we trigger the disposal of the `MultiCallbackSet`.
- Disposing the `MultiCallbackSet` will dispose something on the component

# `getOrCreate` at read time

- when a resolver is read, we `getOrCreate` its value, which is stored in a specific location in the store
- this store location is something like: `id` + `resolverName` + `parameters`

## High-level, how subscriptions work
- Imagine we have `AddressComponent -> fullName -> {firstName, lastName}`. Then changes to `firstName` should trigger the `fullName` resolver, which should trigger the `AddressComponent` re-render

```
AddressComponent -> fullName -> firstName
                                \-> firstName      

FirstNameComponent -> firstName
```

- we will re-read AddressComponent fragment twice, and call `setState` twice, but React will coalesce these.
- we will call `FirstNameComponent`'s `setState` first, then `AddressComponent`, but that will still show up as one (consistent) re-render in React

## Garbage collection + subscriptions

- When a stateful resolver (`fullName`) is created (by `AddressComponent`), then we create a FragmentSubscription, with `callback` -> `new MultiCallbackSet().callback`, and with fragmentReference `{ firstName, lastName }`
[...]

PR #913, opened 2025-12-29
include resolvers `fieldName` in reader ast

PatrykWalach:
This is required to properly write resolver results to cache. Part of #335


rbalicki2:
Shouldn't the alias be a read-time-only property? e.g. if we have one fragment with `dave: bestFriend` and one with `john: bestFriend`, that `bestFriend` field should be shared for both?

PatrykWalach:
It should, I think the field name is already present in top level resolver so no need to generate this twice

PR #918, opened 2026-01-05
clippy: match_wild_err_arm

ecalifornica:


crates/isograph_fixture_tests/src/main.rs
```diff
@@ -63,6 +63,7 @@ fn generate_fixtures_for_files_in_folder(
     current_working_directory: CurrentWorkingDirectory,
 ) {
     for read_dir in folder.read_dir().expect("read_dir call failed") {
+        #[expect(clippy::match_wild_err_arm)]
         match read_dir {
```
rbalicki2, line 67: good catch, this should just be `read_dir.unwrap_or_else(|_| panic!("Failed to read an item in {folder:?}"))`. Nice!

PR #923, opened 2026-01-07
clippy: needless_for_each

ecalifornica:
https://rust-lang.github.io/rust-clippy/stable/index.html?groups=pedantic#needless_for_each


rbalicki2:
LGTM, let's also enable the clippy rule somewhere

However, we should just disable clippy in relay-crates. Those are there just for intern, and are never changed.

ecalifornica:
image

rbalicki2:
lgtm as well

PR #925, opened 2026-01-08
clippy: allow all for relay crates

ecalifornica:
https://github.com/isographlabs/isograph/pull/923#issuecomment-3721266007


ecalifornica:
Looks like the workspace key is invalid?

rbalicki2:
okay if it's not easy to enable it only for crates/ then sgtm, let's just enable it and fix it

ecalifornica:
Removed the opt-in-only workspace lint key.

rbalicki2:
Amazing!! Thanks for doing this

PR #928, opened 2026-01-09
fix CI not typechecking

PatrykWalach:
There are some type errors that slipped through CI because tsc was not running

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #931 
- <kbd>&nbsp;2&nbsp;</kbd> #930
- <kbd>&nbsp;1&nbsp;</kbd> #928 👈

### DIFF
--- a/libs/isograph-react/src/core/cache.ts
+++ b/libs/isograph-react/src/core/cache.ts
@@ -61,12 +61,11 @@ export type NetworkResponseValue =
 export type NetworkResponseObject = {
   // N.B. undefined is here to support optional id's, but
   // undefined should not *actually* be present in the network response.
-  readonly [index: ScalarNetworkResponseKey]:
-    | undefined
-    | NetworkResponsePlural<NetworkResponseScalarValue>;
-  readonly [index: LinkedNetworkResponseKey]:
-    | undefined
-    | NetworkResponsePlural<NetworkResponseObject>;
+  readonly [K in
+    | ScalarNetworkResponseKey
+    | LinkedNetworkResponseKey]: K extends ScalarNetworkResponseKey
+    ? undefined | NetworkResponsePlural<NetworkResponseScalarValue>
+    : undefined | NetworkResponsePlural<NetworkResponseObject>;
 } & {
   readonly id?: DataId;
   readonly __typename?: TypeName;


libs/isograph-react/src/core/cache.ts
```diff
@@ -46,18 +47,23 @@ export function getOrCreateItemInSuspenseCache<
...
+  | NetworkResponsePlural<NetworkResponseObject>;
 
 export type NetworkResponseObject = {
   // N.B. undefined is here to support optional id's, but
   // undefined should not *actually* be present in the network response.
-  readonly [index: string]: undefined | NetworkResponseValue;
+  readonly [index: ScalarNetworkResponseKey]:
+    | undefined
+    | NetworkResponsePlural<NetworkResponseScalarValue>;
+  readonly [index: LinkedNetworkResponseKey]:
```
PatrykWalach: TIL: this let's us infer types based on ast kind

Image

the `string & { /* something */}` is required to have to separate indexes.

It is unfortunate `Brand<string, ...>` doesn't work as it cannot be used as an index

libs/isograph-react/src/core/cache.ts
```diff
@@ -233,12 +239,8 @@ function normalizeScalarField(
     return existingValue === undefined || existingValue != null;
   }
 
-  if (isScalarOrEmptyArray(networkResponseData)) {
-    targetStoreRecord[parentRecordKey] = networkResponseData;
-    return existingValue !== networkResponseData;
-  } else {
-    throw new Error('Unexpected object array when normalizing scalar');
-  }
+  targetStoreRecord[parentRecordKey] = networkResponseData;
```
PatrykWalach, line 245: Removed the check, the type of custom scalar is unknown so don't have to validate it at runtime

libs/isograph-react/src/core/cache.ts
PatrykWalach: we could use the same trick for typing the store, this way we are guaranteed, linked fields contain links

PatrykWalach: done this for the store, now assertLink and getLink accepts `DataTypeValueLinked` so it's much easier to check

libs/isograph-react/src/core/cache.ts
rbalicki2: sgtm

rbalicki2 (approved):
lgtm shipit

PR #931, opened 2026-01-10
feat: Throw graphql errors at component level

PatrykWalach:
Reopening after closing #808

---
This is **part 1 of 3 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #931 
- <kbd>&nbsp;2&nbsp;</kbd> #930
- <kbd>&nbsp;1&nbsp;</kbd> #928 👈

### DIFF
--- a/libs/isograph-react/src/core/IsographEnvironment.ts
+++ b/libs/isograph-react/src/core/IsographEnvironment.ts
@@ -146,7 +146,7 @@ export function isWithErrors<T>(
   _value: T | WithErrors<T, StoreError>,
   isFallible: boolean,
 ): _value is WithErrors<T, StoreError> {
-  return isFallible === true;
+  return isFallible;
 }
 
 export function isParentRecordKeyFallible(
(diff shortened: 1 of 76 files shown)


libs/isograph-react/src/core/cache.ts
```diff
@@ -253,26 +328,42 @@ function normalizeLinkedField(
...
+          JSON.stringify(existingValue.errors) !== JSON.stringify(errors)
+        );
+      }
+      targetParentRecord[parentRecordKey] = {
+        kind: 'Data',
+        value: null,
+      };
+      return existingValue?.kind !== 'Data' || existingValue.value != null;
+    }
+    // same thing as below :/
```
PatrykWalach: This is weird, if it's fallible we have to do the same thing, but store data in different format and return different comparison but the rest has to be kept in sync

PatrykWalach: Spit this into multiple checks

libs/isograph-react/src/core/IsographEnvironment.ts
```diff
@@ -120,6 +130,40 @@ export type StoreLink = {
...
+    };
+
+export function isWithErrors<T>(
+  _value: T | WithErrors<T, StoreError>,
+  isFallible: boolean,
+): _value is WithErrors<T, StoreError> {
+  return isFallible;
+}
+
+export function isParentRecordKeyFallible(
```
PatrykWalach, line 152: This gives us type safe read and write

PR #941, opened 2026-01-13
fix nested Query normalization

### LINKED ISSUE #937: Query.__ROOT bugs
- __refetch fields will retrieve the query's id, which is not necessarily __ROOT, e.g. it might be the string "query" (postgraphile)
- this data will be written into the slot for Query.query, instead of Query.__ROOT
- the original data was read from Query.__ROOT
- so the refetch query will not trigger a refetch

PatrykWalach:
Closes #937

### DIFF
--- a/libs/isograph-react/src/core/cache.ts
+++ b/libs/isograph-react/src/core/cache.ts
@@ -590,9 +590,17 @@ function getDataIdOfNetworkResponse(
   variables: Variables,
   index: number | null,
 ): DataId {
+  const __typename =
+    astNode.concreteType ?? dataToNormalize[TYPENAME_FIELD_NAME];
+  if (__typename == null) {
+    throw new Error(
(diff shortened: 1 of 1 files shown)


libs/isograph-react/src/tests/normalizeData.test.ts
```diff
@@ -96,100 +93,37 @@ describe('normalize undefined field', () => {
...
-          query: {
-            __link: ROOT_ID,
-            __typename: 'Query',
-          },
-        },
-      },
-    } satisfies BaseStoreLayerData);
-  });
-
-  test('should be read', () => {
```
PatrykWalach, line 146: I've removed the read test. As long as we normalized link correctly - this is asserted in the previous test - then we are good to go, there's no special behavior during reading that needs to be tested

rbalicki2:
It worked!!! 🎉

PR #942, opened 2026-01-13
convert reader artifacts to function to prevent cycles

### LINKED ISSUE #935: Artifact default export should be a function
- Every default export of every artifact should be a function
- This will fix the issue where there are circular dependencies, e.g. if one doesn't have the babel transform:
- Root imports entrypoint Query.Foo
- Foo imports iso.ts
- iso.ts imports entrypoint
- result: circular dependency, on initial tick, resolver is undefined, and the default export { resolver } contains resolver: null

PatrykWalach:
Manually tested inside pet demo

Closes #935

### DIFF
--- a/crates/artifact_content/src/eager_reader_artifact.rs
+++ b/crates/artifact_content/src/eager_reader_artifact.rs
@@ -255,16 +255,16 @@ pub(crate) fn generate_eager_reader_condition_artifact<TCompilationProfile: Comp
         "import type {{ EagerReaderArtifact, ReaderAst, Link }} from '@isograph/react';\n\
         {reader_import_statement}\n\
         const readerAst: ReaderAst<{reader_param_type}> = {reader_ast};\n\n\
-        const artifact: EagerReaderArtifact<\n\
+        const artifact = (): EagerReaderArtifact<\n\
         {}{reader_param_type},\n\
         {}{reader_output_type}\n\
-        > = {{\n\
+        > => ({{\n\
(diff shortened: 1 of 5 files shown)


crates/artifact_content/src/eager_reader_artifact.rs
```diff
@@ -118,16 +118,16 @@ pub(crate) fn generate_eager_reader_artifacts<TCompilationProfile: CompilationPr
...
             const readerAst: ReaderAst<{reader_param_type}> = {reader_ast};\n\n\
-            const artifact: EagerReaderArtifact<\n\
+            const artifact = (): EagerReaderArtifact<\n\
             {}{reader_param_type},\n\
             {}{reader_output_type}\n\
-            > = {{\n\
+            > => ({{\n\
             {}kind: \"EagerReaderArtifact\",\n\
             {}fieldName: \"{client_selectable_name}\",\n\
             {}resolver,\n\
```
PatrykWalach, line 127: Either the whole artifact needs to be wrapped or just the resolver:
`resolver: (...args) => resolver(...args)`

PatrykWalach: I think the identity of the resolver will matter when we fix HMR

rbalicki2: I'm not sure I follow tbh, but what happens now lgtm

rbalicki2 (approved):
shipittt thank you for doing this!!

PR #951, opened 2026-01-25
clippy: deny items_after_statements

ecalifornica:
https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#items_after_statements 

TIL a thing about scoping


rbalicki2:
LGTM! TIL about this lint. Seems very reasonable.

I would not support a "no use before define" lint, though, in case such a monstrosity exists in Rust :-P

PR #956, opened 2026-01-27
remove fetchable types from network protocol

### LINKED ISSUE #944: Get rid of fetchable types
Overall goal is to get rid of the concept of fetchable types from Isograph, and instead, move that info to network protocol entity associated data. They are a GraphQL-ism. Blocks #848 

## Fetchable types that flow into creating the query text

In several locations, the fetchable type ultimately flows into generate_query_text.

Let's change the signature of `generate_query_text` to not accept `root_operation_name`, and instead read the fetchable type from within the trait method, and remove all calls to `fetchable_types` that are now unneeded. These fetchable types are used to determine the keyword (query/mutation/subscription) to use.

`ImperativelyLoadedFieldVariant.root_object_entity_name` only flows into generate_query_text and can be removed

(`query_name` also doesn't need to be passed in)

## `generate_artifacts:339`, `create_merged_selection_set:1166`

- flows into WrappedSelectionMapSelection::LinkedField.parent_object_entity_name
- which flows into MergedLinkedFieldSelection.parent_object_entity_name (create_merged_selection_set.rs:1500)
- which flows into raw_response_type.rs:99 

So here, we should get rid of `pub parent_object_entity_name: EntityName`, and change `pub concrete_target_entity_name: Option<EntityName>,` to be an enum with variants `Concrete` and `Abstract`.

Then, we don't need fetchable types anywhere!

## Add associated data

- Modify the network protocol entity associated data to include whether a type is fetchable or not, i.e. `Option<RefetchStrategy>` (??).
- If it is fetchable, we want to know
    - how to wrap the merged selection map (options are "do nothing", and "use the node field and use an inline fragment". Later, we can add the option to use another field).
    - what keyword to use (query/mutation/subscription)

## Add a `wrap_merged_selection_map` method to `NetworkProtocol`

- `fn wrap_merged_selection_map(db, root_entity, merged_selection_map) -> DiagnosticResult<MergedSelectionMap>`
- in graphql-land, accesses the network protocol entity associated data

## Then, modify `generate_query_text` to not access fetchable types internally

- As of this step, `generate_query_text` looks up the fetchable types to determine what keyword to use.
- generate_query_text should take a `root_entity: EntityName`
- look up the network protocol entity associated data and get the keyword from there

## Allow entrypoints that aren't on root fields!!! Maybe.

ch1ffa:
Closes #944


crates/graphql_network_protocol/src/graphql_network_protocol.rs
```diff
@@ -234,26 +272,112 @@ impl TargetPlatform for JavascriptTargetPlatform {
...
+                Diagnostic::new(format!("Type `{root_entity}` is not fetchable."), None)
+            })?;
+
+        match fetchable_info.wrap_strategy {
+            GraphQLWrapStrategy::None => Ok(WrapMergedSelectionMapResult {
+                root_entity,
+                merged_selection_map,
+            }),
+            GraphQLWrapStrategy::Node { query_root } => {
+                let already_wrapped_with_node_field = merged_selection_map.keys().any(|key| {
```
rbalicki2: - under what circumstances would this occur?
- would this have false positives for `AnyOtherType.node` for non-Query types?
- and if we're on the query type, don't we have `GraphQLWrapStrategy::None`?

crates/isograph_schema/src/create_merged_selection_set.rs
```diff
@@ -119,15 +121,28 @@ impl MergedScalarFieldSelection {
...
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
+pub enum ConcreteTargetEntityName {
+    Concrete(EntityName),
+    Abstract,
+}
+
+impl fmt::Display for ConcreteTargetEntityName {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            ConcreteTargetEntityName::Concrete(name) => write!(f, "\"{name}\""),
```
rbalicki2, line 146: this is clean, love it

crates/artifact_content/src/imperatively_loaded_fields.rs
```diff
@@ -75,44 +75,43 @@ pub(crate) fn get_paths_and_contents_for_imperatively_loaded_field<
...
-            "Expected root type to be fetchable here. \
-            This is indicative of a bug in Isograph.",
-        );
-
     let query_name = format!("{root_parent_object}__{client_selection_name}")
         .intern()
         .into();
 
     let query_text_selection_map_wrapped =
         selection_map_wrapped(nested_selection_map.clone(), subfields_or_inline_fragments);
```
rbalicki2, line 83: in `generate_artifacts`, we never call `selection_map_wrapped` and then `wrap_merged_selection_map`. (In fact, maybe that should be reflected in the types: `UnwrappedMap -> WrappedMap`?)

Is it a mistake that we are doing both here? 

`root_object_entity_name` is always `Query` or `Mutation` (or probably Subscription, but not in our demos), so perhaps we should avoid calling `wrap_merged_selection_map`?

rbalicki2: (I couldn't get what I'm suggesting to work, FWIW, so I'm not opposed to this, but let's at least leave a TODO to figure out what's going on.)

crates/artifact_content/src/raw_response_type.rs
```diff
@@ -30,157 +33,175 @@ pub fn generate_raw_response_type<TCompilationProfile: CompilationProfile>(
...
+                                .reference(),
+                            inner_text
+                        )
+                    ));
+                }
+                _ => {}
             }
         }
+
+        return;
```
rbalicki2, line 169: is it invalid for a merged selection map to have both fragments and non-fragments?

... it looks like we have that case in a few places! I'll investigate on main. Seems like a bug!

rbalicki2: damn... wtf is going on with `MergedSelectionMap`?? We have all sorts of invalid states allowed in it. Okay, then never mind, I'll create an issue for this, and this lgtm here

crates/graphql_network_protocol/src/graphql_network_protocol.rs
rbalicki2: At least one reason it panics is because we are calling `selection_map_wrapped` then `wrap_merged_selection_map` in imperatively_loaded_fields.rs. 🤔

rbalicki2 (commented):
Looks awesome to me, basically only two (related) questions:

- can we avoid the duplicate wrapping, and
- can we avoid the is already wrapped check

crates/artifact_content/src/imperatively_loaded_fields.rs
ch1ffa: Hmm, for pet demo `root_object_entity_name` could be `ICheckin` for `RefetchQueryNormalizationArtifact` in `SmartestPetRoute`

crates/graphql_network_protocol/src/graphql_network_protocol.rs
ch1ffa: If I remove this, we well panic in `generate_raw_response_type` because somewhere we call wrap the second time and push `node(id)` under `AdItem` in pet demo for example

crates/artifact_content/src/imperatively_loaded_fields.rs
rbalicki2: woohoo, broke it down into tiny steps, claude got a decent solution.

I don't really like having separate methods on NetworkProtocol, but I couldn't figure out a way to avoid double wrapping otherwise, and I like the `already_wrapped_in_node` check even less.

PR #960, opened 2026-02-01
[FIX] items in `ParentCache` can be cleared by callback twice

PatrykWalach:


libs/isograph-react-disposable-state/src/CacheItem.test.ts
```diff
@@ -153,7 +153,7 @@ describe('CacheItem', () => {
 
       expect(() => {
         disposeTemporaryRetain();
-      }).toThrow();
```
PatrykWalach, line 156: In this test it would throw `A temporary retain was cleared, for which the CacheItem is in an invalid state.` but at runtime it was possible to go into a different branch and cause issues (temporary retain count could have been decremented more than once).

PR #961, opened 2026-02-01
typecheck tests

### LINKED ISSUE #958: type errors in startUpdate.test.ts?

PatrykWalach:
Closes #958

### DIFF
--- a/libs/isograph-react/src/core/cache.ts
+++ b/libs/isograph-react/src/core/cache.ts
@@ -439,10 +439,11 @@ function normalizeNetworkResponseObject(
 export type ParentRecordKey = string;
 
 /**
- * these don't actually start with a or b, but this let's us use multiple indexes e.g.
+ * these don't actually start with the different prefixes, but they let
+ * us use multiple indexes, e.g.
  * {
- *   readonly [key: `a-${string} `]: ...
- *   readonly [key: `b-${string} `]: ...
+ *   readonly [key: `NOT_AN_ACTUAL_TYPE_LinkedParentRecordKey_${string} `]: ...
+ *   readonly [key: `NOT_AN_ACTUAL_TYPE_ScalarParentRecordKey_${string} `]: ...
  * }
  */
 export type LinkedParentRecordKey =
@@ -522,10 +523,11 @@ function getStoreKeyChunkForArgument(argument: Argument, variables: Variables) {
 export type NetworkResponseKey = string;
 
 /**
- * these don't actually start with a or b, but this let's us use multiple indexes e.g.
+ * these don't actually start with the different prefixes, but they let
+ * us use multiple indexes, e.g.
  * {
- *   readonly [key: `a-${string} `]: ...
- *   readonly [key: `b-${string} `]: ...
+ *   readonly [key: `NOT_AN_ACTUAL_TYPE_LinkedNetworkResponseKey_${string} `]: ...
+ *   readonly [key: `NOT_AN_ACTUAL_TYPE_ScalarNetworkResponseKey_${string} `]: ...
  * }
  */
 export type LinkedNetworkResponseKey =


rbalicki2:
updating...maybe the errors are transient?

rbalicki2:
```ts
    const store: BaseStoreLayerData = {
      Economist: {
        '1': {
          __typename: 'Economist',
          id: '1',
        },
      },
      Query: {
        [ROOT_ID]: {
          node____id___1: {
            __typename: 'Economist',
            __link: '1',
          },
          query: {
            __link: ROOT_ID,
            __typename: 'Query',
          },
        },
      },
    };
```
    
fails to typecheck without `as StoreRecord`. Alternatively, we can do:

```ts
const store: BaseStoreLayerData = {
  ['Economist' as TypeName]: {
    ['1' as DataId]: {
      __typename: 'Economist' as TypeName,
      id: '1' as DataId,
    },
  },
  Query: {
    [ROOT_ID]: {
      ['node____id___1' as DataId]: {
        __typename: 'Economist' as TypeName,
        __link: '1' as DataId,
      },
      ['query' as DataId]: {
        __link: ROOT_ID,
        __typename: 'Query' as TypeName,
      },
    },
  },
};
```

but this also fails to typecheck. Claude code tells me its because typescript cannot handle the fact that `__typename` and `ScalarParentRecordKey | LinkedParentRecordKey` conflict. It suggests we make brand not optional in `& { brand: ... } `

I'm not particularly concerned about this, nor really know what to do in this situation.

Anyway maybe we should use a raw response type to write data into the store in normalizeData.test.ts?

rbalicki2:
from claude

  Summary: Making the brands required (brand: instead of brand?:) fixes the issue because:

  1. With optional brands, plain strings like __typename structurally match both ScalarParentRecordKey and LinkedParentRecordKey
  2. TypeScript then requires the value to satisfy both index signatures
  3. string isn't assignable to DataTypeValueLinked, hence the error

  With required brands, plain strings don't match the branded types, so __typename and id are only checked against their explicit property definitions in the intersection type, not the index signatures.

  The one place that needed fixing was optimisticProxy.ts:536 where Object.entries returns plain string keys - a cast is appropriate there since we're iterating over actual record keys at runtime.

libs/isograph-react/src/core/cache.ts
```diff
@@ -433,15 +436,19 @@ function normalizeNetworkResponseObject(
...
+ * these don't actually start with a or b, but this let's us use multiple indexes e.g.
+ * {
+ *   readonly [key: `a-${string} `]: ...
+ *   readonly [key: `b-${string} `]: ...
+ * }
+ */
+export type LinkedParentRecordKey =
+  `NOT_AN_ACTUAL_TYPE_LinkedParentRecordKey_${string}`;
+export type ScalarParentRecordKey =
+  `NOT_AN_ACTUAL_TYPE_ScalarParentRecordKey_${string}`;
```
PatrykWalach, line 452: This brand was way too hacky, this works slightly better

PatrykWalach: The branded strings were not allowing normal strings, with this it's possible to do an escape hatch
```tsx
{
  [key: LinkedParentRecordKey]: "Foo"
  [key: ScalarParentRecordKey]: "Bar"
  [key: string]: "Foo" | "Bar"
}
```

rbalicki2: if it works, lets shipit

PatrykWalach:
There are errors in `__isograph` dir

libs/isograph-disposable-types/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.5.3",
   "description": "Common types for disposable items",
   "homepage": "https://isograph.dev",
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "main": "./dist/index.cjs",
```
PatrykWalach, line 6: This is changed because I've upgraded `tsdown`

libs/isograph-react/tsconfig.test.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.build.json",
+  "compilerOptions": {
+    "jsx": "react",
+    "noUnusedLocals": false,
```
PatrykWalach, line 5: Separate config to disable unused locals and resolve paths

PatrykWalach:
Ready for review

libs/isograph-react-disposable-state/src/useLazyDisposableState.ts
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-import type { ItemCleanupPair } from '@isograph/isograph-disposable-types';
+import type { ItemCleanupPair } from '@isograph/disposable-types';
```
rbalicki2, line 3: how did this work??

rbalicki2 (approved):
shipitttt

libs/isograph-react-disposable-state/src/useLazyDisposableState.ts
PatrykWalach: It was probably bundling the src files of other packages, in root tsconfig.json we have "paths"

PR #966, opened 2026-02-16
clippy: deny needless_continue

ecalifornica:
https://rust-lang.github.io/rust-clippy/rust-1.92.0/index.html#needless_continue

TIL about loop labels, neat.
https://doc.rust-lang.org/beta/reference/expressions/loop-expr.html#loop-labels


ecalifornica:
Looks like this timeout might've been a flake?

